Issue #4, opened 2019-10-11
Can't someone find who's working on an implementation in JIRA?


lehors:
The README says that "If you are interested in working on the implementation for an active RFC, but cannot determine if someone else is already working on it, feel free to ask".

However, the README also says that "Every accepted RFC has an associated issue tracking its implementation in the Fabric JIRA issue tracker".

So, shouldn't you be able to find if someone else is working on the RFC implementation by checking whether the associated JIRA ticket is assigned?

kchristidis:
Good catch.

kchristidis:
Pushed a fix, and made that section less verbose. LMK what you think.

PR #5, opened 2019-10-11
Address subteam and implementation tracking references

### LINKED ISSUE #3: Subteam is undefined
The README refers to "subteam" but it's unclear what is considered to be a subteam.
I can see @kchristidis replaced several occurrences with maintainers so I don't know whether the remaining references are intentional or a mere oversight.

kchristidis:
Closes #3, #4, and #6.

Signed-off-by: Kostas Christidis <kostas@christidis.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,17 +190,11 @@ a comment describing the rationale for the decision.
 Some accepted RFCs represent vital changes that need to be implemented right
 away. Other accepted RFCs can represent changes that can wait until a
 developer feels like doing the work. Every accepted RFC has an associated
-issue tracking its implementation in the [Fabric JIRA issue tracker](https://jira.hyperledger.org/projects/FAB/issues);
-thus that associated issue can be assigned a priority via the triage process
-that the team uses for all issues related to Fabric.
+issue tracking its implementation in the [Fabric JIRA issue tracker](https://jira.hyperledger.org/projects/FAB/issues).
 
 The author of an RFC is not obligated to implement it. Of course, the RFC
 author, as any other developer, is welcome to post an implementation for review
-after the RFC has been accepted.
-
-If you are interested in working on the implementation for an active RFC, but
-cannot determine if someone else is already working on it, feel free to ask
-(e.g. by leaving a comment on the associated issue).
+after the RFC has been accepted. Use JIRA for this.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -127,8 +127,9 @@ postpone).
   - For RFCs with lengthy discussion, the motion to FCP is usually preceded by
   a *summary comment* trying to lay out the current state of the discussion and
   major trade-offs/points of disagreement.
-  - Before actually entering FCP, *all* maintainers must sign off; this is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,9 +127,8 @@ postpone).
   - For RFCs with lengthy discussion, the motion to FCP is usually preceded by
   a *summary comment* trying to lay out the current state of the discussion and
   major trade-offs/points of disagreement.
-  - Before actually entering FCP, *all* members of the subteam must sign off;
-  this is often the point at which many subteam members first review the RFC in
-  full depth.
+  - Before actually entering FCP, *all* maintainers must sign off; this is
```
kchristidis: I think we should reconsider the "all maintainers" policy and move to a 2/3rds rule, or a similar majority rule. Otherwise we run the risk of being blocked by maintainers who are not active.

kchristidis: See #6.

lehors: I agree, requiring sign off from all the maintainers is not practical. This is actually why I wasn't sure the reference to subteam was merely an oversight. I thought it might refer to a subset of the maintainers.

lehors (approved):
Other than the issue related to the "all maintainers rule" duly noted, this PR addresses issues #3 and #4.
Thanks.

README.md
C0rWin: Why not to identify components and components owners so they will provide consent by signing-off?

jyellick:
Agreed that all maintainer's is not practical.  But, since there is no such thing as a 'subteam' (at least as Fabric exists today), getting their agreement seems even less practical.  To me, it makes sense to remove the unknown terminology, rather than invent a definition to try to make the paragraph make sense.  If we want to address the 'all maintaners' bit (which I think we do), then that can would be another PR.

kchristidis:
@jyellick: Is this a response to @C0rWin's comment, or to this PR in general? (I'm guessing the former, in which case I agree, but it was posted as a comment on the entire PR so the intent is a bit lost on me.)

jyellick:
A bit of both.  Primarily it was meant in support of merging this PR as is, since 'subteam' is an undefined and meaningless term.

I was also in a way responding to @C0rWin with respect to his suggestion to identify component owners, that if we want to go that route, we can do it in another PR, but I'd rather we construct the text to match our definitions, rather than create definitions to match the text.

C0rWin:
@jyellick I have just pointed out suggestion, wasn't trying to suggest making changes into current PR. I agree we have to make more concrete formulation which could be a different PR.

README.md
denyeart: Many rfcs will span components. I'd prefer we go with a simple majority of maintainers here.

christo4ferris: Agree that we need a majority of maintainers agreeing. All is impractical and even 2/3s seems extreme.

kchristidis: This PR is not meant to address this issue. This PR removes the subteam reference.

I will push a PR that closes #6 with a "majority" comment.

christo4ferris (approved):
LGTM thanks all

PR #7, opened 2019-10-13
Add sections for testing and known dependencies

C0rWin:
Proposing to add information which Fabric component is in subject of RFC change. Add sections to outline testing plan and list known dependencies on JIRA issues or relations to other RFCs.

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,6 +1,7 @@
 - Feature Name: (fill me in with a unique identifier, my_awesome_feature)
 - Start Date: (fill me in with today's date, YYYY-MM-DD)
 - RFC PR: (leave this empty)
+- Fabric Component: (fill me in with underlined fabric component, core, orderer/consensus and etc.)
 - Fabric Issue: (leave this empty)
 
 # Summary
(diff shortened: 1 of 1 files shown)


0000-template.md
```diff
@@ -91,6 +93,23 @@ it does not on its own motivate an RFC.  Please also take into consideration
 that Fabric sometimes intentionally diverges from common distributed
 ledger/blockchain features.
 
+# Testing
```
kchristidis, line 96: I understand we don't want the RFCs to become a hodgepodge of hand-wavy suggestions, but this passage in particular strikes me as too much process, and makes the RFC process a bit too onerous. It'd probably be more useful to discuss these issues when we go about the RFC, rather than making them a prerequisite for a proposal.

kchristidis: ^^ This is meant to apply to lines 96-112.

C0rWin: This is passage is to make sure that we are not missing dependencies and understand complications of the suggested RFC. Though I understand this might feel too much process this is quite common, for example this is approach adopted for many years in JEP (Java Enhancement Proposal).

christo4ferris (approved):

PR #1, opened 2019-10-11
Fix README

kchristidis:
1. Replace reference to sub-teams with maintainers — addresses #2.
2. Explain reference to "RFC assignee" — closes #2.
3. Normalize references to Fabric: existing version mixes up "Hyperledger Fabric", "fabric", "Fabric"; this CR refers to the project by its full name the first time, and switches to "Fabric" for all remaining references.
4. Fix broken table of contents.
5. Remove unused or broken links.
6. Remove "help this is all too informal!" section.
7. Conform to Markdown linter directives: surround headings with blank lines, etc.
8. Enforce consistency with line breaks: existing version goes anywhere from ~80 to 100 characters, this one sticks to 80 as the upper bound with the exception lines that include links

PR #16, opened 2019-12-11
Ledger API

mbwhite:
Proposal to formalize and improve upon the Ledger APIs currently in the Programming Model.

**EDIT** : Feb 5th - added more info on implicit collections. Create/Update semantics have replaced put based on community feedback.


mbwhite:
This RFC was presented to the [contributors meeting](https://wiki.hyperledger.org/display/fabric/Contributor+Meetings) on 2019-12-11. The recording is available [here](https://wiki.hyperledger.org/download/attachments/6426811/20191211_contributors_meeting.mp4?api=v2).

It was also discussed as the main topic on the Developer Community call -  [recording here](https://wiki.hyperledger.org/download/attachments/6423554/2019-1212-fabricdevcommunitycall.zip?api=v2)

andrew-coleman:
+1

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+- Feature Name: Ledger_API
+- Start Date: 2019-12-03
+- RFC PR: (leave this empty)
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This feature is a continuation of the updates made in recent releases to the Fabric Programming Model. This RFC's aim is to clarify the abstractions used to work with ledger data in a contract. It's scope is restricted to what is today referred to as the 'ChaincodeStub API', and specifically at the APIs around states.
```
mastersingh24:
```suggestion
This feature is a continuation of the updates made in recent releases to the Fabric Programming Model. This RFC's aim is to clarify the abstractions used to work with ledger data in a contract. It's scope is restricted to what is today referred to as the 'ChaincodeStub API'-- specifically  the APIs around states.
```

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+This is a cross-language RFC for Golang, JavaScript, TypeScript and Java.
+
+# Motivation
+[motivation]: #motivation
+
+The updated Programming Model, has introduced higher level abstractions to both the contracts and the client SDKs. Many users had already written high-level utility or support code within their own applications. Abstractions help us think about and solve problems, improving usability, speed of development and performance.
```
mastersingh24:
```suggestion
The updated Programming Model has introduced higher level abstractions to both the contract and the client SDKs. Many users have already written high-level utility or support code within their own applications. Abstractions help us think about and solve problems, improving usability, speed of development and performance.
```

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+Examples, such as the Commercial Paper tutorial, also take this approach. It simplifies the concepts and makes working with business level data simpler. (c.f. https://jira.hyperledger.org/browse/FABCN-42)
```
mastersingh24:
```suggestion
Examples, such as the Commercial Paper tutorial, also take this approach. It simplifies the concepts and makes working with business-level data simpler. (c.f. https://jira.hyperledger.org/browse/FABCN-42)
```

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+This RFC (which is anticipated is the first of a series) will formalize the existing 'stub-api' into a conceptual architecture. A key aspect in all the programming model work is that the conceptual model is consistent between languages. The polyglot developer will therefore have minimal friction moving between languages.  Note that this a *conceptual* model; each implementation will use idiomatic language code, and where appropriate, specifics of the language to give the best developer experience.
```
mastersingh24:
```suggestion
This RFC (which is anticipated as the first in a series) will formalize the existing 'stub-api' into a conceptual architecture. A key aspect in all the programming model work is that the conceptual model is consistent between languages. The polyglot developer will therefore have minimal friction moving between languages.  Note that this a *conceptual* model; each implementation will use idiomatic language code, and where appropriate, specifics of the language to provide the best developer experience.
```

mastersingh24 (changes requested):
more to come

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+Across all the languages today, the stub API is broadly consistent between languages. However there are variations that are confusing when looking across languages. Within each language as well there are APIs that have minor inconsistencies. Whilst not major, these differences are sufficient to cause a developer uncertainty in how an API should be used. Introducing a ledger formalizes and integrates existing concept..
```
mastersingh24:
```suggestion
The stub API is broadly consistent across languages. There are, however, variations and minor inconsistencies that cause confusion when looking across languages. Whilst not major, these differences are sufficient to cause a developer uncertainty in how an API should be used. Introducing a ledger formalizes and integrates existing concepts.
```

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+Future RFCs will focus on the persistence of the business data; many of the same helper classes used today are based around the idea of serializing data, and move the knowledge of key creation into the business objects themselves. (c.f. https://jira.hyperledger.org/browse/FABCN-92)
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A complete solution using Fabric can be represented as follows:
+![](0000-ledger-api-ConceptualArchitecture.png)
+
+A typical sequence is for an application to connect to a peer (the 'gateway' into the network channel), and obtain a reference to a smart contract. The Network-shared transaction logic in the Smart Contract is then be invoked by name, and with optional data arguments and return values. The data arguments can include transient data if needed. The logic in the Smart Contract will typical reference and update the shared transaction data, subject of course to exactly how the code is written.
```
mastersingh24:
```suggestion
A typical sequence is for an application to connect to a peer (the 'gateway' into the network channel), and obtain a reference to a smart contract. The Network-shared transaction logic in the Smart Contract is then be invoked by name with optional data arguments and return values. The data arguments can include transient data if needed. The logic in the Smart Contract will typical reference and update the shared transaction data, subject of course to exactly how the code is written.
```

text/0000-ledger-api.md
```diff
@@ -0,0 +1,305 @@
...
+
+The 'Ledger' is, therefore, a primary abstraction in a blockchain system, including in Hyperledger Fabric; this abstraction allows the concepts of states, world state, collections, transactions, blocks etc. to be tied together and formalised. 
+
+Looking at the objects that make up this conceptual architecture, a full set of objects would have the following relationships. 
+
+![](0000-ledger-api-relationships.png)
+
+Note that this RFCs is concerned with the ledger-api and the ledger,collection and state; the 'white' boxes will be subsequent RFCs.
```
mastersingh24:
```suggestion
Note that this RFC is concerned with the ledger-api and the ledger,collection and state; the 'white' boxes will be subsequent RFCs.
```

mastersingh24 (changes requested):

denyeart:
There is a lot of merit to this proposal, but before approving, I'd like to see more information about differences with existing APIs, why the new APIs are better (I believe they are), and whether this would replace or complement existing APIs.
Additionally, I'd like to see more attention to the new application patterns that have become available in Fabric v2.0, to ensure this is consistent with and supports the new patterns. Potentially use a concrete sample such as asset transfer with a combination of public channel and private data, to prove out that the APIs make sense.

mbwhite:
Thanks @denyeart, will update the text, but to briefly address some of the points

- All existing APIs are present with the same function - much as the ContractAPI augmented the Chaincode interface.
- The code examples in the RFC are reworkings of some of the Commercial Paper functions, we could link to a AssetTransfer variant in addition
- The new patterns such as the private data, implicit collections have been considered and the functions are there to support this approach.  We can highlight these.

denyeart:
Thanks @mbwhite . Note that Commercial Paper is indeed an asset transfer scenario, so no need to provide two different scenario variants. There is good coverage for the basic puts/gets in both the existing APIs and new proposal. I'm focusing my question on the more advanced options that are needed in real world contract scenarios, such as state-based endorsement, implicit private data collections, indicating agreement and coming to consensus on a business transaction, access control decisions (ownership credentials to transfer an asset, private data access, etc), and sharing/verifying private. Basically all the concepts that are mentioned in https://hyperledger-fabric.readthedocs.io/en/release-2.0/private-data/private-data.html#example-scenario-asset-transfer-using-private-data-collections , which are the same concepts envisioned for a more advanced Commercial Paper scenario. Do the existing APIs have gaps in those areas, that the new APIs will close?

I realize this comment goes beyond the scope of the specific RFC, but I want to ensure alignment of this RFC with the bigger picture. I'll provide specific comments on this RFC in next comment.

denyeart:
Some thoughts on the specific collection proposals:

1) The RFC is proposing collections as a key abstraction layer on top of Fabric. I agree, but would actually like to go further and not just have it as an abstraction, but push collections (not just private collections, but public collections) into core Fabric. This way it would be possible to set endorsement policies on public collections, and then collections would be a first class concept straight through from programming model to core Fabric. These changes are not prioritized nor designed in core Fabric, so we need to be cautious about programming model changes in this area.

2) It is unclear how the new org-specific implicit collections would be handled in the proposed API. This is a critical piece of commercial paper advanced scenario, and achieving consensus on business transactions in general. And as mentioned in #1, public implicit collections may be added to Fabric in the future, to support public agreement in addition to private agreement.

3) There is not parity in public data and private data shim APIs. For example paging is not yet supported on private data. Additionally, you cannot combine private data range queries and updates in a transaction (all the peers may not have the same private data, so consistent results cannot be verified at validation time). For these reasons it may not make sense to combine public and private collection APIs.

denyeart (changes requested):
Some thoughts on the specific collection proposals:

1) The RFC is proposing collections as a key abstraction layer on top of Fabric. I agree, but would actually like to go further and not just have it as an abstraction, but push collections (not just private collections, but public collections) into core Fabric. This way it would be possible to set endorsement policies on public collections, and then collections would be a first class concept straight through from programming model to core Fabric. These changes are not prioritized nor designed in core Fabric, so we need to be cautious about programming model changes in this area.

2) It is unclear how the new org-specific implicit collections would be handled in the proposed API. This is a critical piece of commercial paper advanced scenario, and achieving consensus on business transactions in general. And as mentioned in #1, public implicit collections may be added to Fabric in the future, to support public agreement in addition to private agreement.

3) There is not parity in public data and private data shim APIs. For example paging is not yet supported on private data. Additionally, you cannot combine private data range queries and updates in a transaction (all the peers may not have the same private data, so consistent results cannot be verified at validation time). For these reasons it may not make sense to combine public and private collection APIs.

mbwhite:
> Some thoughts on the specific collection proposals:
> 
> 1. The RFC is proposing collections as a key abstraction layer on top of Fabric. I agree, but would actually like to go further and not just have it as an abstraction, but push collections (not just private collections, but public collections) into core Fabric. This way it would be possible to set endorsement policies on public collections, and then collections would be a first class concept straight through from programming model to core Fabric. These changes are not prioritized nor designed in core Fabric, so we need to be cautious about programming model changes in this area.

We've approached the API by using variations of the Factory pattern, and also via the use of interfaces. The abstractions here are well defined but are written with the future in mind. If additional capabilities are provided on other types of collections, this abstraction would stand, and could be extended.  Caution is good, and the approach here should be sustainable.

> 2. It is unclear how the new org-specific implicit collections would be handled in the proposed API. This is a critical piece of commercial paper advanced scenario, and achieving consensus on business transactions in general. And as mentioned in #1, public implicit collections may be added to Fabric in the future, to support public agreement in addition to private agreement.

This wasn't spelled out in detail, will update the RFC to pull this out; in essence, the name of the collection is the key, and there is a helper function to provide the ability to map a mspID to the collection name. Likewise if another type collection such as public implicit collections, these could be handled in a similar manner. 

> 3. There is not parity in public data and private data shim APIs. For example paging is not yet supported on private data. Additionally, you cannot combine private data range queries and updates in a transaction (all the peers may not have the same private data, so consistent results cannot be verified at validation time). For these reasons it may not make sense to combine public and private collection APIs.

One of the motivations was to make clear differences in the APIs, be that limitation of the underlying technology, or just variations that have occurred. The aim is to make developers' life easier and more productive. I do not believe that separating out the APIs into different naming or classes would provide a developer with vastly greater insight. 

Have updated the RFC to indicate that for these limitations, we have the ability to be able spot them as soon as possible and fail-fast with a predictable error case. Some of the restrictions are dependent on the sequence of APIs

text/0000-ledger-api.md
```diff
@@ -0,0 +1,380 @@
...
+
+Examples, such as the Commercial Paper tutorial, also take this approach. It simplifies the concepts and makes working with business-level data simpler. (c.f. https://jira.hyperledger.org/browse/FABCN-42)
+
+This RFC (which is anticipated as the first in a series) will formalize the existing 'stub-api' into a conceptual architecture. A key aspect in all the programming model work is that the conceptual model is consistent between languages. The polyglot developer will therefore have minimal friction moving between languages.  Note that this a *conceptual* model; each implementation will use idiomatic language code, and where appropriate, specifics of the language to give the best developer experience.
+
+The stub API is broadly consistent between languages. There are, however, variations and minor inconsistencies that cause confusion when looking across languages. Whilst not major, these differences are sufficient to cause a developer uncertainty in how an API should be used. Introducing a ledger formalizes and integrates existing concepts.
```
cendhu, line 28: Curious to know about these inconsistencies. An example would be better.

text/0000-ledger-api.md
```diff
@@ -0,0 +1,380 @@
...
+
+The stub API is broadly consistent between languages. There are, however, variations and minor inconsistencies that cause confusion when looking across languages. Whilst not major, these differences are sufficient to cause a developer uncertainty in how an API should be used. Introducing a ledger formalizes and integrates existing concepts. 
+
+To allow for functional differences between different but very similar operations the stubAPI has chosen an approach to have a 'broad-and-flat' approach to APIs. For example there are query APIs per type of collection, `getPrivateDataByRange` and `getStateByRange`, by apply modifiers to the function name the function is extended. The length of the resulting list of APIs is larger and should additional function become available will increase. A long list of APIs increases the difficulty of finding the API needed, especially without autocomplete-style tool support. By formalizing the APIs in the proposed structure makes an overall more sustainable model to support future changes.
```
cendhu, line 30: why _getPrivateDataByRange_ and _getStateByRange_ are referred to as the API accessing two different types of collections. By collection/private data, we store the private data locally on the peer (as per the membership) and store only the hash of the private data (as proof of the existence of data) on the blockchain. The world/public state refers to data present in the blocks -- which are public data and hash of private data.

mbwhite: Correct yes at a Technical Fabric level; From a developer of the Smart Contract, the way they access the data follows the same semantics. Hence allowing the same style of API to access them - but what they are doing 'underneath' is as you state.

text/0000-ledger-api.md
mbwhite: I've also added some links to JIRAs later in the RFC.

A lot of these are minor in the spelling of the APIs varies between languages. 
The return type of the iterators is some APIs in node.js wrapped in other objects. 
APIs in Java have been missed out for example.

text/0000-ledger-api.md
```diff
@@ -0,0 +1,380 @@
...
+
+![](0000-ledger-api-relationships.png)
+
+Note that this RFC is concerned with the ledger-api and the ledger,collection and state; the 'white' boxes will be subsequent RFCs.
+
+## Ledger
+
+A Ledger's current value is held in a set of collections. A single ledger is available to any given deployed contract, though other ledgers may be available within the Network Channel and can be accessed via other contracts
+
+Of the collections that are reference by the ledger, there a single collection in the ledger that is public (the 'WorldState') and held by every organization in the Network Channel. A ledger may have 1 or more private collections that are available only in some organizations. Starting at Fabric v2.0 there will be an implicit private collection per organization.
```
cendhu, line 56: Not sure whether we can call both the world state and private state as a collection. As far as I know, the term collection has been used only for private data. By collection, I always thought that the actual data gets into a collection and the hash of the data gets into the block. Whatever data present in the block is called the world state is what I understood. Please correct me if I have misunderstood. 

I remember that one of the usages of the collection is to not leak the data to ordering service by creating a collection with all members. I thought this to be the public collection.

mbwhite: Your understanding of Fabric is correct - and this API is not changing anything about the concept. 
Today for example if you look at putState() and putPrivateState() there are different APIs, but in terms of the gRPC flows to the peer - it's the same low-level flow.. the gRPC level-putstate, takes a string, which is the empty string for the worldstate, and a name for private data.

text/0000-ledger-api.md
```diff
@@ -0,0 +1,380 @@
...
+
+	// with a convience version of
+	Collection org1PrivateData = Ledger.getOrganizationCollection("orgsmspID");
+
+```
+
+There are underlying technical restrictions on some operations. For example you cannot combine private data range queries and updates in a transaction.
+Nor do private data collections support pagination. Should these not be supported, a 'not-supported' error/exception will be thrown.
+
+As there are two essential types of collection - public and private - this is available via an enumerated value. Each languages approach to enumerations is syntactical different but conceptually similar
```
cendhu, line 85: Some of the comment that I mentioned above is applicable here.

mbwhite: It is true that the word collection is being overloaded?  Is there an alternative you can think of?

text/0000-ledger-api.md
```diff
@@ -0,0 +1,380 @@
...
+
+## Example Interfaces
+
+To see how the concepts above can be rendered in code we can walk-through an updated version of the Java Commercial Paper example.
+
+### Putting data
+
+The 'issue' method will create a new business object based on supplied data, and store this within the Ledger.
+
+```java
```
cendhu, line 109: **Proposed Way**
 ```
Collection collection = Ledger.getLedger(ctx).getCollection(Collection.WORLD);
String key = State.makeComposite(CommercialPaper.class.getName(), new String[] { paperNumber });
collection.createState(key, CommercialPaper.serialize(paper));
```
**Old way**
```
String key = CreateCompositeKey(CommercialPaper.class.getName(), new String[] { paperNumber });
PutState(key, CommercialPaper.serialize(paper));
```
I would like to understand how the proposed way is better than the old way of writing contracts. After reading further, I understand that if the user wants to store the paper on a private collection later, just changing one line is adequate. Does it happen often? Isn't the old way look simpler (as we have to type a few characters) and readable? It may be subjective. 

Is the motivation of this proposal to reduce the number of chaincode interfaces to simplify the task of chaincode developers? I am asking this because, with this proposal, the 9 APIs dedicated to the private data would not be needed. Given that, I would like to state the reasoning for defining separate APIs for private data. The following are some of the points copied from an email chain dated July 2017 that decided to use separate APIs for private data. 

- Is private data really part of ledger state? The way we currently manage it (or propose to manage) is outside of the ledger. Its hash is on the ledger, so its hash is part of the state but the actual data is not. So I think we should be very clear to the developers that this is outside of the blockchain. IMO, the best would be to express this intention with completely different APIs than get/putState.  Perhaps stub.get/putPriviateData(...)
- exposing the same APIs for public and private data may be a bit confusing for the users. 
- Also, the use of collections in the public data is not clear as such. But in future, if we find any use of collections in the public data, we would be back to the same problem - how to differentiate public data and private data. This would demand again a change in the APIs.

text/0000-ledger-api.md
cendhu: We are trying to put the following three variants under a single bucket called a collection. I am completely fine with that provided that it does not create any confusion for the developers. 

1. **Private collections**: only a sub-set of channel members are part of this collection. It provides data privacy within a channel. 
2. **Public collections**: everyone in the channel is a member of this collection. This allows us to have different endorsement policy for public data (coarse-granular compared to a key-based endorsement but fine-granular compared to chaincode level endorsement policy). Further, it helps in avoiding the data leak at the ordering service, for example, if the deployment uses a thrid-party ordering service (provided by some cloud vendors). Note that this concept is not explicitly discussed in Fabric but we can create such a collection even now as the only difference compared to a private collection is the member list. We did discuss the concept of the public collection while designing the private data but didn't write it on any docs. In one of the comments above, Dave mentioned about bringing the public collection concept to core fabric. 
3. **World state**: States that are visible to everyone in the channel including the ordering service. This is the state that actually presents in the blocks. We can also term this as blockchain states or ledger data. This includes the hash of the private data which can be accessed by everyone in the channel through `GetPrivateDataHash()`. Further, a few patterns have been introduced by Dave using this new API which makes the developer aware of hash of some private data belonging to a sub-set of organizations is present in the blockchain.  

IMO, public and private collection can be called a collection without the actual differentiation as the underlying concept is the same (these data do not get into blocks) but there is a slight difference in the goal. Given that, we might call the world state as public collection and the other as a private collection. Again, I am not sure whether it would confuse the developers.

davidkhala:
@mbwhite If it is accepted and ready to use, will current ContractAPI be substituted or could be used as a parallel option?

The Ledger API looks more abstract than Contract API. It is good since Contract API is still too close than directly using raw shim stub.

mbwhite:
@davidkhala thanks for the comment! 

LedgerAPI would replace the shim/stub components - so the Contract definitions remain, but the shim/stub wouldn't be the primary API.

This specific RFC is being retired in favour of an updated one that will be published shortly (I introduced it on the community call this week - 27thMay). (by shortly think days)

Think of this updated one as being like LedgerAPI / Contract API ++ :-)

mbwhite:
This RFC has been superseded by https://github.com/hyperledger/fabric-rfcs/pull/28

PR #17, opened 2020-01-07
Clarify RFC final comment period does not require strict majority

denyeart:
Clarify RFC final comment period guidance. Specifically, indicate
that a group of at least three interested maintainers should agree
that an RFC is ready for final comment period. The requirement
for a "majority of maintainers" is an unnecessary burden, since the
final comment period itself allows for objections to be raised. Furthermore
it is unclear what majority means when an RFC impacts multiple
Fabric repositories with different sets of maintainers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,14 +131,18 @@ postpone).
   - For RFCs with lengthy discussion, the motion to FCP is usually preceded by
   a *summary comment* trying to lay out the current state of the discussion and
   major trade-offs/points of disagreement.
-  - Before actually entering FCP, *the majority* of maintainers must sign off;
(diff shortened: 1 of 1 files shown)


jyellick:
+1 'a majority' is too strong simply to get to the final comment period.  Three seems like a fine starting place, we can always adjust as experience informs us.

guoger:
+1

andrew-coleman:
+1

pamandrejko:
+1

denyeart:
Reviewed at Fabric contributor meeting. No disagreements... merging.

PR #23, opened 2020-03-18
LTS release strategy RFC

denyeart:


text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+Administrators of production deployments find LTS releases desirable as they provide confidence that a certain software release will be maintained for an extended period of time.
+
+Specifically, for Hyperledger Fabric, administrators can expect the following of a LTS release:
+
+- Stable and well-tested code
+- API contract stability
+- 3rd digit patch releases for bug and security fixes, delivered for an extended period of time on a regular cadence (or more frequently as needed)
+- Minimal set of feature additions and other changes, reducing the risk of functional regressions and bugs
+
```
pamandrejko: .. that can be applied via an upgrade process that does not require data migration.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+![LTS release diagram](../images/lts_release_diagram.png)
+
+## LTS and channel capabilities
+
+Channel capabilities will remain available for two LTS releases, but may be removed in subsequent LTS releases.  For example, v1.x capabilities such as legacy chaincode will be available through v2.2 (assuming v2.2 is a LTS release), but are expected to be removed before the subsequent LTS release. This provides one full LTS cycle for channels to transition to new capabilities. For example, a production Fabric network may perform a rolling upgrade from v1.4.x nodes to v2.2.x nodes, then enable v2.x channel capabilities to take advantage of the new chaincode lifecycle, and then eventually upgrade to v3.x LTS nodes, where the legacy chaincode lifecycle may not be available.
+
+## LTS and upgrade testing
+
+The Fabric project expects to test single step upgrades (e.g. v2.0 to v2.1), and LTS to next-LTS upgrades (e.g. v1.4.x to v2.2).
+
```
pamandrejko: The statement `Fabric project expects to test..` is a little vague.

You might want to be very clear here and say that in order to upgrade to the next LTS, a customer environment must first be running at the current LTS level. And if they are not there yet, they need to get to the current LTS, before they can upgrade to the next LTS. 

 In other words, they will not be able to go from 1.4.2 to 2.1, if 2.1 was the next LTS release.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+Each major release will eventually have at least one minor release designated by the Fabric maintainers as the LTS release, e.g. v1.4 for v1.x, v2.2 for v2.x (all references to future versions are hypothetical).
+
+After a LTS release is designated, quarterly releases will continue as normal. The Fabric maintainers will decide whether a next release is another minor release, or a new major release. There is no pre-defined timing for next major version. The decision is typically based on semantic versioning considerations, such as whether API changes are needed or older capabilities need to be removed. Other considerations may also apply, for example significant upgrade steps may motivate a shift to a new major version.
+
+If a major release is not delivered for an extended period of time, the maintainers may designate a subsequent minor release as the next LTS release, for example if v3.1 is a LTS release and there is no need to increment to 4.0 for several quarters, the maintainers may decide to designate v3.4 as the next LTS release.
+
```
pamandrejko, line 59: So customers can set a general expectation for planning how often their businesses will want to update, is there any general timeline about how often an LTS will be released? e.g. expect around 1 LTS/per year or no more than 2 LTS per year or something like that?

pamandrejko (changes requested):
Thanks for putting this together Dave. Very helpful to see it all on paper, so to speak.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+Hyperledger Fabric will continue the pattern of quarterly minor releases, e.g. v2.0, v2.1, v2.2, etc.
+
+## LTS release cadence
+
+Because a new major release typically has large new features that may not yet be tried by the user community, and because SDKs may lag in support of the new release, it is not expected that a new major release will immediately be designated as a LTS release, for example, v2.0 is not a LTS release.
+
+Each major release will eventually have at least one minor release designated by the Fabric maintainers as the LTS release, e.g. v1.4 for v1.x, v2.2 for v2.x (all references to future versions are hypothetical).
+
+After a LTS release is designated, quarterly releases will continue as normal. The Fabric maintainers will decide whether a next release is another minor release, or a new major release. There is no pre-defined timing for next major version. The decision is typically based on semantic versioning considerations, such as whether API changes are needed or older capabilities need to be removed. Other considerations may also apply, for example significant upgrade steps may motivate a shift to a new major version.
```
mastersingh24:
```suggestion
After a LTS release is designated, quarterly releases will continue as normal. The Fabric maintainers will decide whether a next release is another minor release, or a new major release. There is no predefined timing for next major version. The decision is typically based on semantic versioning considerations, such as whether API changes are needed or older capabilities need to be removed. Other considerations may also apply, for example significant upgrade steps may motivate a shift to a new major version.

```

mastersingh24: typo

text/0000-lts-release-strategy.md
denyeart: simplified as "that can easily be applied"

text/0000-lts-release-strategy.md
denyeart: The thing is, other upgrade paths typically work, and some users choose to do their own testing and proceed as such.
I've clarified by adding "Other upgrade paths will not be tested and are not guaranteed to work." This should signal that users should proceed at their own risk.

text/0000-lts-release-strategy.md
denyeart: Well, the reason we're going with this approach is that we can't predict the future content and wanted to keep things flexible. Take Fabric CA for example... there is no other release planned and therefore v1.4.x will be the LTS for the foreseeable future. So I don't think we can predict timeframes at this moment.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+If a major release is not delivered for an extended period of time, the maintainers may designate a subsequent minor release as the next LTS release, for example if v3.1 is a LTS release and there is no need to increment to 4.0 for several quarters, the maintainers may decide to designate v3.4 as the next LTS release.
+
+## LTS 3rd digit patch releases
+
+For LTS releases, 3rd digit patch releases will be provided for bug and security fixes at least every three months, but may be more frequent if important fixes become available sooner. Significant new features and other changes will not be included in 3rd digit patch releases, in order to ensure the stability of the LTS release and reduce the risk of functional regressions and bugs.
```
jyellick: It's hard to imagine that there will be no bug or security fixes within 3 months, but it seems a little odd to commit to putting out a new release, even if there is no content.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+It is worth noting that the v1.4 LTS release did not follow the above principles. Had the v1.4 LTS release followed the above principles, the Raft ordering service feature would likely have been introduced in a v1.5 release, and Kafka to Raft migration in a v1.6 release. We mention only to illustrate the above principles.
+
+## LTS release duration
+
+Once a *subsequent* LTS release is designated, users can expect 3rd digit patch releases to address critical bugs on the *prior* LTS release for approximately 9 more months. The maintainers will determine which fixes should to be backported to the latest LTS release, versus which fixes need to be further backported to the prior LTS release. The overlap period is intended to provide users a time window to upgrade their nodes, and any channel capabilities (if required).
```
jyellick: Perhaps adds some language here to stress that in this transition period, we do not intend releases every 3 months, since critical fixes should be quite rare.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+## LTS release diagram
+The following diagram represents the LTS release cadence and duration using hypothetical Fabric releases:
+![LTS release diagram](../images/lts_release_diagram.png)
+
+## LTS and channel capabilities
+
+Channel capabilities will remain available for two LTS releases, but may be removed in subsequent LTS releases.  For example, v1.x capabilities such as legacy chaincode will be available through v2.2 (assuming v2.2 is a LTS release), but are expected to be removed before the subsequent LTS release. This provides one full LTS cycle for channels to transition to new capabilities. For example, a production Fabric network may perform a rolling upgrade from v1.4.x nodes to v2.2.x nodes, then enable v2.x channel capabilities to take advantage of the new chaincode lifecycle, and then eventually upgrade to v3.x LTS nodes, where the legacy chaincode lifecycle may not be available.
```
jyellick: This section has a couple issues for me.

First, I think using the word 'channel capabilities' is a bit overloaded.  Do you mean literally the things in the channel config we call capabilities? Until we having archive and pruning, we do not have the ability to remove any of them.

If this is meant more broadly to be 'features', and not 'capabilities', then this seems like a very firm line to take.  This could be misinterpreted to claim that an esoteric CLI flag from 2.2 must be supported through v3.1 even if usage is low and migration easy.

The idea here seems to be "there is a sane migration path from LTS to LTS", but this language is a bit strong to me.

jyellick (commented):
Generally looks good to me.  A few thoughts though.

mastersingh24 (approved):
Not perfect but good enough.  I think there are a few lines which *could* be interpreted slightly differently, but overall this is a good start.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Administrators of production deployments find LTS releases desirable as they provide confidence that a certain software release will be maintained for an extended period of time.
+
+Specifically, for Hyperledger Fabric, administrators can expect the following of a LTS release:
+
+- Stable and well-tested code
+- API contract stability
```
cendhu, line 32: Are we suggesting that there is a possibility for the contract API to change with each LTS release but not within an LTS? Many of the platforms I have used are very strict about the APIs exposed to the user. Corda has promised that it would never break backward compatibility. Many databases provide that guarantee. Only programming languages have not promised such a compatibility guarantee.  

Even minor changes like moving chaincode to `fabric-chaincode-go`, `fabric-proto-go` repos as well as changing the `logger` have frustrated a few internal users as they have to change the chaincode (good that it was just experimental, not production). I think we made this hard decision considering the code maintenance in the long term but I doubt whether this would happen frequently.

Hence, I am not sure whether we are conveying the wrong message in this bullet point.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+- 3rd digit patch releases for bug and security fixes, delivered for an extended period of time on a regular cadence (or more frequently as needed)
+- Minimal set of feature additions and other changes that can easily be applied, reducing the risk of functional regressions and bugs
```
cendhu, line 34: This comment is applicable to the ledger package. As specified here, we did move a certain minimal set of features to v14 from v2. Though it was minimal, the code changes were moderate to significant including the logics. As a result, it was quite hard at times to move back and forth between v14 and v2, taking additional notes, etc... was bit painful. In this regard, my question is that why can't we just push bug fixes alone to LTS? I am not saying that we cannot manage multiple code versions but if it is required, we will prepare for it accordingly. When a risk is associated with both simple or complex feature, maybe we can restrict to bugs. If required, on a case by case, we can pick the most important minimalistic feature from master to LTS.

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+The following diagram represents the LTS release cadence and duration using hypothetical Fabric releases:
+![LTS release diagram](../images/lts_release_diagram.png)
+
+## LTS and channel capabilities
+
+Channel capabilities will remain available for two LTS releases, but may be removed in subsequent LTS releases.  For example, v1.x capabilities such as legacy chaincode will be available through v2.2 (assuming v2.2 is a LTS release), but are expected to be removed before the subsequent LTS release. This provides one full LTS cycle for channels to transition to new capabilities. For example, a production Fabric network may perform a rolling upgrade from v1.4.x nodes to v2.2.x nodes, then enable v2.x channel capabilities to take advantage of the new chaincode lifecycle, and then eventually upgrade to v3.x LTS nodes, where the legacy chaincode lifecycle may not be available.
+
+## LTS and upgrade testing
+
+The Fabric project expects to test single step upgrades (e.g. v2.0 to v2.1), and LTS to next-LTS upgrades (e.g. v1.4.x to v2.2). Other upgrade paths will not be tested and are not guaranteed to work.
```
cendhu, line 80: I understand that support for LTS-next-LTS upgrade is a must to have but we have TODOs for certain code removal in ledger specifically to support this. This is because the upgrade is more like a peer failure and recovery. As it affects ledger, we have some codes in ledger just to manage upgrade from one LTS to another. I think we need to live with it.

cendhu (commented):
LGTM. Just a few minor comments.

text/0000-lts-release-strategy.md
cendhu: A random philosophical thought of the day (couldn't resist me from posting this): 
I could call Fabric as a decentralized and replicated blockchain database.

1. Every peer is a replica of one other and smart-contract is more like a SQL procedure sitting close to the data. 
2. It is decentralized because there is no central authority. 
3. It is blockchain because it provides non-repudiation, immutability, provenance, etc..

If we hide endorsement knowledge from the client that submits the transaction and lets the peer or SDK library collect it, client would feel as if he/she is using a database such as PostgreSQL (i.e., invoking a stored SQL procedure) or MongoDB. 

APIs exposed are `GetXXX()` and `PutXXX()` for writing smart-contract. It is more like a database APIs. If these APIs are bad, we can change it in the next LTS. Going forward, it would be very nice to keep a stable API to keep our users happy. Users who have experience in using databases would appreciate it.

text/0000-lts-release-strategy.md
denyeart: Changing to:
For LTS releases, 3rd digit patch releases will be provided for bug and security fixes approximately every three months (may be more frequent if important fixes become available sooner, and may be deferred if there are no fixes to publish).

text/0000-lts-release-strategy.md
denyeart: Adding:
3rd digit patch releases should be expected less frequently for prior LTS releases, since only critical fixes will be published.

text/0000-lts-release-strategy.md
denyeart: Changing to:
Features that are related to capabilities will remain available for two LTS releases, but may be removed in subsequent LTS releases.  For example, v1.x legacy chaincode management will be available through v2.2 (assuming v2.2 is a LTS release), but is expected to not be usable in the subsequent LTS release. This provides one full LTS cycle for channels to transition to new capabilities. For example, a production Fabric network may perform a rolling upgrade from v1.4.x nodes to v2.2.x nodes, then enable v2.x channel capabilities to take advantage of the new chaincode lifecycle, and then eventually upgrade to v3.x LTS nodes, where the legacy chaincode lifecycle may not be available.

text/0000-lts-release-strategy.md
denyeart: There is no suggestion that APIs will change across LTS releases. That is not something we want to do. However given the amount of innovation in the blockchain space, we can't predict what changes future maintainers will want to make.

text/0000-lts-release-strategy.md
denyeart: We intend to only backport bug fixes to LTS releases. The last bullet is just a caveat that something small may get backported. For example if operators needed an additional metric in order to operate the LTS release effectively, we wouldn't necessarily block such a small change.

text/0000-lts-release-strategy.md
denyeart: @cendhu  I'm not understanding the main point here? What is the takeaway?

text/0000-lts-release-strategy.md
cendhu: > I think there are a few lines which could be interpreted slightly differently

I am one of them who interpreted differently (like Gari mentioned). 

What does this API Contract Stability refer too? Code stability? or existing API changes? I understood it as existing APIs wouldn't change within an LTS.

I agree that appending APIs for a new feature is no issue. Every databases/products do that.

text/0000-lts-release-strategy.md
cendhu: This was just a comment I had on LTS-next-LTS rolling upgrade. 

Whenever we add features to the ledger code or remove code or refactor code, we always need to keep LTS-next-LTS in mind. This could introduce bugs as our LTS timeline is long and we might change the code a lot in between. Takeaway can be that we need to very extra cautious while changing the ledger code.

text/0000-lts-release-strategy.md
cendhu: IMO, calling a `metric addition` as a feature is misleading. If we intend to backport only bug fixes, I am not sure whey we have this confusing bullet. Anyway, this is just my opinion and feel free to ignore it.

text/0000-lts-release-strategy.md
denyeart: Agreed!

text/0000-lts-release-strategy.md
```diff
@@ -0,0 +1,90 @@
...
+
+It is worth noting that the v1.4 LTS release did not follow the above principles. Had the v1.4 LTS release followed the above principles, the Raft ordering service feature would likely have been introduced in a v1.5 release, and Kafka to Raft migration in a v1.6 release. We mention only to illustrate the above principles.
+
+## LTS release duration
+
+Once a *subsequent* LTS release is designated, users can expect 3rd digit patch releases to address critical bugs on the *prior* LTS release for approximately 9 more months. The maintainers will determine which fixes should to be backported to the latest LTS release, versus which fixes need to be further backported to the prior LTS release. 3rd digit patch releases should be expected less frequently for prior LTS releases, since only critical fixes will be published. The overlap period is intended to provide users a time window to upgrade their nodes, and any channel capabilities (if required).
```
yeasy, line 68: Somehow, 9 month is a little shorter as "long term".
Typically it's several years for most existing software from https://en.wikipedia.org/wiki/Long-term_support.
I think it should make sense to extend it to 12 month at least.

lindluni: 9 months is the time for which two LTS releases will live simultaneously. This is saying once 2.x is marked as LTS we will continue delivering bug fixes to 1.4 for 9 months. It doesn't speak to length of time 2.x will actually be LTS for, that could be several years, and then once we deliver a new LTS after 2.x, then 2.x will continue to receive bug fixes for 9 months, while living simultaneously with the LTS after 2.x

yeasy: Make sense!
Besides, it would be better describe the interval between two LTS versions, too.

jyellick (approved):

denyeart:

jmason90:
the prior LTS release support for 9 months when the new LTS is announced is a short time frame compared to other enterprise products. If you have implemented production Fabric frameworks in a network on a prior LTS updating in 9 months may be a real challenge, depending on the quality of the migration tools provided.  Also, regulations may require running only LTS versions which can be a problem. Many other enterprise products provide a 12 month prior LTS support window.

PR #24, opened 2020-03-20
Channel participation API without system channel

tock-ibm:


jyellick:
This generally looks good to me.  I'd like to see if any other reviews have any higher level objections before doing a thorough review though.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+# Summary
+[summary]: #summary
+
+Currently in order to create a channel one needs to issue a config transaction to the system channel. This has 
+privacy, scalability and operational disadvantages, since the system channel (and the orderers running it) is aware of 
+all channels and of all channel members (at creation time). In this feature we propose to expose a 
+"channel participation" API that would allow a local orderer administrator to join and leave a channel, as well as 
+to list all the channels that the local orderer is part of. This allows a Fabric network to be operated without the use 
+of a system channel, which improve the privacy and scalability of the ordering service and the network as a whole.
```
mastersingh24:
```suggestion
of a system channel, which improves the privacy and scalability of the ordering service and the network as a whole. 
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+of a system channel, which improve the privacy and scalability of the ordering service and the network as a whole. 
+
+
+# Motivation
+[motivation]: #motivation
+
+Creating channels by issuing a transaction to the system channel presents several disadvantages.
+
+## Privacy problems
+A system channel requires that all ordering service nodes (OSNs) be aware of all channels. As Fabric depricated the
```
mastersingh24:
```suggestion
A system channel requires that all ordering service nodes (OSNs) be aware of all channels. As Fabric deprecated the 
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+Creating channels by issuing a transaction to the system channel presents several disadvantages.
+
+## Privacy problems
+A system channel requires that all ordering service nodes (OSNs) be aware of all channels. As Fabric depricated the 
+Kafka-based ordering service and moved to Raft (and later is planned to introduce BFT consensus), it is no longer
```
mastersingh24:
```suggestion
Kafka-based ordering service and moved to Raft, it is no longer 
```

No need to mention any future stuff here IMHO

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+There are workarounds that reduce the privacy problem, but not eliminate it. For example, using the system channel it 
+is possible to create an application channel with only a single org, and then add additional orgs by updating the 
+channel itself. However, these type of solutions are cumbersome and only partial.
+
+## Scalability problems
+All OSNs are members of the system channel, which creates a scalability problem in a large scale Fabric network. When 
+the number of channels increases, Raft allows us do decouple the consenters sets of the different application channels, 
+achieving linear horizontal scalability in the number of channels. That is, more resource can be added as the number of 
+channels increase. However, the system channel is an exception: decoupling application channel consenters sets as
```
mastersingh24:
```suggestion
channels increase. However, the system channel is an exception; decoupling application channel consenters sets as 
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+Moreover, in order to start a new orderer and have it join an existing channel (perform 'on-boarding'), a new orderer 
+starts by scanning the system channel in order to discover existing channels and their membership. As the number of 
+channels increase, this scan prolongs the process of adding a new orderer.  
+    
+## Operational problems
+There are several operational problems that affect large scale deployment and hinder efficient management of cloud 
+environments using automatic tooling.
+     
+Channel creation (using the system channel) utilizes information contained in the system channel - e.g. MSP information 
+of the organization - in order to apply it onto the newly created channel. This create the convoluted situation in
```
mastersingh24:
```suggestion
of the organization - in order to apply it to the newly created channel. This creates the convoluted situation in 
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+which an organization that wants to update its details in the consortium must coordinate that action with third parties 
+that have write privileges to the system channel.
+
+There is no good way to list all the channels that an orderer is a member of. 
+
+There is no good way to dispose of the resources associated with a channel, once an orderer is no longer a member of 
+the consenters set of a that given channel.
+
+In many cloud environments, it is necessary to restart an orderer in order for it to discover immediately that it has 
+been added to a channel it was not a part of previously, because otherwise detection takes minutes.
```
mastersingh24: These should probably be a bulleted list for better readability

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+the consenters set of a that given channel.
+
+In many cloud environments, it is necessary to restart an orderer in order for it to discover immediately that it has 
+been added to a channel it was not a part of previously, because otherwise detection takes minutes. 
+
+   
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+We introduce a new mode of channel participation management that does not depend on the existence of a system channel.
```
mastersingh24:
```suggestion
We introduce a new mode of channel management that does not depend on the existence of a system channel.
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+We introduce the ability to start an OSN process without any channels. The OSN exposes a new **"channel participation" 
+API** that allows the operator to join and leave channels on that specific OSN.   
+
+A **"local orderer channel participation administrator"** is a role that is associated with the operation and
```
mastersingh24, line 78: Do we really need to be fine-grained here or is a local orderer admin fine (similar to a local peer admin)?

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+A **"local orderer channel participation administrator"** is a role that is associated with the operation and 
+administration of an OSN. The identity associated with this role is able to invoke a new **"channel participation" 
+API**, and perform the following commands:
+
+- **Join** this OSN to a new or an existing channel,
+- **Remove** this OSN from an existing channel,
+- **List** the channels this OSN is a member of.   
+
+The new mode of channel participation management is meant to be incompatible with channel creation using a system 
+channel, in the sense that if the system channel exists, the `Join` operation will not be supported.
```
mastersingh24, line 87: Does this have to be the case?  Seems like it should be possible to support both ... but I'll read about "migration" first

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+API**, and perform the following commands:
+
+- **Join** this OSN to a new or an existing channel,
+- **Remove** this OSN from an existing channel,
+- **List** the channels this OSN is a member of.   
+
+The new mode of channel participation management is meant to be incompatible with channel creation using a system 
+channel, in the sense that if the system channel exists, the `Join` operation will not be supported. 
+However, there is a transition path from a network operated with a system channel to a network that is operated without
+one. In addition, it should be possible to start OSN processes without channels, create a system channel with the
```
mastersingh24:
```
... it should be possible to start OSN processes without channels, create a system channel with the
```
This is an important point ... let's make sure it does not get lost

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+  ```Join channel-name channel-config-block```
+  
+  or
+  
+  `POST /channels/channel-name` with body containing `channel-config-block`
+  
+The body contains a config block that contains all the details about the channel. 
+
+If the OSN is joining a new channel, the config block is the genesis block of the channel, generated by the 
+`configtxgen` tool, as described below. It looks like the first block of an
```
mastersingh24:
```suggestion
If the OSN is joining a new channel, the config block is the genesis block of the channel. It looks like the first block of an 
```
we don't need to say how the genesis block is created here as it's not relevant

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+The OSN will try to find itself in the consenters set, and determine if it is a follower or a member of the cluster
+executing consensus. 
+
+If it is not in the set, it will continue as a **follower**, and start to pull blocks from other OSNs. It will continue
+to do so until it catches up with the config block received in the invocation. From then on, it will continue to pull blocks, 
+but will inspect each incoming config block and check whether it is included in the consenters set. If it is, it will 
+start the runtime components that execute consensus, thus joining the cluster. 
+
+If the OSN is in the consenters set, it will continue as **member**, first pulling blocks from other OSMs util it catches
```
mastersingh24:
```suggestion
If the OSN is in the consenters set, it will continue as **member**, first pulling blocks from other OSNs util it catches
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+up with the given config block, and then starting the runtime components that execute consensus, thus joining the cluster.
+
+Note that joining an OSN as a member means that the channel was updated to include that OSN in the consenters set before it
+was joined by the participation API. Therefore, during the catch-up process and until the new OSN starts executing 
+consensus, the cluster is a state of reduced fault tolerance. For long chains this might be a problem, since catch-up
+can take a long time. The solution is to join the new OSN as a follower. After it catches up with the cluster the 
+channel admin can update the channel config and include it in the consenters set.
```
mastersingh24, line 158:
```
The solution is to join the new OSN as a follower. After it catches up with the cluster the 
channel admin can update the channel config and include it in the consenters set.
```
This is actually a new feature as well, correct?

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+The new OSN can pull blocks from other OSNs only if it has authorization to do so. A minimal requirement is that its 
+owning org will be defined in the channel configuration, and that the signature of the new OSN on the `Deliver`
```
mastersingh24:
```suggestion
owning organization will be defined in the channel configuration, and that the signature of the new OSN on the `Deliver` 
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+Note that joining an OSN as a member means that the channel was updated to include that OSN in the consenters set before it
+was joined by the participation API. Therefore, during the catch-up process and until the new OSN starts executing 
+consensus, the cluster is a state of reduced fault tolerance. For long chains this might be a problem, since catch-up
+can take a long time. The solution is to join the new OSN as a follower. After it catches up with the cluster the 
+channel admin can update the channel config and include it in the consenters set.
+
+The new OSN can pull blocks from other OSNs only if it has authorization to do so. A minimal requirement is that its 
+owning org will be defined in the channel configuration, and that the signature of the new OSN on the `Deliver` 
+request satisfy the `/Channel/Readers` policy.
```
mastersingh24:
```suggestion
request satisfies the `/Channel/Readers` policy.
```

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+The command refers to a profile `SampleRaftChannelCreation` in a `configtx.yaml` file, as usual. The YAML file 
+will point to the file-system path of the MSP configuration of each org, for application as well as ordering orgs. 
+In addition, it will point to the path of the certificates of the consenters. Generating a genesis block for an 
+application channel means that:
+- the `configtx.yaml` **does not** include a `Consortiums` config group
+- the `configtx.yaml` **must include** a `Channel/Application` config group
+ 
+
+The minimally required elements should include enough details for the OSN to achieve consensus and allow further 
+configuration of all the config groups.
```
mastersingh24: You should not mention `configtxgen` here ... it's just a tool.
I think the important thing to detail here is what needs to be in a channel config and the fact that there is no longer a consortium defined.  This means that you can no longer simply reference MSP content but must explicitly include it in the channel create / config transaction / genesis block.

At least I think that's the case

mastersingh24 (changes requested):
Overall, I'm really happy about the general content and idea here.

mastersingh24:
> This generally looks good to me. I'd like to see if any other reviews have any higher level objections before doing a thorough review though.

No objections from me on the overall concept.  Can you review the actual APIs in detail?

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+If the channel already exists, the call will fail.
+
+If the system channel exists, the call will fail. 
+          
+The OSN receiving this call will inspect the block and conduct some sanity checks in order to make sure that the block 
+is well-formed.
+
+The OSN will try to find itself in the consenters set, and determine if it is a follower or a member of the cluster
+executing consensus.
```
sykesm, line 128: The roles of "follower" and "member" should be more formally defined. It's referenced here with assumptions about the reader's definition of those words which may not match the intent.

I realize the term was emphasized earlier with an implied definition but I believe an explicit definition of the concept would make this more straightforward.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+
+Generating the channel genesis block follows the same rationale as generating the system channel genesis block, when 
+bootstrapping a new Fabric network. 
+
+The command refers to a profile `SampleRaftChannelCreation` in a `configtx.yaml` file, as usual. The YAML file 
+will point to the file-system path of the MSP configuration of each org, for application as well as ordering orgs. 
+In addition, it will point to the path of the certificates of the consenters. Generating a genesis block for an 
+application channel means that:
+- the `configtx.yaml` **does not** include a `Consortiums` config group
+- the `configtx.yaml` **must include** a `Channel/Application` config group
```
sykesm: `configtx.yaml` should not be explicitly referenced here. It's a (poor) means to an end.

jyellick:
> > This generally looks good to me. I'd like to see if any other reviews have any higher level objections before doing a thorough review though.
> 
> No objections from me on the overall concept. Can you review the actual APIs in detail?

Perhaps I should have been a bit clearer.  Prior to publication, I reviewed this quite a bit and after a few rounds of communication was pretty happy with it.  Since not much has changed I was planning to wait to do a thorough re-review once feedback from the other reviewers was incorporated.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+but some details (i.e. exact formats, responses) are omitted for clarity.
+   
+### Joining a channel
+A channel is joined by providing the OSN with a config block:
+
+  ```Join channel-name channel-config-block```
+  
+  or
+  
+  `POST /channels/channel-name` with body containing `channel-config-block`
```
sykesm, line 109: It's a detail but can we add a small section on API versioning please? Resource versioning? Accept headers?

Or are we simply using REST as nomenclature instead of the target implementation pattern?

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+- The channel participation commands "List" and "Delete" (of system channel only) will continue to function.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## The API layer
+We intend to extend the "operations API" (package `core/operations`) by adding a handler for channel participation 
+commands, for path `/channels` as described above. These commands will only be enabled if specified explicitly in the `orderer.yaml` 
+configuration file. These commands will only be available in the orderer and not in the peer.
+
```
sykesm: There is no "operations API". There's an embedded server that this can hook into to provide the orderer API. We must keep them separate.

The implementation also does *not* belong in core/operations.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+- The channel participation command "Join" will no longer accept commands, and channel creation can only 
+  be done using a config transaction on the system channel.
+- The channel participation commands "List" and "Delete" (of system channel only) will continue to function.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## The API layer
+We intend to extend the "operations API" (package `core/operations`) by adding a handler for channel participation 
+commands, for path `/channels` as described above. These commands will only be enabled if specified explicitly in the `orderer.yaml`
```
sykesm: `/channels` should probably be scoped to your API instead of hung off the root.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,497 @@
...
+- https://jira.hyperledger.org/browse/FAB-15709
+- https://jira.hyperledger.org/browse/FAB-15710
+- https://jira.hyperledger.org/browse/FAB-15711
+- https://jira.hyperledger.org/browse/FAB-15712
+- https://jira.hyperledger.org/browse/FAB-15713
+
+These will be combined in the development plan once the RFC is approved.
+
+## Authorization and Authentication
+There is an ongoing effort to improve the AA mechanisms of the "operations" API.
```
sykesm, line 503: Where is that effort taking place? I'm very interested as it has larger impacts and is explicitly an area that I have opinions on.

sykesm (dismissed):

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Yes, this is a new feature.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: For the sake of generality, I think it is useful to formally define this role. 
Our current design choice is to have this role will use the same AA mechanism as the "operations API" admin (see Security section). However it is conceivable that these roles would be separated some day.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: "Mixed mode" is a problem... things can stop working in ugly ways, e.g. on-boarding. There are solutions of course, but they are complicated and costly. Therefore, for simplicity, I think we should avoid that and encourage users to use either one or the other. 

In addition, moving from system-channel mode to participation-api style is easy (delete system-channel on all OSNs). The reverse, going back to system-channel style once there are already app-channels  is complicated. Here too we choose simplicity and do not allow the reverse path. The only exception is bootstrap: starting from empty OSNs and creating the system channel.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: I split the paragraph to make this more prominent.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Done

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Done

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Done

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: I demoted the reference to configtxgen to a single example, rather then making it a prescription for generating the genesis block. I also described the changes w.r.t. a consortium & consortiums.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Done, I demoted that to an example.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: I was planing (hoping...) to go into the fine-grain details when I write the Jira item dedicated to the REST API. Here it is meant to be just a sketch, in order not to confuse the reader with too many details.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Gari is doing some exploratory work on that, I believe. 
@mastersingh24

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: You mean something like this?

`channel-participation/channels`

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Most of the implementation will be under orderer/... 
Some tiny bits that hook into the HTTP server might reside elsewhere (not sure where though).

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Agreed. I will improve the presentation.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Done

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+
+It should be possible to start OSN processes without channels, create a system channel with the
+channel participation API, and continue to operate the network in the usual way - creating channels through the system 
+channel. 
+
+"Mixed mode" management, that is, using the system channel to create channels on some OSNs and the 
+participation API to create channels on other OSNs is not supported and highly discouraged.  
+
+## Channel participation commands 
+In this section channel participation commands are described in a high level. The commands are presented in REST style
```
yeasy, line 99: Any specific reason to use REST rather than the existing gRPC style?

tock-ibm: We want to make it easier for cloud operators to manage the OSN channel participation. This way they won't need additional tools that implement gRPC, and can use a browser, for example.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+An OSN operating as a cluster follower is a new feature introduced by this work.
+
+#### Two ways to join     
+When an OSN joins a channel as a **follower**, it start to pull blocks from other OSNs. It will continue
+to do so until it **catches up** with the config block received in the invocation. From then on, it will continue to pull 
+blocks, but will inspect each incoming config block and check whether it is included in the consenters set. 
+When channel admins add it to the consenters set, it will detect that and start the runtime components that execute 
+consensus, thus becoming a member the cluster. 
+
+When an OSN joins a channel as a **member**, it first pulls blocks from other OSNs util it **catches
```
jyellick: Very minor typo -- "util" instead of "until"

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+The channel participation commands will only be enabled if specified explicitly in the `orderer.yaml` 
+configuration file. These commands will only be available in the orderer and not in the peer.
+
+The _local orderer participation admin_  role will have the same authentication and authorization
+as the _operations admin_ (i.e. the same mutual TLS certificate). We choose this option because it is the simplest 
+to implement at this time. This is expected to change in the future (see below).
+
+## Participation management
+We intend to implement the channel participation management functionality in a new package 
+(`orderer/common/channelparticipation`). In this package a new class will keep a map of channel names and their status,
```
jyellick, line 366: This follows the existing patterns of the orderer codebase, but we have been moving towards placing new code not to be consumed by external parties as a library (though this is essentially all of Fabric) into `./internal`.  I'd welcome opinions from other maintainers, but either way, I don't see the location of this package as being significant enough to hold up the RFC.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+and will be in charge of coordinating the operations of following or joining a channel as a member. 
+Every channel that needs to follow or catch-up on the cluster will have a dedicated goroutine. When the channel reaches
+a state where it can join the cluster, the respective methods in `orderer/common/multichannel/registrar.go` will be
```
jyellick, line 369: Being the primary author of much of the multichannel code, I would not get too hung up on preserving our existing channel startup and management structure.  If this code needs to be gutted and replaced, all the better.  Regardless, this seems like an implementation detail outside the required scope of the RFC, so not worth holding anything up over.

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+
+## Participation management
+We intend to implement the channel participation management functionality in a new package 
+(`orderer/common/channelparticipation`). In this package a new class will keep a map of channel names and their status,
+and will be in charge of coordinating the operations of following or joining a channel as a member. 
+Every channel that needs to follow or catch-up on the cluster will have a dedicated goroutine. When the channel reaches
+a state where it can join the cluster, the respective methods in `orderer/common/multichannel/registrar.go` will be 
+invoked in order to start the chain runtime.
+
+When a channel is first joined with a config block, the channel's ledger folder is created, and inside it the config
```
jyellick, line 372: This seems like it might be a bit racy -- we could crash after creating the ledger folder but before writing the block.  Perhaps it would be better to reverse the order and locate the block outside the ledger dir so that we could resume creating the ledger dir if interrupted?

text/0000-channel-participation-api-without-system-channel.md
```diff
@@ -0,0 +1,536 @@
...
+Listing the channel or channel is achieved by querying the channel map in the `.../channelparticipation` package 
+implementation.
+
+The interface of executing channel participation commands is expected to stay stable even if the  authentication and 
+authorization mechanisms change, or even if this API is separated on to a different service endpoint.
+
+## Boot sequence of Orderer
+In the boot sequence of the orderer (`orderer/server/main.go`) we will have to check whether the orderer is working
+with or without the system channel, and then start either the current on-boarding code (eventually calling 
+`orderer/common/cluster/Replicator`) or the new channel participation component.
```
jyellick: I was initially thrown for a loop here, but I think you mean the component which implements the following/joining go routines, not the the participation service which still allows for the deletion of the system channel and listing of channels.  Could be clarified.

jyellick (approved):
As far as I'm concerned, this is good to go.  I've added just a few comments, but I don't think it's anything critical mostly typos, and some implementation details.

text/0000-channel-participation-api-without-system-channel.md
mastersingh24: I'm a bit torn on this one ... I tend to think we should probably stick with gRPC here.  But I will not hold things up ... the APIs generally look ok and it's not hard to translate these to protobuf defs.

mastersingh24 (approved):

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: You are correct, this needs to be done atomically. One way to do this is to create a temporary folder with all the resources and then rename it into place. Renaming is atomic so if we fail before the rename nothing happens.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: I clarified the text.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: Fixed.

text/0000-channel-participation-api-without-system-channel.md
tock-ibm: OK, I'll take the liberty to streamline that code if necessary.

jyellick (approved):

denyeart:

SamyDgz:
@SamyDgz

PR #25, opened 2020-03-20
[FAB-17628] Create RFC proposal for Config Transaction Library

caod123:


text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+This package has high coverage of core features on both a unit level and integration test level. Example usage tests are also
+included in the godocs.
+
+# Dependencies
+[dependencies]: #dependencies
+
+In an effort to make this library as lightweight as possible, we've reduced the number of required dependencies to just a handful including
+the golang standard library and the following `hyperledger` packages and components:
+- `fabric-protos-go`
+- `fabric/common/policydsl`
```
denyeart, line 77: Clarify whether the new repo will depend on Fabric for this package, or would the package just be copied into the new repo?

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+[alternatives]: #alternatives
+
+It is under the impression that eventually it will become too onerous to maintain current workarounds for channel configuration management
+as new features are added around channel config. Rather than having to manually update bits and pieces of workaround
+code, consumption of this library will become a consistent method for adopting new features.
+
+# Prior art
+[prior-art]: #prior-art
+
+All good golang dependencies are built up as a consumable, consistent, well-documented with examples, libraries with effective
```
denyeart: This sentence was difficult for me to parse at first, probably because it jumps right in and mentions 'dependencies', while this library has not explicitly been called a 'dependency' above.

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+
+This feature will implement a consumable standalone library that supports generating configuration envelopes for 
+operations including application and system channel creation, channel configuration updates, and attaching 
+endorsing signatures to generated envelopes.
+
+# Motivation
+[motivation]: #motivation
+
+The motivation for this work is to provide a consistent API that is actually intended for consumption by external
+users for any operations involving channel config transactions. Currently `configtxgen` is available and frequently
```
denyeart: Consumption by 'external users' makes it sound like end users could use it. I like how you discussed in the Epic the example of somebody using the library to build a CLI... saying something similar here would help people understand the intended usage of the library.

mastersingh24:
```suggestion
clients for any operations involving channel config transactions. Currently `configtxgen` is available and frequently
```
not sure if `clients` or `applications` makes the most sense here

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+misused as a tool for these operations, but it was never intended for production usage. Much of the existing process for
+generating configuration updates is both tedious and error prone involving decoding a config transaction to JSON
+with the protolator tool and manually modifying fields prior to re-encoding back to proto and submitting for update.
```
mastersingh24:
```suggestion
with the protolator tool and manually modifying fields prior to re-encoding to a protobuf and submitting for update.
```
or

```suggestion
with the protolator tool and manually modifying fields prior to encoding back to a protobuf and submitting for update.
```

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+API documentation already exists in the godocs for this package and can be referenced [here](https://godoc.org/github.com/hyperledger/fabric/pkg/config).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Currently many consumers already have a stable method for updating configuration transactions and may be
```
mastersingh24:
```suggestion
Many potential consumers may already have a stable method for updating configuration transactions and may be
```

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+Work on this consumable library has already begun in [`fabric/pkg/config`](https://github.com/hyperledger/fabric/tree/master/pkg/config) 
+and tracked in [FAB-17628](https://jira.hyperledger.org/browse/FAB-17628). Example usage has been included in the 
+[godocs](https://godoc.org/github.com/hyperledger/fabric/pkg/config#pkg-examples) for this package and can be referenced for external tooling.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+API documentation already exists in the godocs for this package and can be referenced [here](https://godoc.org/github.com/hyperledger/fabric/pkg/config).
+
+# Drawbacks
```
mastersingh24, line 44: The other potential drawback is that we are only going to provide the library in Golang

mastersingh24: Of course I don't think this is a drawback as you can build CLIs or even a Restful / web application using Golang

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+[drawbacks]: #drawbacks
+
+Currently many consumers already have a stable method for updating configuration transactions and may be
+unwilling to make amendments to their processes by consuming this library. It could be argued that even if this library
+exists, no external users already using `configtxgen` would want to switch to consuming it. 
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+It is under the impression that eventually it will become too onerous to maintain current workarounds for channel configuration management
```
mastersingh24:
```suggestion
Eventually it will become too onerous to maintain current workarounds for channel configuration management
```

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+as new features are added around channel config. Rather than having to manually update bits and pieces of workaround
+code, consumption of this library will become a consistent method for adopting new features.
+
+# Prior art
+[prior-art]: #prior-art
+
+All good golang dependencies are built up as a consumable, consistent, well-documented with examples, libraries with effective
+APIs and universal coding standards.
```
mastersingh24:
```suggestion
Consumable Golang packages have consistent and effective APIs, are well-documented, leverage universal coding standards, and provide useful examples of common usage.
```

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,80 @@
...
+[testing]: #testing
+
+This package has high coverage of core features on both a unit level and integration test level. Example usage tests are also
+included in the godocs.
+
+# Dependencies
+[dependencies]: #dependencies
+
+In an effort to make this library as lightweight as possible, we've reduced the number of required dependencies to just a handful including
+the golang standard library and the following `hyperledger` packages and components:
```
mastersingh24:
```suggestion
the Golang standard library and the following `hyperledger` packages and components:
```
nit but I think it's more correct to capitalize Golang

mastersingh24 (commented):
I like the content overall.  Nice work.  Suggestions mostly focused on making things a bit more clear and some nit-picky writing style stuff

text/0000-config-transaction-library.md
caod123: By dependencies this really just means imported packages. I'll add that to the sentence.

text/0000-config-transaction-library.md
caod123: Noted, will mention that we may eventually support TinyGo compilation to web assembly since that also addresses that a bit.

tock-ibm:
Will this library be able to generate a channel genesis block? Can it be used to bootstrap a network, or, in light of PR #24 to bootstrap an application channel?

caod123:
> Will this library be able to generate a channel genesis block? Can it be used to bootstrap a network, or, in light of PR #24 to bootstrap an application channel?

Currently the library does not support creating a system channel create transaction, but the next track of work after the initial application channel work will focus on system channel. It was pushed back in priority because we're under the impression that system channel is currently in the works of being removed, but we would still like to support it for our API in the meantime.

The library will not be intended for bootstrapping a network and will only be used to build channel create transactions for application channels (and later system channels), as well as to modify config transactions and produce the desired config update envelope. There was some discussion around whether we'd like to eventually support something similar for peer configuration but that is far outside of the scope of this library.

There will be some considerations after the work in #24 is completed though, but for now we're creating the library as if #24 does not exist.

denyeart (approved):
LGTM now

mastersingh24:
Thx @caod123 ... LGTM

mastersingh24 (approved):

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,84 @@
...
+This package has high coverage of core features on both a unit level and integration test level. Example usage tests are also
+included in the godocs.
+
+# Dependencies
+[dependencies]: #dependencies
+
+In an effort to make this library as lightweight as possible, we've reduced the number of required dependencies to just a handful including
+the Golang standard library and the following imported `hyperledger` packages and components:
+- `fabric-protos-go`
+- `fabric/common/policydsl`
```
cendhu, line 77: We also need to consider the imports made by the test files. Currently, the following are imported by the test 

- "github.com/hyperledger/fabric/common/tools/protolator"
- "github.com/hyperledger/fabric/common/tools/protolator/protoext/commonext"
- "github.com/hyperledger/fabric/common/tools/protolator/protoext/ordererext"
- "github.com/hyperledger/fabric/common/tools/protolator/protoext/peerext"

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+As part of this RFC process we would like to also move the config package code we've already started on in `hyperledger/fabric/pkg/config` to
```
cendhu: For the following reasons, I think it's good to keep the `config` pkg in a separate repo. 

1. config pkg is a library used by _clients_ and neither by _peer_ nor _orderer_ (i assume this from the RFC)
2. the overall goal is to keep the fabric code minimal
3. we already have taken code out of fabric and created `fabric-proto-go` and `fabric-chaincode-go` repos. We do import them back into fabric repo. Even if we put `config` pkg in a separate repo, I assume we won't even import it into Fabric.

Due to these reasons, I think it might be good to have a separate repo. I am not sure whether `hyperledger/fabric-config` alone warrant a repo. Hence, we can also have a repo called `hyperledger/fabric-library` we can then add config library and other future libraries such as 
- `block_decoder` -- a library that can aid the client to open and inspect a block or a transaction
-  `snapshot` -- a library to process the snapshot file, which is created using ledger checkpoint, at the client side for creating a local database for analytics
- `peer-config` -- a library to create a peer configuration tx to dynamically change the internal peer configuration specified in the `core.yaml `without employing the current `stop-reconfigure-start` approach (assumption: we introduce a module at the peer to process local configuration tx to update these config values). 

I just speculate that these libraries might be needed in the longer-term when people want to build a lot of addon on top of Fabric. Hence, a folder called `channelconfig` under `hyperledger/fabric-library` might be a suitable one IMO.

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,84 @@
...
+This CLI tool could then pass these parameters down to the config package for generating a config update envelope with the computed
+read and write sets. Currently `configtxgen` is available and frequently misused as a tool for these operations, but it was never intended for
+production usage. Much of the existing process for generating configuration updates is both tedious and error prone involving
+decoding a config transaction to JSON with the protolator tool and manually modifying fields prior to encoding back to a protobuf
+and submitting for update.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Work on this consumable library has already begun in [`fabric/pkg/config`](https://github.com/hyperledger/fabric/tree/master/pkg/config)
```
cendhu: `config` is too generic IMO. We have `peer config`, `gossip config`, `ledger config`. The config values present in `core.yaml` are retrieved and these configs are constructed during the peer startup and passed to each initializer accordingly.

May be `channelconfig` or `configtx` can be used as it seems to match the description of the library.

> a consumable standalone library that supports generating configuration envelopes for operations including application and system channel creation, channel configuration updates, and attaching endorsing signatures to generated envelopes.

If there is some better name. it can be used as well but `config` looks like disinformation to me.

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,84 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+Work on this consumable library has already begun in [`fabric/pkg/config`](https://github.com/hyperledger/fabric/tree/master/pkg/config) 
+and tracked in [FAB-17628](https://jira.hyperledger.org/browse/FAB-17628). Example usage has been included in the 
+[godocs](https://godoc.org/github.com/hyperledger/fabric/pkg/config#pkg-examples) for this package and can be referenced for external tooling.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+API documentation already exists in the godocs for this package and can be referenced [here](https://godoc.org/github.com/hyperledger/fabric/pkg/config).
```
cendhu: Danny, the below comments are already discussed in the code review. I am just conveying my realization. 

_The client usage flow looks like a state-machine to me. Every API call can result in an `error` and every error result in a `panic`. The user cannot move from one state to another without passing certain criteria. Using these APIs, I am sure that the client may not submit any invalid channel config tx to orderer and it would remove the need for any code that validates the received tx at the orderer node (kidding, still we cannot remove it)._

 I think, for a general-purpose library such as golang `map`, `delete()` not returning error might be okay. After some thought (and some discussions with some maintainers), I realized that this is more of a special purpose library and hence, the `error` returns & state-machine approach are fine. 

We still need to think about whether we should return an `error` or [error type](https://golang.org/doc/effective_go.html#errors) to the library user.

text/0000-config-transaction-library.md
caod123: I'll throw these all in under `fabric/common/tools/protolator` as the overarching component to save a couple lines

text/0000-config-transaction-library.md
caod123: I agree, I'm open to the name of the repo and would be interested in hearing other's thoughts on where it should live whether that's `fabric-libraries` or some other name

text/0000-config-transaction-library.md
caod123: `configtx` has already been decided upon as the name of the package. That rename will be part of moving it out of fabric but the initial package name is `pkg/config` hence why it's listed here as such.

text/0000-config-transaction-library.md
caod123: Agreed, will have that conversation with the config squad as we wrap up on most of the feature work and move on to cleanup/improvements.

text/0000-config-transaction-library.md
caod123: Although not explicitly mentioned as part of this RFC, do note that protolator will likely also move alongside the config package and live as a consumable library as well (used both by config package and fabric)

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,85 @@
...
+production usage. Much of the existing process for generating configuration updates is both tedious and error prone involving
+decoding a config transaction to JSON with the protolator tool and manually modifying fields prior to encoding back to a protobuf
+and submitting for update.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Work on this consumable library has already begun in [`fabric/pkg/config`](https://github.com/hyperledger/fabric/tree/master/pkg/config) 
+and tracked in [FAB-17628](https://jira.hyperledger.org/browse/FAB-17628). Example usage has been included in the 
+[godocs](https://godoc.org/github.com/hyperledger/fabric/pkg/config#pkg-examples) for this package and can be referenced for external tooling.
```
denyeart: the link should be 
https://godoc.org/github.com/hyperledger/fabric/pkg/configtx

jyellick: I certainly understand the desire not to duplicate documentation, and to have to maintain it in two places, but a link to dynamically generated content, untied to any particular commit, version, etc. makes for a very poor historical document.

I realize this work was ongoing when we were still finalizing the RFC process, so we probably shouldn't view this one as a model moving forward anyway, but if the goal of an RFC is to agree on a path forward, pointing to "the current state of the code" doesn't exactly make sense.

Since this one is straddling the boundary between the RFC process, I'm fine as is, but I just wanted to go on record that linking to go-doc shouldn't be a model for future RFCs, and that instead, I would expect to see much more substantial details in the RFC itself.

jyellick (approved):

text/0000-config-transaction-library.md
mastersingh24: We have https://github.com/hyperledger/fabric-lib-go which was for common libraries shred across Fabric projects (specifically fabric and fabric-ca).  I do think this should be it's own repository.   While it is possible to have multiple, independently versioned modules in the same repository, it is a bit tricky if not done right.

text/0000-config-transaction-library.md
caod123: That makes a lot of sense, will note for future RFCs. Thanks!

text/0000-config-transaction-library.md
caod123: I'm fine with `hyperledger/fabric-config` for now then

denyeart:

text/0000-config-transaction-library.md
```diff
@@ -0,0 +1,85 @@
...
+
+API documentation already exists in the godocs for this package and can be referenced [here](https://godoc.org/github.com/hyperledger/fabric/pkg/configtx).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Many potential consumers may already have a stable method for updating configuration transactions and may be
+unwilling to make amendments to their processes by consuming this library. It could be argued that even if this library
+exists, no external clients already using `configtxgen` would want to switch to consuming it. Another consideration is this library
+will only be provided in Golang, but may eventually support TinyGo compilation to WebAssembly for usage as a web module.
```
davidkhala, line 50: Some years ago, I have similar ideas discussed with nodejs-sdk maintainers that we could make nodejs version of configtxgen or even configtxlator, It is always possible. I am quite happy that we could have this as an isolated go package. And if we plan to provide a Config Transaction Library on nodejs, I am one of volunteer.

PR #27, opened 2020-04-22
RFC for ledger checkpointing

manish-sethi:
Signed-off-by: manish <manish.sethi@gmail.com>

### DIFF
--- a/text/0000-ledger-checkpointing.md
+++ b/text/0000-ledger-checkpointing.md
@@ -1,0 +1,380 @@
+---
+layout: default
+title: Ledger Checkpointing
+nav_order: 3
+---
+
+* Feature Name: ledger_checkpointing
+* Start Date: 2020-04-22
+* RFC PR: (leave this empty)
+* Fabric Component: core, ledger
+* Fabric Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


denyeart:
I had reviewed a draft of the RFC, so I am good with current version.

roger505 (commented):
Is it possible makes sure that it is possible to create the check the checkpoint at the  current ledger height, and not wait for a triggering block. Our client does not work 24*7 so would want to take the checkpoint when no new blocks are being cut.

manish-sethi:
> Is it possible makes sure that it is possible to create the check the checkpoint at the current ledger height, and not wait for a triggering block. Our client does not work 24*7 so would want to take the checkpoint when no new blocks are being cut.

Yes, sure. We can take the height `0` in the checkpoint request as a special case, which would be interpreted as the current height. However, the term current height will be a bit loosely defined here.

denyeart (approved):

denyeart:
Not many comments on this RFC, probably because the draft was well reviewed even before the RFC got opened.
Let's enter Final Comment Period to give people one final week to review.

mastersingh24 (approved):

PR #28, opened 2020-06-01
Wasm Smart Contract RFC

mbwhite:
Signed-off-by: Matthew B White <whitemat@uk.ibm.com>

[Overview presentation](https://docs.google.com/presentation/d/1yEcdEC_D9S-Ax_Zc05K1-Rvid4MIN3zOsWxiIxorvpY/edit?usp=sharing) 

Discussion on Community Call 10June

### DIFF
--- a/text/0006-wasm-smart-contracts.md
+++ b/text/0006-wasm-smart-contracts.md
@@ -1,0 +1,440 @@
+---
+layout: default
+title: Wasm_Smart_Contracts
+parent: RFCs
+---
+
+- Feature Name: Support writing Smart Contracts to be executed within a WebAssembly runtime
+- Start Date: 2020-05-20
+- RFC PR: (leave this empty)
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC is providing:
+
+* a WebAssembly runtime that can be used as a Chaincode implementation in Fabric
+* implementation of a set of APIs to permit Smart Contracts to be written in Rust and then compiled to Wasm bytecode
+* documented interface to permit the community to implement Smart Contracts in other languages. 
+
+A Rust developer would write their Smart Contract with Rust and produce a library. This they would then target to be compiled to Wasm bytecode - the resulting wasm binary would then be deployed using Fabric v2 "External Builders" into a Wasm runtime. This contract is then treated just the same as any other contract or chaincode. 
+
+The APIs will be implemented at a level consistent with the existing Contract APIs but taking the opportunity to make improvements. 
+
+Note: code examples are provided for illustration, and should be considered pseudo-code to outline the functionality that will be produced.
+
+Prototypes are available for the [Rust API](https://github.com/hyperledgendary/fabric-contract-api-rust), [Wasm Chaincode](https://github.com/hyperledgendary/fabric-chaincode-wasm); work is still taking place in these repos so please get contact with questions.
+
+# Motivation
+[motivation]: #motivation
+
+Wasm provides a "well defined, tightly specified, isolated" runtime. It is ideally suited to executing code in a well-controlled and safe manner. Originally intended for use within the browser it is perfectly suited to running outside the browser. The requirements of chaincode and smart contracts are ideally suited to the Wasm model. 
+
+The Contract Programming Model initially introduced in v1.4 provided higher-level abstractions to both the contract and the client SDKs. Many users have already written high-level utility or support code within their applications. Abstractions help us think about and solve problems, improving usability, speed of development and performance. The API in this RFC will be the evolution of the existing work. 
+
+Language support for compiling to Wasm is growing; Rust is one of the languages today that can be efficiently compiled to Wasm. C/C++ and Assembly Script are also candidates. Other languages will arrive over time, for example, Go. One of the deliverables of the Wasm runtime will be the protocol and API to allow the community to develop bindings for new languages. 
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A complete solution using Fabric can be represented as follows:
+![](0000-Complete-ConceptualArchitecture.png)
+
+An application would typically connect to a peer (the 'gateway' into the network channel), and obtain a reference to a smart contract. The Network-shared transaction logic in the Smart Contract is then be invoked by name with optional data arguments and return values. The data arguments can include transient data if needed. The logic in the Smart Contract will typically reference and update the shared transaction data, subject of course to exactly how the code is written.
+
+There are no application API changes as part of this RFC, the existing SDKs should be used.
+
+To write a Smart Contract in Rust for deployment to the Wasm chaincode runtime, a developer will follow these steps
+
+- create a Rust library crate
+    - add the crate dependencies for the Contract-API and macros
+    - and anything else they wish.  The developer is responsible for determining that the other libraries can be compiled correctly to Wasm
+- implement the Contract trait for any struct they wish to be a Contract
+    - using the documented macros for transaction functions
+    - use the Ledger API to interact with data
+- implement the DataType trait for any struct they wish to be passed to and from Transaction Functions
+    - using the documented macros again here
+- compile the crate targeted at Wasm
+- configure the supplied External Builder to run the Wasm binary
+    - using the external chaincode builder scripts supplied to run the Wasm chaincode
+- interact with the deployed chaincode as per other chaincode implementations
+
+## Development APIs
+ For writing contracts, the API logically splits into three parts
+
+- Definitions to define what a contract is, and mark transaction functions
+- Definitions to define how the implementation of the contract will 
+    - work with the data in the ledgers, and other collections.  
+    - work with the concepts of Transactions, Invoking other Contracts. 
+- Definitions to let the contract implementation discover identity information about the organization submitting the current transaction.
+
+The first is generally called the Contract API, the second Ledger API. 
+
+### Contract
+
+The model used here is consistent with the existing Contract Model, with some minor improvements (that can be cascaded back to Java/Nodejs/Golang). 
+
+- A deployed 'chaincode' will be implemented by one or more 'contracts'. 
+- Each 'contract' will consist of one or more 'transaction functions'. 
+- Each 'transaction function' can take arguments at the discretion of the developer. 
+- Within each 'transaction function' the developer will then use the Ledger API to interact with the ledger data.
+- Transaction Functions are addressed by name defined as the name of the contract, followed by the name of the transaction function
+    - A default contract is defined implying that the transaction functions within the default contract can be addressed by transaction function name alone.
+    - If only one contract is defined, this will implicitly be assumed to be the default.
+
+For example:
+
+```rust
+// macros for marking up the contract
+use contract_macros::contract_impl;
+use contract_macros::transaction;
+
+/// Structure for the AssetContract, on which implemenation transaction functions will be added
+pub struct AssetContract {
+
+}
+/// Implementation of the contract trait for the AssetContract
+/// There are default implementation methods, but can be modified if you wish
+/// 
+/// Recommended that the name() function is always modified
+impl Contract for AssetContract {
+    //! Name of the contract
+    fn name(&self) -> String {
+        format!("AssetContract")
+    }
+}
+
+#[contract_impl]
+impl AssetContract {
+    #[transaction]
+    pub fn asset_exists(assset_id:  String) -> Result<bool,ContractError> {
+        let ledger = Ledger::access_ledger();
+        let world = ledger.get_collection(CollectionName::World);
+
+        Ok(world.state_exists(assset_id))
+    }
+}
+```
+
+The developer could write several such contracts. As rust is a static language without runtime introspection, the developer will need to explicitly identify the contracts to the runtime.
+
+For example:
+
+```rust
+launch_handler!(register);
+
+/// Register the contracts
+/// Look to fold this into the macro (or the contract_impl macro)
+pub fn register() -> i32 {
+    ContractManager::register_contract(Box::new(AssetContract::new()));
+
+    // return 0 to indicate success, other values imply failure and will terminate
+    // the chaincode container
+    0
+}
+```
+
+### Ledger
+
+A Ledger's current value is held in a set of collections. A single ledger is available to any given deployed contract, though other ledgers may be available within the Network Channel and can be accessed via other contracts
+
+Of the collections that are reference by the ledger, there a single collection in the ledger that is public (the 'WorldState') and held by every organization in the Network Channel. A ledger may have 1 or more private collections that are available only in some organizations. Starting at Fabric v2.0 there will be an implicit private collection per organization.
+
+## Collection
+
+A collection is a set of states, with each state holding a business object or data. Each state being addressed by a key. Private collections are identified by name and held within a set of policy-defined organizations.
+
+Accessing any given collection is via the Ledger instance. 
+
+```rust
+    let Ledger = Ledger::access_ledger()
+
+    // predefined constant for the world state
+    let world = ledger.get_collection(CollectionName::World);
+
+    // Private Data collections are reference by name
+    let private_collection = ledger.get_collection(CollectionName::Private(String::from("PrivateSalesData")));
+
+    // Each organization will have an implicit private data collection created for them
+    let orgs_collection = ledger.get_collection(CollectionName::Organization(String::from("org1")));
+
+```
+
+Should any operation error or not be applicable in some circumstances, a 'not-supported' error will be returned.
+
+## State
+
+A state holds the value of a business object or data, addressed by a key. The format of this object or data is currently defined by the overall solution, and encoding the format of 'the value' of the state is the responsibility of the contract developer.
+
+The state can also anchor the following functions
+ - Access to individual states can be controlled by state-level endorsement policies, APIs are provided to assist in the construction and parsing of these policies. 
+ - Access to the Private Data Hash - essential for private data scenarios.
+ - Access to history details of the state, what transactions affect this state in the past
+ - Creation of composite keys and general key handling helpers
+
+
+The approach being adopted is that data will be put into 'states' that are then given to 'collections. Using the `from` and `into` pattern in rust - more information is in the DataTypes section below. 
+
+## Example Interfaces
+
+### Putting data
+
+The 'issue' method will create a new business object based on supplied data, and store this within the Ledger.
+
+```rust
+
+    #[Transaction(submit = true)]
+    fn create_asset(assetIdString issuer, String assetId, int faceValue) -> Result<MyAsset,ContractError>  {
+
+        // get the collection that is backed by the world state
+        let ledger = Ledger::access_ledger();
+        let world = ledger.get_collection(CollectionName::World);
+
+        let new_asset = MyAsset::new(my_assset_id,value);
+        world.create(State.from(new_asset));
+
+        Ok(())
+     }
+```
+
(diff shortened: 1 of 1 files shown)


text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+    let sbe2 = Endorsement::build("AND( OR ('Org1.member','Org2.member'), 'Org3.member' )");
+    assetState.set_endorsement(sbe2);
+        
+```
+
+### Query
+
+There are two broad approaches to querying data. Either by passing a query to be parsed by CouchDB (also called Rich Query), or by using the State keys. Composite keys form a hierarchy that can be used for querying a range of states.
+
+To unify the different approaches, the API has a query method taking a configuration of the query to be performed, that returns an iterable object of results.  This iterable will contain the results, and be iterated with the appropriate language idioms. Control of pagination and the request for more data is controlled within this iterable.
```
yacovm, line 258: > To unify the different approaches, the API has a query method taking a configuration of the query to be performed, that returns an iterable object of results.  This iterable will contain the results, and be iterated with the appropriate language idioms.

I don't think you can really unify them, as the developer should be aware of what which of the two query types is going to be run. 
The reason is that rich queries do not get re-executed during commit, and therefore you cannot combine writes with rich queries.

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+- RFC PR: (leave this empty)
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC is providing:
+
+* a WebAssembly runtime that can be used as a Chaincode implementation in Fabric
+* implementation of a set of APIs to permit Smart Contracts to be written in Rust and then compiled to Wasm bytecode
```
yacovm, line 18: Why specifically Rust? Do we know of significant adoption of Rust within the ecosystem target audience of Hyperledger Fabric developers?

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+The APIs will be implemented at a level consistent with the existing Contract APIs but taking the opportunity to make improvements. 
+
+Note: code examples are provided for illustration, and should be considered pseudo-code to outline the functionality that will be produced.
+
+Prototypes are available for the [Rust API](https://github.com/hyperledgendary/fabric-contract-api-rust), [Wasm Chaincode](https://github.com/hyperledgendary/fabric-chaincode-wasm); work is still taking place in these repos so please get contact with questions.
+
+# Motivation
+[motivation]: #motivation
+
+Wasm provides a "well defined, tightly specified, isolated" runtime. It is ideally suited to executing code in a well-controlled and safe manner. Originally intended for use within the browser it is perfectly suited to running outside the browser. The requirements of chaincode and smart contracts are ideally suited to the Wasm model.
```
yacovm, line 32: > Wasm provides a "well defined, tightly specified, isolated" runtime. It is ideally suited to executing code in a well-controlled and safe manner. 

How do you see the typical use of containerization of the chaincode runtime we have today? Is is contradictory or complementary to WASM's isolation property?

text/0006-wasm-smart-contracts.md
lindluni: I think it has more to do with the wasm ecosystem. Of the high-level, first-class languages, Rust has the most well developed support for wasm. C and C++ have pretty equal support, but Rust has much better javascript interfacing and thus makes it easier to build wasm in Rust. Its a much more manual process when compiling C and C++ to Rust. Outside of C, C++ and Rust there aren't many languages that have full support for wasm yet, making Rust the natural choice at this point.

text/0006-wasm-smart-contracts.md
mbwhite: The developer is aware of the type of query they are performing: precise syntax might vary but...
```
collection.get_states(KeyQueryHandler::Range("Car001","Car002"));
collection.get_states(RichQueryHandler::Range("select * where car_age>2 and car_price>20000"));
```

text/0006-wasm-smart-contracts.md
mbwhite: Complementary; today the sets of languages suitable for compilation to Wasm have minimal overlap.

Many Wasm runtimes are planning features such as limited execution limits, so could be halted.  

Early days as well for Wasm,.

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+### Wasm Engine
+Several different implementations are available; initial prototypes have used the LifeVM, but for final implementation, Wasmer is being considered. The WebAssembly landscape is still developing so there is some risk with changing standards and implementations, however, some of the new standards could be of benefit in the future.
+
+### grpc Peer Communications
+This will use the go chaincode shim today to provide connectivity to the peer.
+
+### Data Marshalling
+Currently, the API to import/export functions and data between the host and the executing code in the wasm runtime is limited to primitive data types only. A data and function marshalling interface is required, waPC will be used to provide this. 
+
+The logical functions that will be marshalled over waPC from host to client will form part of the deliverables. This is to allow for new languages to be supported and also to permit changes to the peer communications protocol. (these logical functions are referred to as the Contract Core SPI). It will be designed to use the protobuf specification to permit accurate description.
```
pkedy, line 404: Hi! 👋 I am one of the creators of waPC.  I think it is so awesome that you are considering it for smart contracts.  As you've experienced, waPC only handles sending payloads as byte arrays and leaves it to the implementation to handle serialization/deserialization.

I wanted to briefly share our findings using Protobuf.  In Rust, you will find Protobuf works great (everything for the most part in Rust compiles down to Wasm as expected).  In languages like TinyGo (The best Wasm compiler for Go outside the browser right now), reflection and some basic standard libraries that Protos need are not implemented.  Additionally, we really like [AssemblyScript](https://docs.assemblyscript.org/) but likewise there is no Protobuf support.

Because of this, we decided to use MessagePack as the serialization format over Protobuf.  We even started experimenting with a "fit-for-purpose" waPC IDL (called WIDL) and [code generator](https://github.com/wapc/as-codegen) that creates all the necessary glue code (waPC + MsgPack).  So far, developers seem to like the simplicity.  Here is an [example](https://github.com/wapc/as-actor-test).  We'll be releasing an improved code generator that supports extensibility soon.

We hoped that these tools would be needed only for a short time until [interface types](https://github.com/WebAssembly/interface-types) arrived.  Unfortunately, (and I could be wrong) it does not seem like interface types solve for exactly what waPC is trying to accomplish.  It is still early days and I'm hopeful it will take shape.  Regardless, feel free to reach out if you need anything regarding waPC.  Cheers!

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+- Along with required code to bootstrap the whole Wasm engine and load in the binary
+
+### Wasm Engine
+Several different implementations are available; initial prototypes have used the LifeVM, but for final implementation, Wasmer is being considered. The WebAssembly landscape is still developing so there is some risk with changing standards and implementations, however, some of the new standards could be of benefit in the future.
+
+### grpc Peer Communications
+This will use the go chaincode shim today to provide connectivity to the peer.
+
+### Data Marshalling
+Currently, the API to import/export functions and data between the host and the executing code in the wasm runtime is limited to primitive data types only. A data and function marshalling interface is required, waPC will be used to provide this.
```
MHBauer, line 402: need to define wapc, https://github.com/wapc

very interesting to see this solution to the procedure call issue. Amazing how fast wasm is advancing. A bit like watching language advancement in fast-forward.

text/0006-wasm-smart-contracts.md
mbwhite: Thanks for your comment; and the information about MessagePack vs Protobuf. We already have protobuf in use in the peer-chaincode coms protocols so didn't ideally want to add another piece of technology. 

When I looked at Interface Types initially I likewise wasn't quite sure; waPC makes a pretty good fit for what I wanted.  And we don't need to write our own go host now... though do I have a node.js version for testing.

mbwhite:
For reference, a Hyperledger Lab https://github.com/hyperledger-labs/fabric-chaincode-wasm provided very useful background on running Wasm in chaincode.

text/0006-wasm-smart-contracts.md
pkedy: Yeah, great point @mbwhite.  After posting this I got curious about creating a Protobuf library for AssemblyScript and found that at least the library and hand-coding marshaling/unmarshaling would be relatively easy.  It might even take less coding time than our [MsgPack](https://github.com/wapc/as-msgpack) library.  The harder part might be creating a `protoc` plugin.  I'll try to find time to investigate that but it all feels very doable.  My feeling is that AssemblyScript, being so similar to TypeScript/JavaScript, might be more the "target audience" vs. Rust developers but obviously supporting both is ideal.

mbwhite: For some later work, I've prototyped a protoc plugin;  the basic structure looks confusing at first but I was lucky to find a good example;  the actual code generation part.. well that's a different matter!

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+
+A Rust developer would write their Smart Contract with Rust and produce a library. This they would then target to be compiled to Wasm bytecode - the resulting wasm binary would then be deployed using Fabric v2 "External Builders" into a Wasm runtime. This contract is then treated just the same as any other contract or chaincode. 
+
+The APIs will be implemented at a level consistent with the existing Contract APIs but taking the opportunity to make improvements. 
+
+Note: code examples are provided for illustration, and should be considered pseudo-code to outline the functionality that will be produced.
+
+Prototypes are available for the [Rust API](https://github.com/hyperledgendary/fabric-contract-api-rust), [Wasm Chaincode](https://github.com/hyperledgendary/fabric-chaincode-wasm); work is still taking place in these repos so please get contact with questions.
+
+# Motivation
```
cendhu, line 29: I would like to understand the problem being solved by this RFC. Currently, we can write smart-contract using go, java, nodejs, evm, and contract-api. Why do we need yet another way to write a smart-contract? Sometimes having too many choices is bad too. Especially when we need to add a new chaincode API, we need to ensure that all repos are updated, works correctly, and being maintained too. 

For example, if it helps in (a) improving the performance compared to go or other supported languages or (b) enabling deterministic execution or (c) enabling new use-cases or (d) avoiding bugs, this RFC can be understood better.

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+    let Ledger = Ledger::access_ledger()
+
+    // predefined constant for the world state
+    let world = ledger.get_collection(CollectionName::World);
+
+    // Private Data collections are reference by name
+    let private_collection = ledger.get_collection(CollectionName::Private(String::from("PrivateSalesData")));
+
+    // Each organization will have an implicit private data collection created for them
+    let orgs_collection = ledger.get_collection(CollectionName::Organization(String::from("org1")));
```
cendhu, line 161: These two comments are applicable here too but I am fine with it if maintainers do not have any objections.  
https://github.com/hyperledger/fabric-rfcs/pull/16#discussion_r376212045 
https://github.com/hyperledger/fabric-rfcs/pull/16#discussion_r375970927

text/0006-wasm-smart-contracts.md
```diff
@@ -0,0 +1,440 @@
...
+    - external builder scripts and Dockerfile for a Wasm aware peer
+- `fabric-chaincode-wasm`
+    - the chaincode as a server to host wasm contracts
+- `fabric-contract-api-rust`  
+    - contracts 
+- `fabric-ledger-protos`      
+    - source for protos
+- `fabric-ledger-protos-go`   
+    - auto-generated from protos repo
+- `fabric-ledger-protos-rust`
```
cendhu, line 432: We already have too many repos. Not sure whether 6 more repos are needed.

text/0006-wasm-smart-contracts.md
mastersingh24: The primary WebAssembly languages these days are Rust, AssemblyScript and C/C++.
The official Go compiler supports WebAssembly but assumes a browser/JavaScript host env.  TinyGo looks like a good option but is still missing support for a few standard packages we might need.

Using Rust to start with is our best bet as most WASM stuff can be found in Rust

text/0006-wasm-smart-contracts.md
mastersingh24: We actually want to deprecate the current chaincode model over time.  While the current chaincode implementation made sense at the time given we did not want to write our own virtual machine / byte code interpreter as well as the fact that we did not feel EVM byte code would meet all of our needs, if something like WebAssembly had existed at the time, it's the route we would have taken.

Benefits:
- reduce duplication of code - we no longer have to write shims in multiple languages
- identical model can be used for out-of-process and in-process execution (for out-of-process we'd only need one external host process / shim as well)
- reduce overall complexity of the system

Additionally, the current chaincode stub "interface" has become polluted with functions as things have evolved.  To overcome that complexity, the contract API wrapper layer was created which while making things more usable, actually makes the code base more complex.  So we are not simply going to "port" the existing stub / contract API to this implementation but rather create a new set of APIs which better map to the overall Fabric model.

text/0006-wasm-smart-contracts.md
mbwhite: possibly bad phrasing - consider these the main components of the system; they may in their repos or not.

text/0006-wasm-smart-contracts.md
diegobes: If you are talking about simplicity to start writing and playing with Wasm, I would suggest AssemblyScript and go-wasmer ( as vm host ). AssemblyScript is very close to TypeScript ( with some limitation related to Wasm ) that compiles to wasm bytecodes.
The toolset is very easy to install and use, the usual nmp install and npm run

We are actually running Fabric +  Go Execution environment ( using  go-wasmer as wasm host ) + AssemblyScript ( wasm guest ).
We wrote type binding between AssemblyScript and the Go host application.

text/0006-wasm-smart-contracts.md
diegobes: Regarding protobuf for AssemblyScript. Whay not use Prot3 bindings for JSON and manipulate the proto message entirely in json format in AssemblyScript ( we are already using protobuf serialized json objects inside AssemblyScript this way )

diegobes:
Wasm for chaincode can be a great addition alongside Go, Java and Node JS. I understand the motivation of adding wasm, We actually already using fabric+wasm to execute some BL on our application.

What will be really interesting to see, and get feedback from the community, is an embedded wasm VM in Fabric core to securely execute ledger stored wasm code as part of a transaction process.

text/0006-wasm-smart-contracts.md
autodidaddict: Having been down this road (deciding on supporting tech for both host and guest) a couple of times now, Rust certainly makes things easier on both sides. However, I think more important than the compilers is the contract between the guest and the host (see the [waPC](https://github.com/wapc) discussion below).  The ideal route is to design your contract such that the language choice of either the guest or the host can be swapped out without ever breaking the contract.

text/0006-wasm-smart-contracts.md
cendhu: Thanks, Gari. I wasn't aware of the whole plan. Your response answered both of my comments and now I understand the goal of this RFC.

text/0006-wasm-smart-contracts.md
cendhu: I see. Got it. Thanks, Matthew.

text/0006-wasm-smart-contracts.md
mastersingh24: great comments / questions ... needed to be answered for sure

text/0006-wasm-smart-contracts.md
mbwhite: @autodidaddict Absolutely agree

mbwhite:
This RFC was discussed on the Community Call 10 June (recording will be on https://wiki.hyperledger.org/display/fabric/Contributor+Meetings+2020)

Discussion on the if the go shim was the best library based on the FIPS status of the crypto library. That is outside the scope of this RFC; however should it ever be required this work could be rebased on another shim. 

Discussion on performance; as soon as practical we'll run Caliper over this to get an idea of performance.

denyeart:
@mbwhite 
It would be nice if the conclusions of the above discussions get reflected in the main RFC text, especially Gari's comments around motivation.

Also, I think we need to introduce the question of whether the WASM runtime would ultimately be external to the peer as with traditional chaincode, or embedded in the peer as a new pattern. We don't necessarily need an answer at this point, but the options should be discussed in the RFC text along with pros and cons of each approach. Having an open question of this magnitude may cause the RFC to remain in open state for an extended time period, but I don't think that is necessarily a problem. RFCs can stay open and evolve and mature over an extended period of time. For example we may want to prototype both approaches, and post findings in the RFC, rather than rush the external chaincode approach to market as a production solution that would have to be supported and maintained for years.

jt-nti:
ICYMI, this RFC now has a tech preview!

[Rust Smart Contracts and Wasm chaincode - Tech Preview](https://lists.hyperledger.org/g/fabric/message/9031)

Please try it out and leave any feedback you have here.

sykesm (commented):
After having a few discussions about this last year, I'm going to provide my feedback directly on this RFC.

While this RFC is advertised as "Smart Contracts within a WebAssembly runtime", the focus is on the design and implementation of a contract API in Rust. I see little in this document that describes the details of what functions a chaincode module would need to export, what functions the hosting runtime exposes to a chaincode module, or any information about the module lifecycle that would help someone determine if, for example, a module is reused to handle multiple requests.

The reference to waPC simply indicates that we expect modules to use low level interactions with the host. This is largely the equivalent of saying that we support `read` to get data from the host and `write` to send data to the host. While these are necessary, they're not sufficient without additional context. I'd expect an RFC for WebAssembly to provide this context.

Given the focus on Rust, I look at this RFC as a request to deliver a Contract API in Rust and I believe it should be rewritten to be evaluated in that context. If we do that, we'll then see the WASM related pieces as gratuitous, additional complexity. Basically, if we want to deliver a Contract API in Rust, we should do that independent of any discussion of Web Assembly.

In terms of Web Assembly, I'd very much like to discuss how we can use it within the peer as a sandboxed extension mechanism. While we can use it to run chaincode in-process, I'd also like to see some thoughts around how we can use it for custom endorsement, validation, and access control scenarios.

likeabbas:
Hi, was there a gavel on the future of Fabric using a wasm runtime?

mbwhite:
@likeabbas there was an idea that it might become embedded within the main Fabric runtime.  Chaincode though is now very much viewed as something that is managed externally from Peer... so the peer doesn't know what the runtime of the chaincode is (see the chaincode-as-a-service and the k8s builder).

Writing a chaincode in WASM and running is still perfectly possible/plausible and would be good to do;  and I'd be happy to advise/assist if possible. 

But from the core-fabric maintainers there's only so many hours in a day :-)

likeabbas:
Completely understandable! I will start messing around with https://github.com/hyperledgendary/fabric-contract-api-rust to see if I can get it working with the latest version of fabric. I appreciate the response I will reach out for advice at the appropriate stages :)

mbwhite:
@likeabbas please - it was working when we last touched it :-)

Feel free to reach out via the Hyperledger Discord server... (id=calanais)

PR #32, opened 2020-09-15
Fabric Gateway RFC

andrew-coleman:
RFC for new Fabric Gateway component.

Signed-off-by: andrew-coleman <andrew_coleman@uk.ibm.com>

### DIFF
--- a/text/0000-fabric-gateway.md
+++ b/text/0000-fabric-gateway.md
@@ -25,12 +25,13 @@ By providing a single entry point to a Fabric network, client applications can i
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-The Fabric Gateway is an embodiment of the high-level 'gateway' Fabric programming model in a server component that will form part of a Fabric network alongside Peers, Orderers and CAs.  It will either be stood up inside its own process in a docker container, or it will be hosted inside a peer.  Either way, it exposes its functionality to clients via a gRPC interface.
+The Fabric Gateway is an embodiment of the high-level 'gateway' Fabric programming model in a server component that will form part of a Fabric network alongside Peers, Orderers and CAs.  It will either be stood up inside its own process (optionally in a docker container), or it will be hosted inside a peer.  Either way, it exposes its functionality to clients via a gRPC interface.  The Gateway server component will be implemented in Go.
 
-Lightweight client SDKs are used to connect to the Gateway for the purpose of invoking transactions.  The gateway will intereact with the rest of the network on behalf of the client eliminating the need for the client application to connect directly to peers and orderers.
+Lightweight client SDKs are used to connect to the Gateway for the purpose of invoking transactions.  The gateway will interact with the rest of the network on behalf of the client eliminating the need for the client application to connect directly to peers and orderers.
 
 The concepts and APIs will be familiar to Fabric programmers who are already using the new programming model.
 
+The scope of functionality exposed to the client by this component will include transaction evaluate/submit and event listening.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -47,9 +48,13 @@ service Gateway {
 }
 ```
 
+This exact detail of this protobuf definition may evolve slightly depending on implementation experience, but this shows the general design.
+
 The client credentials (private key) are never passed to the Gateway.  Client applications are responsible for managing their user keys (optionally using SDK Wallets) and signing the protobuf payloads.
 
-Submitting a transaction is a two step process (performed by the client SDK):
+The SDKs will use these services to implement the `SubmitTransaction` and `EvaluateTransaction` functions as follows:
+
+__Submitting a transaction__ is a two step process (performed by the client SDK):
 - __Endorse__
   - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTransaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
   - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
@@ -60,26 +65,26 @@ Submitting a transaction is a two step process (performed by the client SDK):
     - The Gateway will register transaction event listeners for the given channel/txId.
     - It will then broadcast the `Envelope` to the ordering service.
     - The success/error response is passed back to the client in the stream
-    - The Gateway awaits suffient tx commit events before returning and closing the stream, indicating to the client that transaction has been committed.
+    - The Gateway awaits sufficient tx commit events before returning and closing the stream, indicating to the client that transaction has been committed.
 
-Evaluating a transaction is a simple process of invoking the `Evaluate` service passing a SignedProposal message.  The Gateway passes the request to a peer of it's choosing according to a defined policy (probably same org, highest block count) and returns the chaincode function return value to the client.
+__Evaluating a transaction__ is a simple process of invoking the `Evaluate` service passing a SignedProposal message.  The Gateway passes the request to a peer of it's choosing according to a defined policy (probably same org, highest block count) and returns the chaincode function return value to the client.
 
 ## Launching the Gateway
 
 #### Standalone process
-When running standalone, the Gateway is effectively a 'client' application to a set of peers.  It will be configured to connect to a peer in the organization in order to invoke discovery.  To do this it will need a signing identity in the channel writers policy.
+When running standalone, the Gateway is effectively a 'client' application to a set of peers.  A gateway instance is associated with an organization and will be configured to connect to a peer in the same organization in order to invoke discovery.  To do this it will need a signing identity in the channel writers policy.
 
-To run the gateway server, the following parameters need to be supplied:
+To run the gateway server, the following parameters must to be supplied:
 - The url of at least one peer in the org.  Once connected, the discovery service will be invoked to find other peers.
 - The MSPID associated with the gateway.
-- The signing identity of the gateway (cert and key), e.g. location of PEM files on disk.
-- The TLS certificate so that the gateway can connect to other components in the organization
-The following is an example command line invocation using a prototype:
-- `gateway -h peer0.org1.example.com -p 7051 -m Org1MSP -id ../../fabric-samples/fabcar/javascript/wallet/gateway.id -tlscert ../../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/tlsca/tlsca.org1.example.com-cert.pem`
+- The signing identity of the gateway (cert and key), e.g. location of PEM files on disk.  
+- The CA certificate so that the gateway can make TLS connections to other components (peers/orderers) in the organization
+
+Note that the signing identity is required so the gateway server can make discovery requests and to register event listeners.  It is not used for submitting any transactions on behalf of end users.
 
 #### Embedded in Peer
 When embedded in a peer, the gateway will register its gRPC service with the host peer's gRPC server.
-The gateway does not need its own (client) identity since it can get all the discovery information directly from the host peer rather than via a gRPC signed request.
+The gateway does not need its own (client) identity since it can get all the discovery and event information directly from the host peer rather than via a gRPC signed request.
 
 ### Discovery
 
@@ -124,7 +129,23 @@ Implementation options:
 
 Currently a client application is responsible for load balancing its requests between multiple peers.  The Gateway will handle this on behalf of the set of currently connected client applications.
 
-Ideally, the Gateway itself will be stateless (i.e. not maintain client session state) allowing clients to be routed though an appriopriate load balancer.
+Ideally, the Gateway itself will be stateless (i.e. not maintain client session state) allowing clients to be routed though an appropriate load balancer.
+
+## Trust assumptions
+
+The Gateway is acting as a proxy between the client application and the peers/orderers.  The signed transaction proposal is created in the client using one of the SDKs, and gateway passes that directly to the endorsing peers without any modification.  If the client wishes, it can inspect the contents of the 'PreparedTransaction' message returned by the Endorse function prior to signing and submitting.   The Gateway will partially unpack the proposal responses in order to add the chaincode function return value and the transaction id to the PreparedTransaction message.  This allows the clients to access these commonly used values without the overhead of unpacking the deeply nested transaction 'Envelope' payload.
+
+Trusting clients can call the SDK's submitTransaction function to endorse/sign/submit in a single line of code for convenience. Users of the existing high level programming model are already comfortable with this procedure. 
+
+## Authentication and Authorization
+
+The Gateway itself will not add any message level authentication mechanism other than that already provided by the peers.  As a proxy, authentication and authorization is handled by peer/orderer nodes exactly as if the client was connecting to them directly.
+
+Connection of the client applications to the Gateway can be secured by mutual TLS if required.
+
+The standalone Gateway does need to have an identity that allows it to connect to peers as a client in its own right to:
+- Do network discovery to obtain network topology and endorsement plans.
+- Listen to events to allow it to check for commit disposition of submitted transactions that it has routed to the ordering service on behalf of a client.
 
 ## Gateway SDKs
 
@@ -135,7 +156,6 @@ The following classes/structures will be available as part of the gateway progra
 - Network
 - Contract
 - Transaction
-- Wallet
 
 The existing `submitTransaction()` and `evaluateTransaction()` API methods will be implemented using the Gateway gRPC services described above.
 
@@ -148,7 +168,7 @@ SDKs will be created for the following languages:
 - Go
 - Java
 
-It is proposed that these SDKs will be maintained in the same GitHub respository as the gateway itself (`hyperledger/fabric-gateway`) to ensure that they all stay up to date with each other and with the core gateway.  This will be enforced in the CI pipeline by running the end-to-end scenario tests across all SDKs.
+It is proposed that these SDKs will be maintained in the same GitHub repository as the gateway itself (`hyperledger/fabric-gateway`) to ensure that they all stay up to date with each other and with the core gateway.  This will be enforced in the CI pipeline by running the end-to-end scenario tests across all SDKs.
 
 Publication of SDK releases will be as follows:
 - Node SDK published to NPM.
@@ -161,7 +181,7 @@ Contributors are invited to create SDKs for other languages, e.g. Python, Rust.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-When run as a standalone process, the gateway would add another docker containter to an (arguably) already complex Fabric architecture.  Also it will require its own signing identity in the channel writers policy in order to invoke the discovery service, which might be a security concern in some environments. These concerns are mitigated by embedding the gateway server within the organisation's existing peers.  
+When run standalone, the gateway would add another process to an (arguably) already complex Fabric architecture.  Also it will require its own signing identity in the channel writers policy in order to invoke the discovery service, which might be a security concern in some environments. These concerns are mitigated by embedding the gateway server within the organisation's existing peers.  
 
 However, if embedded in an existing peer, then the gateway will not be able to be scaled independently of the peers.  Also, there are scenarios where organisations don't run their own peer(s).  In both of these cases, running a standalone gateway will be advantageous.
 
@@ -195,7 +215,7 @@ The CI pipeline will run the unit tests for the Gateway and its SDKs as well as 
 # Dependencies
 [dependencies]: #dependencies
 
-- This will depend on the peer implementation in `hyperledger/fabric-protos` and the protobuf definitions in `hyperledger/fabric-protos` repo.
+- This will depend on the peer implementation in `hyperledger/fabric` and the protobuf definitions in `hyperledger/fabric-protos` repo.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/text/0000-fabric-gateway.md
+++ b/text/0000-fabric-gateway.md
@@ -1,0 +1,213 @@
+---
+layout: default
+title: Fabric Gateway
+nav_order: 3
+---
+
+- Feature Name: Fabric Gateway
+- Start Date: 2020-09-08
+- RFC PR: (leave this empty)
+- Fabric Component: fabric-gateway
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The Fabric Gateway is a new component that will run either as a standalone process, or embedded with the peer, and will implement as much of the high-level 'gateway' programming model as possible.  This will remove much of the transaction submission logic from the client application which will simplify the maintenance of the SDKs.  It will also simplify the administrative overhead of running a Fabric network because client applications will be able to connect and submit transactions via a single network port rather than the current situation where ports have to be opened to multiple peers across potentially multiple organisations.
+
+# Motivation
+[motivation]: #motivation
+
+The high-level "gateway" programming model has proved to be very popular with users since its introduction in v1.4 in Node SDK.  Since then, it has been implemented in the Java and Go SDKs.  This has led to a large amount of code that has to be maintained in the SDKs which increases the cost of implementing new feature which have to be replicated to keep the SDKs in step.  It also increases the cost of creating an SDK for a new language.  Moving all of the high-level logic into its own process (or peer) and exposing a simplified gRPC interface will drastically reduce the complexity of the SDKs.
+
+By providing a single entry point to a Fabric network, client applications can interact with complex network topologies without multiple ports having to be opened and secured to each peer and ordering node.  The client will connect only to the gateway which will act on its behalf with the rest of the network.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The Fabric Gateway is an embodiment of the high-level 'gateway' Fabric programming model in a server component that will form part of a Fabric network alongside Peers, Orderers and CAs.  It will either be stood up inside its own process in a docker container, or it will be hosted inside a peer.  Either way, it exposes its functionality to clients via a gRPC interface.
+
+Lightweight client SDKs are used to connect to the Gateway for the purpose of invoking transactions.  The gateway will intereact with the rest of the network on behalf of the client eliminating the need for the client application to connect directly to peers and orderers.
+
+The concepts and APIs will be familiar to Fabric programmers who are already using the new programming model.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The Gateway runs as a client process associated with an organisation and requires an identity (issued by the org's CA) in order to interact with the discovery service and to register for events (Deliver Client).  
+
+The Gateway also runs as a gRPC server and exposes the following services to client applications (SDKs):
+
+```
+service Gateway {
+    rpc Endorse(ProposedTransaction) returns (PreparedTransaction) {}
+    rpc Submit(PreparedTransaction) returns (stream Event) {}
+    rpc Evaluate(ProposedTransaction) returns (Result) {}
+}
+```
+
+The client credentials (private key) are never passed to the Gateway.  Client applications are responsible for managing their user keys (optionally using SDK Wallets) and signing the protobuf payloads.
+
+Submitting a transaction is a two step process (performed by the client SDK):
+- __Endorse__
+  - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTreansaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
+  - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
+    - The Gateway will determine the endorsement plan for the requested chaincode and forward to the appropriate peers for endorsement. It will return to the client a `PreparedTransaction` message which contains a `Envelope` message as defined in `fabric-protos/common/common.proto`.  It will also contain other information, such as the return value of the chaincode function and the transaction ID so that the client doesn't necessarily need to unpack the `Envelope` payload.
+- __Submit__
+  - The client signs the hash of the Envelope payload using the user's private key and sets the signature field in the `Envelope` in the `PreparedTransaction`.
+  - The `Submit` service is invoked on the Gateway, passing the signed `PreparedTransaction` message.  A stream is opened to return multiple return values.
+    - The Gateway will register transaction event listeners for the given channel/txId.
+    - It will then broadcast the `Envelope` to the ordering service.
+    - The success/error response is passed back to the client in the stream
+    - The Gateway awaits suffient tx commit events before returning and closing the stream, indicating to the client that transaction has been committed.
+
+Evaluating a transaction is a simple process of invoking the `Evaluate` service passing a SignedProposal message.  The Gateway passes the request to a peer of it's choosing according to a defined policy (probably same org, highest block count) and returns the chaincode function return value to the client.
+
+## Launching the Gateway
+
+#### Standalone process
+When running standalone, the Gateway is effectively a 'client' application to a set of peers.  It will be configured to connect to a peer in the organization in order to invoke discovery.  To do this it will need a signing identity in the channel writers policy.
+
+To run the gateway server, the following parameters need to be supplied:
+- The url of at least one peer in the org.  Once connected, the discovery service will be invoked to find other peers.
+- The MSPID associated with the gateway.
+- The signing identity of the gateway (cert and key), e.g. location of PEM files on disk.
+- The TLS certificate so that the gateway can connect to other components in the organization
+The following is an example command line invocation using a prototype:
+- `gateway -h peer0.org1.example.com -p 7051 -m Org1MSP -id ../../fabric-samples/fabcar/javascript/wallet/gateway.id -tlscert ../../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/tlsca/tlsca.org1.example.com-cert.pem`
+
+#### Embedded in Peer
+When embedded in a peer, the gateway will register its gRPC service with the host peer's gRPC server.
+The gateway does not need its own (client) identity since it can get all the discovery information directly from the host peer rather than via a gRPC signed request.
+
+### Discovery
+
+On startup, the gateway will connect to its primary peer(s), as specified in the command line option (or its host, if embedded in a peer).  The discovery service will be invoked to build and maintain a cache of the network topology per channel.  This will be used to identify endorsers based on the chaincode endorsement policy.
+
+#### Collection endorsement policy
+
+Discovery can be used to get endorsing peers for a given chaincode where the transaction function will write to a given collection, and this will take into account the collection level endorsement policy.
+Ideally we could take care of these endorsement requirements in the Gateway without the client needing to have any implicit knowledge of them. Suggest extending contract metadata to include information on collections, which can be queried by Fabric Gateway. If that isn't possible then we might need some way for the client to specify the collections that will be used (not desirable).
+
+### State based endorsement
+
+This is driven by business logic and rules. Requires the client to supply constraints on endorsing organisations as part of the proposal endorsement.
+
+### Chaincode-to-chaincode endorsement
+
+Augment contract metadata for transactions with additional chaincodes / transactions that will be invoked, then we use discovery to narrow the set of endorsers.
+
+### Proposal response consistency checking
+
+Current client SDK implementations do minimal checking of proposal responses before sending to the orderer. There have been some requests for more rigorous checking of proposal responses prior to submit to the orderer.
+
+Prior to sending to the orderer, Fabric Gateway should check that proposal responses are:
+- Consistent, including having matching response payloads.
+- Meet all applicable endorsement requirements.
+
+### Ensure query results reflect latest ledger state
+
+Query results should be obtained from peers whose ledger height is at least that of the most recent events sent to clients.
+
+Scenario:
+- Client receipt of chaincode event triggers business process.
+- Business process queries ledger to obtain private data, which cannot be included in event payload since it is private.
+The private data must be consistent with the state of the ledger after commit of the transaction that emitted the chaincode event, so the query must be processed by a peer with ledger height of at least the block number that contained the chaincode event.
+
+Implementation options:
+- Only direct queries (and endorsement requests) to the peer(s) with the highest ledger height.
+- Client proposals include last observed ledger height so Gateway can target any peers with at least that ledger height.
+ A view of ledger height can be tracked by the Gateway from block events it has received.
+
+## Scaling and load balancing
+
+Currently a client application is responsible for load balancing its requests between multiple peers.  The Gateway will handle this on behalf of the set of currently connected client applications.
+
+Ideally, the Gateway itself will be stateless (i.e. not maintain client session state) allowing clients to be routed though an appriopriate load balancer.
+
+## Gateway SDKs
+
+A set of SDKs will be created to allow client applications to interact with the Fabric Gateway.  The APIs exposed by these SDKs will be, where possible, the same as the current high-level 'gateway' SDKs.
+
+The following classes/structures will be available as part of the gateway programming model:
+- Gateway
+- Network
+- Contract
+- Transaction
+- Wallet
+
+The existing `submitTransaction()` and `evaluateTransaction()` API methods will be implemented using the Gateway gRPC services described above.
+
+Based on user feedback of the existing gateway SDKs, the following additions are planned to be supported by this implementation:
+- Support for 'offline signing', allowing clients to write REST servers that don't directly have access to users' private keys.
+- Support for 'asynchronous submit', allowing client applications to continue processing while the transaction commit proceeds in the background. For Go this will be implemented by returning a channel; for Node by returning a Promise; and for Java by returning a Future.
+
+SDKs will be created for the following languages:
+- Node (Typescript/Javascript)
+- Go
+- Java
+
+It is proposed that these SDKs will be maintained in the same GitHub respository as the gateway itself (`hyperledger/fabric-gateway`) to ensure that they all stay up to date with each other and with the core gateway.  This will be enforced in the CI pipeline by running the end-to-end scenario tests across all SDKs.
+
+Publication of SDK releases will be as follows:
+- Node SDK published to NPM.
+- Java SDK published to Maven Central.
+- Go SDK published to separate GitHub repo (e.g. fabric-gateway-go) and tagged.
+
+Contributors are invited to create SDKs for other languages, e.g. Python, Rust.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+When run as a standalone process, the gateway would add another docker containter to an (arguably) already complex Fabric architecture.  Also it will require its own signing identity in the channel writers policy in order to invoke the discovery service, which might be a security concern in some environments. These concerns are mitigated by embedding the gateway server within the organisation's existing peers.  
+
+However, if embedded in an existing peer, then the gateway will not be able to be scaled independently of the peers.  Also, there are scenarios where organisations don't run their own peer(s).  In both of these cases, running a standalone gateway will be advantageous.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not
+  choosing them?
+- What is the impact of not doing this?
+
+# Prior art
+[prior-art]: #prior-art
+
+The Gateway SDK has been available as the high level programming model in the Fabric SDKs since v1.4.  It has been thoroughly tested in many scenarios.  This Fabric Gateway component is an embodiment of that programming model in a server component.
+
+# Testing
+[testing]: #testing
+
+In addition to the usual unit tests for all functions, a comprehensive end to end 
+scenario test suite will be created from the outset.  A set of 'cucumber' BDD style 
+tests will be taken from the existing Java and Node SDKs to test all the major functions
+of the gateway against a multi-org multi-peer network.  These tests will include:
+- evaluation and submission of transactions
+- event handling scenarios
+- transient / private data
+- error handling
+
+The CI pipeline will run the unit tests for the Gateway and its SDKs as well as the end-to-end scenario (cucumber) tests against all SDKs.
+
+# Dependencies
+[dependencies]: #dependencies
+
+- This will depend on the peer implementation in `hyperledger/fabric-protos` and the protobuf definitions in `hyperledger/fabric-protos` repo.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+### Client inspection of proposal responses
+
+Fabric is designed such that the client should have the opportunity to inspect the transaction response payload and then make a decision on whether or not that transaction should be sent to the orderer to be committed.
+
+There is the possibility of a client being able to send proposals that are subsequently not submitted to the orderer and recorded on the ledger, either because:
+- The proposal is not successfully endorsed; or
+- The client actively chooses not to submit a successfully endorsed transaction.
+
+This is viewed by some as a potential security vulnerability as it allows an attacker to probe the system.
+
+Fabric Gateway should provide the capability for some kind of audit of submitted proposals, including their transaction ID so they can be reconciled against committed transactions.
--- a/text/0000-fabric-gateway.md
+++ b/text/0000-fabric-gateway.md
@@ -51,7 +51,7 @@ The client credentials (private key) are never passed to the Gateway.  Client ap
 
 Submitting a transaction is a two step process (performed by the client SDK):
 - __Endorse__
-  - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTreansaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
+  - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTransaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
   - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
     - The Gateway will determine the endorsement plan for the requested chaincode and forward to the appropriate peers for endorsement. It will return to the client a `PreparedTransaction` message which contains a `Envelope` message as defined in `fabric-protos/common/common.proto`.  It will also contain other information, such as the return value of the chaincode function and the transaction ID so that the client doesn't necessarily need to unpack the `Envelope` payload.
 - __Submit__


text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not
+  choosing them?
+- What is the impact of not doing this?
+
+# Prior art
+[prior-art]: #prior-art
+
+The Gateway SDK has been available as the high level programming model in the Fabric SDKs since v1.4.  It has been thoroughly tested in manu scenarios.  This Fabric Gateway component is an embodiment of that programming model in a server component.
```
sijocherian: typo: manu > many

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+- The signing identity of the gateway (cert and key), e.g. location of PEM files on disk.
+- The TLS certificate so that the gateway can connect to other components in the organization
+The following is an example command line invocation using a prototype:
+- `gateway -h peer0.org1.example.com -p 7051 -m Org1MSP -id ../../fabric-samples/fabcar/javascript/wallet/gateway.id -tlscert ../../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/tlsca/tlsca.org1.example.com-cert.pem`
+
+Alternatively, the Gateway server will be designed to be embeddable within a Peer, in which case the all of the above parameters will be taken from the peer configuration.
+
+### Discovery
+
+On startup, the gateway will connect to its primary peer(s), as specified in the command line option (or its host, if embedded in a peer).  The discovery service will be invoked to build and maintain a cache of the network topology per channel.  This will be used to identify endorsers based on the chaincode endorsement policy.
```
yacovm, line 91: Right now, the design delegates signing the proposal and the transaction to the client, however here you are saying that the gateway requires an identity and a corresponding private key.

The discovery service authenticates clients if they are in the channel writers policy. This means, that if you can query discovery, you can also submit transactions. 

I am wondering if it would make sense, to delegate the signature on discovery queries to the client.

If we can do that, then you could deploy a gateway without any enrollment certificate at all, and this would potentially secure the whole deal much more, because the said gateway cannot sign transactions on its own (since it needs the client), so, if its file system is somehow compromised, no private key corresponding to an enrollment identity would be stolen.

Another reason why this fits well with discovery, is that discovery queries do not have nonces or timestamps, and are meant to be re-used: If you submit a query Q signed with signature S twice, the peer would only perform verification of signature for the first time, and the second time it would lookup in the cache. 

Therefore I think you should consider further delegating the signatures on discovery queries to the client.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+
+By providing a single entry point to a Fabric network, client applications can interact with complex network topologies without multiple ports having to be opened and secured to each peer and ordering node.  The client will connect only to the gateway which will act on its behalf with the rest of the network.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The Fabric Gateway is an embodiment of the high-level 'gateway' Fabric programming model in a server component that will form part of a Fabric network alongside Peers, Orderers and CAs.  It will either be stood up inside its own process in a docker container, or it will be hosted inside a peer.  Either way, it exposes its functionality to clients via a GRPC interface.
```
yacovm: Given it exposes functionality via gRPC, i guess it'll be implemented once, in Go?

bestbeforetoday: Yes, one implementation in Go. Client only interacts using gRPC so doesn't care about the Fabric Gateway implementation. Similar to the existing nodes.

text/0000-fabric-gateway.md
bestbeforetoday: This is an interesting suggestion. I'm not sure how workable it is. If I describe the issues as I see them then perhaps you can see ways to make this work.

The Gateway is intended stateless and scalable, so a single Gateway instance is used by multiple clients, and a client may interact with different Gateway instances on different requests, which may or may not require discovery. So it seems like we might have to have a chatty interaction where a client generates and signs a discovery query on every request. Or at least checks whether it needs to sign a discovery query, which is equally (if not more) chatty.

Gateway will need to listen to events from peers. I think this will also require signing?

I think the Gateway might at some point need to be able to evaluate transactions, such as the `getMetadata` transactions in chaincode to obtain the chaincode metadata. I am thinking of this as a way of obtaining information needed for endorsement in more complex scenarios, such as identifying the collections used by specific transaction functions, or chaincode-to-chaincode calls made by specific transaction functions.

We've got some mechanisms we could use to improve security of server-side credentials, such as injecting Kubernetes secrets and/or using HSM.

Any signing requirements for the Gateway implementation we expose (or delegate) down to the client introduces tighter coupling between client and the specific point-in-time implementation of the Gateway. Ideally we want to minimise the interface between client and Gateway as much as possible, and to expose as little of the Gateway logic to the client as possible. Requiring client-side signature for things like discovery would trade off against that goal.

The Gateway is running server-side in the same cloud environment as your peers and orders. If your server-side is compromised to that degree then I guess all your nodes and their ledgers are similarly compromised. Given that, how much are we buying by removing the need for the Gateway to have its own signing identity?

yacovm: I'm failing to see the issue. Even if a single gateway is used by multiple clients, you can have an in-memory segment for each client, right? 

As for the chatting - Imagine that you're a client and you're sending the gateway some RPC `Foo` and expect to get some `X` in return. 
It (the gateway) either sends you back that `X`, or, instead, it sends you back (`Y`, `h`) - something to sign (`Y`), and a handle `h` (a unique string) which you can then invoke the same RPC call `Foo` with the signed `Y` and that handle `h` as a parameter to retrieve `X` which should be possible to retrieve now that the gateway has the discovery signature.

> Gateway will need to listen to events from peers. I think this will also require signing?

Yes but when do you listen to events? when you submit a transaction beforehand, no? Since when do clients simply listen to events just for the sake of events? They listen for events to know the status of their transaction.



> I think the Gateway might at some point need to be able to evaluate transactions, such as the getMetadata transactions in chaincode to obtain the chaincode metadata. I am thinking of this as a way of obtaining information needed for endorsement in more complex scenarios, such as identifying the collections used by specific transaction functions, or chaincode-to-chaincode calls made by specific transaction functions.

Are you saying that the gateway is expected to be able to invoke chaincode on its own behalf ? How would it know the parameters of the chaincode? How would the developer of the chaincode know what parameters the gateway expects? It seems to me that you can just instead add an API to the peer to obtain metadata, if that's all you want. 

> Any signing requirements for the Gateway implementation we expose (or delegate) down to the client introduces tighter coupling between client and the specific point-in-time implementation of the Gateway. Ideally we want to minimise the interface between client and Gateway as much as possible, and to expose as little of the Gateway logic to the client as possible. Requiring client-side signature for things like discovery would trade off against that goal.

Understood.


> The Gateway is running server-side in the same cloud environment as your peers and orders. If your server-side is compromised to that degree then I guess all your nodes and their ledgers are similarly compromised. Given that, how much are we buying by removing the need for the Gateway to have its own signing identity?

The key might leak in other ways, like backups, compromising only a single node (host) of the cloud environment but not be able to compromise all of them.

Additionally, there are other vectors to extract a private key than to take over the host. For example, if you give the server to sign a lot of things and the signing is **not** constant time, then by collecting enough samples, you can extract some bits of the private key because you know (statistically) how much time it takes to do certain math operations, so you divide and conquer the key space until you end up recovering the key completely. 

Imagine that I somehow made your gateway to send me lots and lots of discovery requests, because I either have compromised a public TLS CA (it happened in the past) and issued a certificate with the same SAN to myself or from some reason, a TLS private-public key pair was generated with bad randomness and I brute forced the certificate and found the private key. 

Having said that, I guess that any problem that is possible with the gateway, you might already have with the SDK.

text/0000-fabric-gateway.md
yacovm: Why not have a single implementation?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+- Feature Name: Fabric Gateway
+- Start Date: 2020-09-08
+- RFC PR: (leave this empty)
+- Fabric Component: fabric-gateway
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The Fabric Gateway is a new component that will run either as a standalone process, or embedded with the peer, and will implement as much of the high-level 'gateway' programming model as possible.  This will remove much of the transaction submission logic from the client application which will simplify the maintenance of the SDKs.  It will also simplify the administrative overhead of running a Fabric network since the client application will only need to connect to a gateway rather than multiple peers across potentially multiple organisations.
```
yacovm: > It will also simplify the administrative overhead of running a Fabric network since the client application will only need to connect to a gateway rather than multiple peers across potentially multiple organisations.

I'm not sure how this improves administration of the network (unless, the gateway is going to be a service inside the peer):

- You are now adding more nodes to the network, and of a new type, so you need new scripts and configuration.
- The gateways are going to be discovery clients, so hypothetically instead of `n` clients doing discovery queries, you have `k<n` (`k` gateways) doing it, but, this doesn't really improve administration, because if `k` gateways can do it, so do all `n` clients, and if `n` clients cannot do it, due to some environmental issue, then so can't `k` gateways. 
- It only adds another layer which can potentially fail, and now SREs need to know to diagnose the gateways as well.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+
+Fabric is designed such that the client should have the opportunity to inspect the transaction response payload and then make a decision on whether or not that transaction should be sent to the orderer to be committed.
+
+There is the possibility of a client being able to send proposals that are subsequently not submitted to the orderer and recorded on the ledger, either because:
+- The proposal is not successfully endorsed; or
+- The client actively chooses not to submit a successfully endorsed transaction.
+
+This is viewed by some as a potential security vulnerability as it allows an attacker to probe the system.
+
+Fabric Gateway should provide the capability for some kind of audit of submitted proposals, including their transaction ID so they can be reconciled against committed transactions.
```
yacovm, line 233: How does a client authenticate to a gateway?

bestbeforetoday: I think that if you want client authentication, it's likely to be mutual TLS similar to the rest of Fabric. I guess things like OAuth are also a possibility. I don't think this is set in stone.

I think the statement you've commented on refers to the opportunity for the Gateway to log client interactions with the network rather than authentication.

text/0000-fabric-gateway.md
bestbeforetoday: I agree that this introduces additional server-side elements to be managed. Hopefully we can minimise the management effort required.

I don't expect there to be anywhere near the number of Gateways as there are clients. Each Gateway instance will serve many clients. I don't agree with your `k>n` assertion.

Today you have to punch holes in your firewall to allow all clients (including ones outside of your organization) to access all of your nodes. And all of those network clients potentially establish multiple connections to all of those nodes, which presents administrative and scaling challenges. The proposed Gateway model means you only need to allow ingress for your own organization clients to your own Gateway node(s) at a single external endpoint.

text/0000-fabric-gateway.md
bestbeforetoday: > Yes but when do you listen to events? when you submit a transaction beforehand, no? Since when do clients simply listen to events just for the sake of events? They listen for events to know the status of their transaction.

Clients absolutely can (and do) listen to events without submitting transactions, and seeing whether a submitted transaction has committed successfully is absolutely not the only reason clients listen to events.

bestbeforetoday: > Are you saying that the gateway is expected to be able to invoke chaincode on its own behalf ? How would it know the parameters of the chaincode? How would the developer of the chaincode know what parameters the gateway expects? It seems to me that you can just instead add an API to the peer to obtain metadata, if that's all you want.

https://hyperledger.github.io/fabric-chaincode-node/release-2.2/api/tutorial-annotated-contract-metadata.html

I don't think the peer holds the information about a given transaction function implementation that I described.

bestbeforetoday: > As for the chatting - Imagine that you're a client and you're sending the gateway some RPC Foo and expect to get some X in return.
It (the gateway) either sends you back that X, or, instead, it sends you back (Y, h) - something to sign (Y), and a handle h (a unique string) which you can then invoke the same RPC call Foo with the signed Y and that handle h as a parameter to retrieve X which should be possible to retrieve now that the gateway has the discovery signature.

As you've demonstrated, that is more chatty than the proposed interaction. It sounds like you are proposing adding session state into the stateless Gateway.

text/0000-fabric-gateway.md
bestbeforetoday: I just said it was a single implementation. I don't understand the question.

yacovm: misread, I thought you said "one of them is in go"

text/0000-fabric-gateway.md
yacovm: I meant `<`, that `>` was a typo :) 

I understand the firewall issue.

text/0000-fabric-gateway.md
bestbeforetoday: > Additionally, there are other vectors to extract a private key than to take over the host...

How are these vulnerabilities addressed in the peer and orderer today?

yacovm: > > Additionally, there are other vectors to extract a private key than to take over the host...
> 
> How are these vulnerabilities addressed in the peer and orderer today?

We always use curve P-256 which is constant time (its implementation). When I wrote the comment I thought you intended to implement the gateway using several languages.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+- Java SDK published to Maven Central.
+- Go SDK published to separate GitHub repo (e.g. fabric-gateway-go) and tagged.
+
+Contributors are invited to create SDKs for other languages, e.g. Python, Rust.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+When run as a standalone process, the gateway would add another docker containter to an (arguably) already complex Fabric architecture.  This could be mitigated by embedding the gateway server within the organisation's existing peers.  This feature will be designed to be embeddable, if required.
```
cendhu: This message is conveyed at least 3/4 times in the RFC. To me, this looks like a major decision and cannot be left out for later to decide during the implementation. IMO, we need to be explicit on the RFC. As the goal is to keep the peer as small as possible, I would prefer keeping the gateway outside the peer. Adding another process does not impact the usability as the entry point is going to be the gateway for clients. I would prefer micro-service architecture which can help in scalability and maintainability over a monolithic peer process with too many features packed into a single entity. Further, I see gateway as an advanced version of nginx proxy or pgbouncer given it does more than just load balancing. Hence, it is better to keep this add-on/optional feature outside the peer. 

It would be good to discuss the pros/cons of two approaches in the RFC itself and then make a decision.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+Implementation options:
+- Only direct queries (and endorsement requests) to the peer(s) with the highest ledger height.
+- Client proposals include last observed ledger height so Gateway can target any peers with at least that ledger height.
+ A view of ledger height can be tracked by the Gateway from block events it has received.
+
+## Scaling and load balancing
+
+Currently a client application is responsible for load balancing its requests between multiple peers.  The Gateway will handle this on behalf of the set of currently connected client applications.
+
+Ideally, the Gateway itself will be stateless (i.e. not maintain client session state) and so can be scaled independently with clients routing though an appriopriate load balancer.
```
cendhu: As a result, it would be good to keep the gateway as a separate process rather than adding it to the peer process.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,207 @@
...
+
+The Gateway runs as a client process associated with an organisation and requires an identity (issued by the org's CA) in order to interact with the discovery service and to register for events (Deliver Client).  
+
+The Gateway also runs as a GRPC server and exposes the following services to client applications (SDKs):
+
+```
+service Gateway {
+    rpc Prepare(SignedProposal) returns (PreparedTransaction) {}
+    rpc Commit(PreparedTransaction) returns (stream Event) {}
+    rpc Evaluate(SignedProposal) returns (Result) {}
```
cendhu: In database/transactional system, prepare means different things. There is a prepared statement and also there is a preparation step after the tx execution but before the commit (2-phase commit in a standalone DB -- prepare(), commit(). Here, the prepare acquires all necessary lock before the commit). 

To avoid this confusion, can't we reflect the fabric architecture in these APIs? One such example is below
```
Execute(SignedProposal) returns (ExecutionResults) {}
Commit(ExecutionResults) returns (stream Event) {}
```
We can also come up with something better but my point is that the prepare does not convey execution.

text/0000-fabric-gateway.md
yacovm: >  I would prefer micro-service architecture which can help in scalability and maintainability over a monolithic peer process with too many features packed into a single entity

Fabric is not a micro-service friendly system, no matter how hard you want it to be:

- It uses TLS pinning in the orderer
- It is not proxy friendly
- It has its own load balancing and failover logic, and attempts to put cloud-y mechanisms usually end up getting in Fabric's way

I actually am in favor of putting this inside the peer, as you can re-use existing authentication mechanisms which are Fabric native and you don't need to re-implement them, and also the downside that requires completely new scripts and configuration files to launch a brand new node type.

text/0000-fabric-gateway.md
bestbeforetoday: This is a really interesting point. I don't speak for Andy, but I think the intention with this naming is to make it relatable to people coming to Fabric who are more likely to be familiar with concepts like two-phase commit. The prepare gets you ready to be able to successfully commit, which is similar to what you are doing by getting a transaction proposal endorsed.

Looking at it from the other end, in the Fabric programming model exposed by the SDKs we talk about _evaluate_ and _submit_ of transactions. Evaluate is essentially a query where we send a proposal to peer(s) for simulation to get a transaction result; we don't care about endorsement. Submit involves getting peer(s) to endorse the proposal before sending a transaction containing the endorsements to the orderer to be included in a block, and waiting for the block to be committed to peer ledgers. So at the client side I might be thinking of the submit flow consisting of a call to `Endorse` followed by a `Submit` to ordering.

I worry that renaming `Prepare` to `Execute` makes it sound like it is behaving like `Evaluate`; just executing a transaction rather than preparing to commit a transaction by gaining endorsement.

I guess my preference would probably be to either have:
- `Prepare` / `Commit` to imply the sequential relationship between them, as in two-phase commit
- `Endorse` / `Commit` to align with terminology used by Fabric
- `Endorse` / `Submit` to align with the higher level terminology used by the client SDKs

text/0000-fabric-gateway.md
davidkhala: If we only aim to achieve client -> gateway -> fabric peer/orderer, fabric providers have already well prepared with firewall solution, satisfying `you only need to allow ingress for your own organization clients to your own Gateway node(s) `
Provider nowadays has gateway node design there already.

text/0000-fabric-gateway.md
davidkhala: A assumption here is grpc is easy to learn to use and have been widely adopted. 
Per my observation it is not realistic. Developer still has to do restful-grpc translation to interact with off chain partners. 

Nowadays, application based on gateway has also incorporate this translator in a rich middleware layer. If this RFC provide a new process in a docker container, then they either have to add another layer or choose not to use. 

Seemingly It is better to be an optional peer service.

text/0000-fabric-gateway.md
cendhu: As I said in the last line, pros/cons are associated with both approaches. I would still keep this gateway outside peer as I haven't see a strong reason for embedding this into peer. At a high level, I still do see this gateway for peer as pgbouncer/HAProxy for PostgreSQL. While coding for the ledger snapshot feature, at least I did feel that our dependency management is very complex and there are too many components to manage. Hence, only if it is absolute necessary, we need to embedd this into peer. If it can stay outside while adding some complexity to BaaS provider and duplicate certain part of the code, it is completely fine as it wouldn't create problem for us in the long term. This is my opinion.

text/0000-fabric-gateway.md
cendhu: In 2 PC, once the prepare succeeds, the commit will be successful. Not sure whether the same semantics apply here. The last two options look good to me.

text/0000-fabric-gateway.md
andrew-coleman: fixed - thanks

text/0000-fabric-gateway.md
andrew-coleman: I agree with Yacov, this will be far simpler architecturally to embed this in the peer.  It won't add too much overhead.  But as a component, it can be designed so that it can be run standalone, if preferred.  I've added this as a discussion point in the RFC

text/0000-fabric-gateway.md
andrew-coleman: I don't feel too strongly about the names we give to these GRPC functions.  The end users will be using the SDK terminology which is already well established

mastersingh24: I don't think we should introduce new terminology for these APIs .... 

I also believe that this should be built external to the peer but with thought it mind that it can be embedded in any other process.

andrew-coleman: > I don't think we should introduce new terminology for these APIs ....

I've changed the grpc API to Endorse/Submit

> I also believe that this should be built external to the peer but with thought it mind that it can be embedded in any other process.

Understood, that's not a problem.  The design will remain the same, it just affects the order in which we deliver the user-stories.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+    rpc Submit(PreparedTransaction) returns (stream Event) {}
+    rpc Evaluate(ProposedTransaction) returns (Result) {}
+}
+```
+
+The client credentials (private key) are never passed to the Gateway.  Client applications are responsible for managing their user keys (optionally using SDK Wallets) and signing the protobuf payloads.
+
+Submitting a transaction is a two step process (performed by the client SDK):
+- __Endorse__
+  - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTreansaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
```
harrisob: spelling of 'ProposedTreansaction'

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+  - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
+    - The Gateway will determine the endorsement plan for the requested chaincode and forward to the appropriate peers for endorsement. It will return to the client a `PreparedTransaction` message which contains a `Envelope` message as defined in `fabric-protos/common/common.proto`.  It will also contain other information, such as the return value of the chaincode function and the transaction ID so that the client doesn't necessarily need to unpack the `Envelope` payload.
+- __Submit__
+  - The client signs the hash of the Envelope payload using the user's private key and sets the signature field in the `Envelope` in the `PreparedTransaction`.
+  - The `Submit` service is invoked on the Gateway, passing the signed `PreparedTransaction` message.  A stream is opened to return multiple return values.
```
harrisob, line 64: A stream is much harder for the client to manage, the client will be waiting for the Gateway to close anyway, it would be easier if this was just a request-response (non-stream) call.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+    - The Gateway will register transaction event listeners for the given channel/txId.
+    - It will then broadcast the `Envelope` to the ordering service.
+    - The success/error response is passed back to the client in the stream
+    - The Gateway awaits suffient tx commit events before returning and closing the stream, indicating to the client that transaction has been committed.
+
```
harrisob, line 69: Should there be a title like heading here of `__Evaluate__` above the discussion on Evaluate similar to the endorse and submit.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+- The url of at least one peer in the org.  Once connected, the discovery service will be invoked to find other peers.
+- The MSPID associated with the gateway.
+- The signing identity of the gateway (cert and key), e.g. location of PEM files on disk.
+- The TLS certificate so that the gateway can connect to other components in the organization
+The following is an example command line invocation using a prototype:
+- `gateway -h peer0.org1.example.com -p 7051 -m Org1MSP -id ../../fabric-samples/fabcar/javascript/wallet/gateway.id -tlscert ../../fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/tlsca/tlsca.org1.example.com-cert.pem`
+
+#### Embedded in Peer
+When embedded in a peer, the gateway will register its gRPC service with the host peer's gRPC server.
+The gateway does not need its own (client) identity since it can get all the discovery information directly from the host peer rather than via a gRPC signed request.
```
harrisob: Without an identity, how will it connect to the event service of this peer or other peers

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+The client credentials (private key) are never passed to the Gateway.  Client applications are responsible for managing their user keys (optionally using SDK Wallets) and signing the protobuf payloads.
+
+Submitting a transaction is a two step process (performed by the client SDK):
+- __Endorse__
+  - The client creates a ProposedTransaction message, which contains a SignedProposal message as defined in `fabric-protos/peer/proposal.proto` and signed with the user's identity.  The ProposedTreansaction message also contains other optional properties, such as the endorsing orgs if the client wants to specify that.
+  - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
+    - The Gateway will determine the endorsement plan for the requested chaincode and forward to the appropriate peers for endorsement. It will return to the client a `PreparedTransaction` message which contains a `Envelope` message as defined in `fabric-protos/common/common.proto`.  It will also contain other information, such as the return value of the chaincode function and the transaction ID so that the client doesn't necessarily need to unpack the `Envelope` payload.
```
harrisob, line 61: Client will already have the transaction ID as it is in the "SignedProposal"

harrisob (commented):
Just to agree with some of the other comments that this will add another layer and added complexity.
Since the client is still dealing with the building and signing, does not seem like much of a savings on common code in that area.
Since the client is still dealing with checking the endorsement results and checking the submitted results, again does not seem like much code savings, (especially if dealing with stream results of the submit). 
The client will have to work with the new gRPC service and messages rather than the existing peer and orderer services with minimal savings.
Will the new client be able to work with "Events" ?

text/0000-fabric-gateway.md
bestbeforetoday: The stream interaction is proposed to allow the client to:
1. Send the transaction to Fabric Gateway.
1. Get a response and do follow-on work when the transaction is successfully delivered to the orderer.
1. Subsequently wait for Fabric Gateway to report successful commit of the transaction.

How would this flow work without streaming two separate response messages back from Fabric Gateway to the client?

text/0000-fabric-gateway.md
bestbeforetoday: Running as part of the peer process, Fabric Gateway would have direct access to the local peer ledger updates / events and discovery information. It could track the ledger height of other peers by directly accessing gossip data. It wouldn't connect to other peer eventing services.

bestbeforetoday:
> Just to agree with some of the other comments that this will add another layer and added complexity.
> Since the client is still dealing with the building and signing, does not seem like much of a savings on common code in that area.
> Since the client is still dealing with checking the endorsement results and checking the submitted results, again does not seem like much code savings, (especially if dealing with stream results of the submit).

I am expecting that a lot of the checking will take place in Fabric Gateway, not the client. The client application has the option of looking at proposal responses if it chooses but this is not necessary.

> The client will have to work with the new gRPC service and messages rather than the existing peer and orderer services with minimal savings.

The current Node SDK (only fabric-network and fabric-common) amounts to somewhere around 12K lines of code. The current Fabric Gateway Node.js client prototype enables fairly complete transaction evaluate and submit (no event listening so far) and amounts to somewhere around 325 lines of code. For sure a complete production-ready version will be larger, but even a 10x increase in client code volume would be a significant saving on what we have today, as well as moving much of the processing complexity (discovery, endorsement, eventing failover etc.) out of the client.

> Will the new client be able to work with "Events" ?

That is the intention. The mechanics of picking eventing peers and failover between them should be offloaded from the client to Fabric Gateway as much as practical.

davidkhala:
> > Just to agree with some of the other comments that this will add another layer and added complexity.
> > Since the client is still dealing with the building and signing, does not seem like much of a savings on common code in that area.
> > Since the client is still dealing with checking the endorsement results and checking the submitted results, again does not seem like much code savings, (especially if dealing with stream results of the submit).
> 
> I am expecting that a lot of the checking will take place in Fabric Gateway, not the client. The client application has the option of looking at proposal responses if it chooses but this is not necessary.
> 
> > The client will have to work with the new gRPC service and messages rather than the existing peer and orderer services with minimal savings.
> 
> The current Node SDK (only fabric-network and fabric-common) amounts to somewhere around 12K lines of code. The current Fabric Gateway Node.js client prototype enables fairly complete transaction evaluate and submit (no event listening so far) and amounts to somewhere around 325 lines of code. For sure a complete production-ready version will be larger, but even a 10x increase in client code volume would be a significant saving on what we have today, as well as moving much of the processing complexity (discovery, endorsement, eventing failover etc.) out of the client.
> 
> > Will the new client be able to work with "Events" ?
> 
> That is the intention. The mechanics of picking eventing peers and failover between them should be offloaded from the client to Fabric Gateway as much as practical.

And BTW, do we really need to provide peer failover in any Fabric helper project even in Fabric Gateway? For my experience, I prefer to inject failover handler codes in existing enterprise API gateway.

bestbeforetoday:
> And BTW, do we really need to provide peer failover in any Fabric helper project even in Fabric Gateway? For my experience, I prefer to inject failover handler codes in existing enterprise API gateway.

I am imagining that there will be a single endpoint URL to access Fabric Gateway (for a given organization), with dispatch and fail-over of client requests across Fabric Gateway instances handled at that endpoint, outside of client or Fabric Gateway code by an API gateway, load balancer, ingress controller etc.

For the Fabric Gateway implementation, as a separate component outside of the peer, it will need to establish gRPC streams to receive events from one or more peers. I suspect there will need to be some logic in Fabric Gateway to select appropriate eventing peers and failover to other peers, but it would be interesting to hear if there are alternatives approaches with low operational overhead.

text/0000-fabric-gateway.md
andrew-coleman: It dies, but it's buried in the SignedProposal and requires much code to access it. This, together with the chaincode return value is added at the top level for ease of access.  Other commonly used values could be added here too.  The idea is to simplify the developer experience by not making them wade through repeatedly marshalled nested protobuf structures.

text/0000-fabric-gateway.md
andrew-coleman: The double underscore in markdown just renders the text in bold font - these are not titles.  It's describing the two step process needed to implement the high-level SDK function` submitTransaction`.   The implementation of `evaluateTransaction` is described in the paragraph below that.

bestbeforetoday (approved):
👍

yacovm (approved):

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The high-level "gateway" programming model has proved to be very popular with users since its introduction in v1.4 in Node SDK.  Since then, it has been implemented in the Java and Go SDKs.  This has led to a large amount of code that has to be maintained in the SDKs which increases the cost of implementing new feature which have to be replicated to keep the SDKs in step.  It also increases the cost of creating an SDK for a new language.  Moving all of the high-level logic into its own process (or peer) and exposing a simplified gRPC interface will drastically reduce the complexity of the SDKs.
```
SamYuan1990, line 21: It means if this RFC passed, stop maintaining work of the high-level "gateway" programming model will be on schedule?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+  - The `Endorse` service is invoked on the Gateway, passing the ProposedTransaction message
+    - The Gateway will determine the endorsement plan for the requested chaincode and forward to the appropriate peers for endorsement. It will return to the client a `PreparedTransaction` message which contains a `Envelope` message as defined in `fabric-protos/common/common.proto`.  It will also contain other information, such as the return value of the chaincode function and the transaction ID so that the client doesn't necessarily need to unpack the `Envelope` payload.
+- __Submit__
+  - The client signs the hash of the Envelope payload using the user's private key and sets the signature field in the `Envelope` in the `PreparedTransaction`.
+  - The `Submit` service is invoked on the Gateway, passing the signed `PreparedTransaction` message.  A stream is opened to return multiple return values.
+    - The Gateway will register transaction event listeners for the given channel/txId.
```
C0rWin, line 65: I might have missed that, but I was wondering how custom chaincodes events will be supported? My understanding is that Gateway introduced to reduce burden and complexity from the clients where Gateway is to orchestrate transaction endorsement and submission, however, chaincode might result in event that client would be interested in. Are you assume that client will have to register to even service bypassing the new Gateway abstraction?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+Implementation options:
+- Only direct queries (and endorsement requests) to the peer(s) with the highest ledger height.
+- Client proposals include last observed ledger height so Gateway can target any peers with at least that ledger height.
+ A view of ledger height can be tracked by the Gateway from block events it has received.
+
+## Scaling and load balancing
+
+Currently a client application is responsible for load balancing its requests between multiple peers.  The Gateway will handle this on behalf of the set of currently connected client applications.
+
+Ideally, the Gateway itself will be stateless (i.e. not maintain client session state) allowing clients to be routed though an appriopriate load balancer.
```
C0rWin: What should be the suggested strategy to provide high availability of the Gateway service?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+The private data must be consistent with the state of the ledger after commit of the transaction that emitted the chaincode event, so the query must be processed by a peer with ledger height of at least the block number that contained the chaincode event.
+
+Implementation options:
+- Only direct queries (and endorsement requests) to the peer(s) with the highest ledger height.
+- Client proposals include last observed ledger height so Gateway can target any peers with at least that ledger height.
+ A view of ledger height can be tracked by the Gateway from block events it has received.
+
+## Scaling and load balancing
+
+Currently a client application is responsible for load balancing its requests between multiple peers.  The Gateway will handle this on behalf of the set of currently connected client applications.
```
C0rWin, line 130: Currently, client has a lot of freedom to prioritize the way transactions are balanced, basically having full control over it, do you plan to provide capabilities where client could dictate how to balance transactions?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+
+By providing a single entry point to a Fabric network, client applications can interact with complex network topologies without multiple ports having to be opened and secured to each peer and ordering node.  The client will connect only to the gateway which will act on its behalf with the rest of the network.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The Fabric Gateway is an embodiment of the high-level 'gateway' Fabric programming model in a server component that will form part of a Fabric network alongside Peers, Orderers and CAs.  It will either be stood up inside its own process in a docker container, or it will be hosted inside a peer.  Either way, it exposes its functionality to clients via a gRPC interface.
```
sykesm: No need to prescribe "in a docker container." A gateway process should not be constrained to that environment.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+
+Lightweight client SDKs are used to connect to the Gateway for the purpose of invoking transactions.  The gateway will intereact with the rest of the network on behalf of the client eliminating the need for the client application to connect directly to peers and orderers.
+
+The concepts and APIs will be familiar to Fabric programmers who are already using the new programming model.
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The Gateway runs as a client process associated with an organisation and requires an identity (issued by the org's CA) in order to interact with the discovery service and to register for events (Deliver Client).
```
sykesm, line 39: Can you add some explicit commentary regarding the trust assumptions? For example, discovery and the transaction ID extraction have implications.

What authentication and authorization will be present on the gateway services?

text/0000-fabric-gateway.md
sykesm: I think I agree with Bret. Why do we need to separate 2 and 3? I'd expect a client to want a single response that indicates the transaction disposition. Whether or not that operation is handled asynchronously should be a detail of the caller and not the API.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+    - It will then broadcast the `Envelope` to the ordering service.
+    - The success/error response is passed back to the client in the stream
+    - The Gateway awaits suffient tx commit events before returning and closing the stream, indicating to the client that transaction has been committed.
+
+Evaluating a transaction is a simple process of invoking the `Evaluate` service passing a SignedProposal message.  The Gateway passes the request to a peer of it's choosing according to a defined policy (probably same org, highest block count) and returns the chaincode function return value to the client.
+
+## Launching the Gateway
+
+#### Standalone process
+When running standalone, the Gateway is effectively a 'client' application to a set of peers.  It will be configured to connect to a peer in the organization in order to invoke discovery.  To do this it will need a signing identity in the channel writers policy.
```
sykesm: > in the organization 

I assume you're trying to say that the gateway is associated with an organization and it is configured to connect to a peer in the same organization?

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+
+To run the gateway server, the following parameters need to be supplied:
+- The url of at least one peer in the org.  Once connected, the discovery service will be invoked to find other peers.
```
sykesm, line 78: > the org

Similar comment; this should be a little more explicit.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+- Network
+- Contract
+- Transaction
+- Wallet
+
+The existing `submitTransaction()` and `evaluateTransaction()` API methods will be implemented using the Gateway gRPC services described above.
+
+Based on user feedback of the existing gateway SDKs, the following additions are planned to be supported by this implementation:
+- Support for 'offline signing', allowing clients to write REST servers that don't directly have access to users' private keys.
+- Support for 'asynchronous submit', allowing client applications to continue processing while the transaction commit proceeds in the background. For Go this will be implemented by returning a channel; for Node by returning a Promise; and for Java by returning a Future.
```
sykesm, line 164: Going back to my earlier comment, this means the asynchronous nature is intended to be part of the client and not the gateway service. Correct?

Unless we have a really good reason for returning streams from gRPC, it's best to avoid them.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+- Java SDK published to Maven Central.
+- Go SDK published to separate GitHub repo (e.g. fabric-gateway-go) and tagged.
+
+Contributors are invited to create SDKs for other languages, e.g. Python, Rust.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+When run as a standalone process, the gateway would add another docker containter to an (arguably) already complex Fabric architecture.  Also it will require its own signing identity in the channel writers policy in order to invoke the discovery service, which might be a security concern in some environments. These concerns are mitigated by embedding the gateway server within the organisation's existing peers.
```
sykesm: Please remove the "docker container" reference. None of this is bound to docker.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,213 @@
...
+- event handling scenarios
+- transient / private data
+- error handling
+
+The CI pipeline will run the unit tests for the Gateway and its SDKs as well as the end-to-end scenario (cucumber) tests against all SDKs.
+
+# Dependencies
+[dependencies]: #dependencies
+
+- This will depend on the peer implementation in `hyperledger/fabric-protos` and the protobuf definitions in `hyperledger/fabric-protos` repo.
```
sykesm: The peer implementation reference is incorrect.

sykesm (commented):
In addition to the comments on the text:

- Will there be health checks and metrics?
- What sort of authentication and authorization will be implemented on the service?

Thanks.

text/0000-fabric-gateway.md
bestbeforetoday: @sykesm The existing fabric programming model in the client SDKs has a single return point from a transaction submit which, by default, returns the transaction result (or an error) to the client application once the transaction has been committed (or the commit was unsuccessful). So the existing SDKs behaves as you describe:
```
result, err := contract.SubmitTransaction(txName, args...)
```

We have had numerous instances of client application developers wanting to:
1. Receive an acknowledgement when the transaction is successfully delivered to the orderer; then subsequently
1. Receive the transaction commit disposition.

Reasons cited for this requirement are that they want to:
- Provide a more responsive UI, where it is updated immediately after delivery to the orderer but can still coordinate follow-on work and/or UI updates after transaction commit.
- Where the Fabric client application is itself a REST server, to provide a more rapid RESTful response to their downstream clients immediately on delivery to the orderer as they might otherwise result in timed out connections, but still handle commit disposition.

Today the only way for client applications to achieve this behaviour is to do the transaction submit in a mode that does not wait for the transaction commit and returns the transaction result immediately after the transaction is delivered to the orderer. They then need to implement their own commit notification using eventing. This is significantly more work for the client application and requires the client to deal with the mechanics of transaction commits.

The intention of this streamed two-stage response from a transaction submit is to:
1. Provide a significantly simpler client application development experience where this decoupled behaviour is desired.
2. Allow commit handling to be implemented entirely on the server side of the connection, where it can be carried out much closer to the Fabric nodes (for improved performance) and in a single language (Go) rather than in multiple client-side languages.

Bear in mind that this streamed two-stage response is implemented at the gRPC level. Client applications are typically developed to the client-side SDK. In the client-side SDK, we can still expose the transaction invocation as a single method call that returns only on commit, in addition to a form that allows a two-stage response. So for the client application there is zero additional complexity associated with a streaming gRPC response. For example, something along these lines:
```
// Returns result only after commit
result, err := contract.SubmitTransaction(txName, args...)

// Returns result on delivery to orderer and signals subsequent commit on the returned channel
result, commitChannel, err := contract.SubmitAsync(txName, args...)
```

If we can achieve the same client application development simplicity and experience without using a streaming gRPC response, and also keep the commit handling implementation server-side, that would be great. How can we do that?

text/0000-fabric-gateway.md
bestbeforetoday: A client-side SDK based on Fabric Gateway would provide pretty much the same API provided by the current client SDKs. So client applications would still be able to listen for chaincode events exactly as they do today.

text/0000-fabric-gateway.md
bestbeforetoday: I imagine load balancing being implemented and configured in the Fabric Gateway server-side component, so managed by the Fabric network operator. To me personally, it makes more sense for the network operator to have control over how load is distributed across their network rather than having client applications burdened with the responsibility of managing load balancing.

text/0000-fabric-gateway.md
bestbeforetoday: > In addition to the comments on the text:
> 
> * Will there be health checks and metrics?

They both seem like nice things to have.

> * What sort of authentication and authorization will be implemented on the service?

https://github.com/hyperledger/fabric-rfcs/pull/32#discussion_r488820919

bestbeforetoday:
> In addition to the comments on the text:
> 
> * Will there be health checks and metrics?

They both seem like nice things to have.

> * What sort of authentication and authorization will be implemented on the service?

https://github.com/hyperledger/fabric-rfcs/pull/32#discussion_r488820919

text/0000-fabric-gateway.md
bestbeforetoday: https://github.com/hyperledger/fabric-rfcs/pull/32#discussion_r520477985

text/0000-fabric-gateway.md
bestbeforetoday: I think the intention is for a Fabric Gateway client SDK to provide pretty much the same API as the existing high-level Fabric programming model. Moving as much of the implementation as practical out of the client-side SDK implementation into a common Fabric Gateway component just helps to reduce the amount of effort required to maintain the "gateway" programming model across multiple client-side languages.

text/0000-fabric-gateway.md
sykesm: My perspective is that the suggested approach is on the cusp of abusing streamed responses. A gRPC stream response is generally intended to iterate over a result set or deliver a stream of events. In this case, it looks like the event stream is actually one event (completion).

Unless I'm misreading, the response from Submit can be delayed until the transaction has been submitted (meeting the first requirement) and the client can then use the response to implement a wait for completion.

If `Event` actually means _n_ events are going to come back from the gateway where _n_ is generally greater than 2, I may not have the same reaction.

sykesm:
> > * Will there be health checks and metrics?
> 
> They both seem like nice things to have.

They are needed for operations and there should at least be a statement about them.

> > * What sort of authentication and authorization will be implemented on the service?
> 
> [#32 (comment)](https://github.com/hyperledger/fabric-rfcs/pull/32#discussion_r488820919)

Unfortunately, that's not the same as adding information to the RFC about the trust assumptions and the proposed authen/authz model and requirements.

text/0000-fabric-gateway.md
bestbeforetoday: At the server end you could use a load balancer at the Fabric Gateway endpoint, which would transparently dispatch incoming requests across a set of Fabric Gateway instances.

At the client end you could use a load-balancing gRPC client connection when connecting the Gateway client, which would transparently dispatch requests across a set of Fabric Gateway endpoints.

bestbeforetoday:
> > > * What sort of authentication and authorization will be implemented on the service?
> > 
> > [#32 (comment)](https://github.com/hyperledger/fabric-rfcs/pull/32#discussion_r488820919)
> 
> Unfortunately, that's not the same as adding information to the RFC about the trust assumptions and the proposed authen/authz model and requirements.

I see Fabric Gateway as essentially being an application-aware proxy. It just routes signed messages between the client and appropriate peer/orderer nodes. It does not participate in authn/authz at the message level. That is handled by peer/orderer nodes exactly as if the client was connecting to them directly.

Fabric Gateway does need to have an identity, which I think is already described in the RFC, that allows it to connect to peers as a client in its own right to:
1. Do network discovery
2. Listen to events (assuming there is agreement to allow Fabric Gateway to check for commit disposition of submitted transactions that it has routed to the ordering service on behalf of a client)

I agree it would be helpful to clarify this in the RFC.

sykesm (approved):

cendhu (approved):

lesleyannjordan:
Mon 16/11/2020 16:29
In accordance with the RFC process (https://github.com/hyperledger/fabric-rfcs#the-rfc-life-cycle), the RFC for 'Fabric Gateway' has had sufficient discussion and received enough maintainer approvals to qualify for entry to Final Comment Period (FCP).

This is to announce the start of FCP, which will last one week.  Please make any final comments directly to the RFC PR (https://github.com/hyperledger/fabric-rfcs/pull/32).

Many thanks,
Andrew Coleman

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,233 @@
...
+The gateway does not need its own (client) identity since it can get all the discovery and event information directly from the host peer rather than via a gRPC signed request.
+
+### Discovery
+
+On startup, the gateway will connect to its primary peer(s), as specified in the command line option (or its host, if embedded in a peer).  The discovery service will be invoked to build and maintain a cache of the network topology per channel.  This will be used to identify endorsers based on the chaincode endorsement policy.
+
+#### Collection endorsement policy
+
+Discovery can be used to get endorsing peers for a given chaincode where the transaction function will write to a given collection, and this will take into account the collection level endorsement policy.
+Ideally we could take care of these endorsement requirements in the Gateway without the client needing to have any implicit knowledge of them. Suggest extending contract metadata to include information on collections, which can be queried by Fabric Gateway. If that isn't possible then we might need some way for the client to specify the collections that will be used (not desirable).
```
denyeart, line 96: Collections are typically not static in a chaincode, they depend on the transaction which means the client must be able to provide them.  For example if I am a client doing a transaction with you, I want the data to go to my collection, your collection, or a bilateral collection between us.

text/0000-fabric-gateway.md
denyeart: We've been talking about a new authentication service. We used mutual TLS for the new orderer channel management participation api as a temporary solution, with the expectation that we'd shift to a token-based authentication service as the strategic solution. I expect we would want to do the same for gateway. Look for a separate RFC in this space.

text/0000-fabric-gateway.md
andrew-coleman: To conclude this discussion (including an offline discussion with @sykesm), we have decided to stay with the proposed Submit function that returns a stream.  In the spirit of agile development, if implementation experience requires us to revisit this, then we can do so before this component is formally released.

denyeart (approved):

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,233 @@
...
+
+### State based endorsement
+
+This is driven by business logic and rules. Requires the client to supply constraints on endorsing organisations as part of the proposal endorsement.
+
+### Chaincode-to-chaincode endorsement
+
+Augment contract metadata for transactions with additional chaincodes / transactions that will be invoked, then we use discovery to narrow the set of endorsers.
```
davidkhala, line 104: I wonder how this is implemented today. Thanks in advance if a refence link to docs/git commit can be shared here.

text/0000-fabric-gateway.md
```diff
@@ -0,0 +1,233 @@
...
+- Node (Typescript/Javascript)
+- Go
+- Java
+
+It is proposed that these SDKs will be maintained in the same GitHub repository as the gateway itself (`hyperledger/fabric-gateway`) to ensure that they all stay up to date with each other and with the core gateway.  This will be enforced in the CI pipeline by running the end-to-end scenario tests across all SDKs.
+
+Publication of SDK releases will be as follows:
+- Node SDK published to NPM.
+- Java SDK published to Maven Central.
+- Go SDK published to separate GitHub repo (e.g. fabric-gateway-go) and tagged.
```
davidkhala, line 176: Are we still planning to have another github repo for Go SDK release like hyperledger/fabric-gateway-go?
I see `module github.com/hyperledger/fabric-gateway` is also enough if not conflicting with other language SDK.

text/0000-fabric-gateway.md
bestbeforetoday: The Gateway automagically identifies the endorsement requirements by inspecting the read/write sets produced from an initial simulation (endorsement) of the transaction. See here for details:

https://hyperledger-fabric.readthedocs.io/en/release-2.4/gateway.html#how-the-gateway-endorses-your-transaction-proposal

text/0000-fabric-gateway.md
bestbeforetoday: The source repository (hyperledger/fabric-gateway) is the Go module location. No plans anymore to create a mirror repository to host only the published Go code.

PR #36, opened 2020-11-11
Creates the RSA feature request

filipeandre:
This RFC re-introduce RSA crypto support

### DIFF
--- a/text/0000-fabric-rsa-alg.md
+++ b/text/0000-fabric-rsa-alg.md
@@ -76,6 +76,8 @@ RSA testing has been made in the past:
 * [b17e846](https://github.com/hyperledger/fabric/commit/b17e846) [FAB-3441](https://jira.hyperledger.org/browse/FAB-3441) bccsp/sw ECDSA/RSA sign test coverage
 * [7aa43d5](https://github.com/hyperledger/fabric/commit/7aa43d5) [FAB-3441](https://jira.hyperledger.org/browse/FAB-3441) bccsp/sw ECDSA/RSA verify test coverage
 
+In order to test this end to end is also required an integration test that instantiates Fabric nodes with RSA certificates.
+
 # Dependencies
 [dependencies]: #dependencies
 
--- a/text/0000-fabric-rsa-alg.md
+++ b/text/0000-fabric-rsa-alg.md
@@ -1,0 +1,88 @@
+---
+layout: default
+title: RSA Algorithm Support
+nav_order: 3
+---
+
+- Feature Name: Re-introduce RSA crypto support
+- Start Date: 2020-11-11
+- RFC PR: 
+- Fabric Component: bccsp, crypto
+- Fabric Issue: 
+
+# Summary
+[summary]: #summary
+
+Fabric had RSA crypto feature implemented since the very beginning even if there was a mismatch between the code and the documentation.
+With the introduction of Fabric 2.x the RSA code was removed, so currently existing production ledgers with transactions signed by RSA no longer could be validated after upgrade to the new version.  Original RSA support commits:
+
+* [35af475](https://github.com/hyperledger/fabric/commit/35af475) BCCSP support for RSA signing
+* [2f7153f](https://github.com/hyperledger/fabric/commit/2f7153f) BCCSP ECDSA/RSA/X509 public/private key import
+
+This RFC was created in response to this issue:
+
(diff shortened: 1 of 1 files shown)


text/0000-fabric-rsa-alg.md
```diff
@@ -0,0 +1,88 @@
...
+
+Not doing this will make businesses that are already using Fabric with RSA in production to start their own forks and have to maintain those themselves.
+
+# Prior art
+[prior-art]: #prior-art
+
+RSA is the most used, studied and tested algorithm proven to be secure. References to the original implementation are in the summary above.
+
+# Testing
+[testing]: #testing
```
yacovm, line 73: I think that it should be required to add an integration test that instantiates Fabric nodes with RSA certificates in order to test this end-to-end.

yacovm:
I am personally in favor. 

Pros: 

- RSA has faster verification which matters in blockchains
- It is safe to assume (and I actually know of some cases) that there are more users of Fabric 1.4.x that made the innocent mistake of using RSA and bringing RSA back would let them upgrade to Fabric 2.x

Cons:
- RSA has larger keys so transactions will be even bigger than they are now

text/0000-fabric-rsa-alg.md
filipeandre: I updated the RFC with this test requirement.

text/0000-fabric-rsa-alg.md
```diff
@@ -0,0 +1,90 @@
...
+- Feature Name: Re-introduce RSA crypto support
+- Start Date: 2020-11-11
+- RFC PR: 
+- Fabric Component: bccsp, crypto
+- Fabric Issue: 
+
+# Summary
+[summary]: #summary
+
+Fabric had RSA crypto feature implemented since the very beginning even if there was a mismatch between the code and the documentation.
```
sykesm, line 16: Please see https://github.com/hyperledger/fabric/pull/2069#issuecomment-726815751 where there are references to Fabric code that demonstrate why RSA is not supported within Fabric transaction processing.

text/0000-fabric-rsa-alg.md
```diff
@@ -0,0 +1,90 @@
...
+With the introduction of Fabric 2.x the RSA code was removed, so currently existing production ledgers with transactions signed by RSA no longer could be validated after upgrade to the new version.  Original RSA support commits:
```
sykesm, line 17: Given the code references cited above, I don't see how transactions could be signed and verified with RSA.

text/0000-fabric-rsa-alg.md
sykesm: I created an end-to-end test on release-1.4 with RSA crypto and nothing worked. The only bit I found to work was the use of RSA public keys in CA certificates.

sykesm (commented):
If RSA signing and verification is desired, an RFC is the correct way to propose it and gather feedback on how to implement it, so thank you for raising this.

That said, this RFC is written in a such a way that it implies RSA was previously supported and functional yet the evidence in the code implies it was not. While BCCSP did have RSA signing and verification implementations, the call sites did not provide the signing options to bccsp that are required for it to work.

I have created pull requests to restore the ability to use an RSA CA in Fabric MSPs for the release-2.x stream. I expect (hope) those changes to resolve the migration issues that were raised as part of https://github.com/hyperledger/fabric/pull/2069 and that motivated the creation of this RFC.

denyeart:
Given that RSA CA was added to v2.3.0 and upcoming v2.2.2, and the fix was confirmed by @filipeandre , I believe this RFC can now be closed. Here's the text from the v2.3.0 release note:

FAB-18308: peer and orderer - Restore support for MSPs that contain RSA certificate authorities

While Fabric has never supported RSA for transaction signatures or validation,
certificate authorities included in MSP definitions could be associated with
RSA keys. This ability was inadvertently removed during the development of
release 2.0 and prevented migration of some networks to a 2.x version. With
these changes, version 2.x components will no longer panic when attempting to
initialize MSPs that include CA certificates associated with RSA keys.

davidkhala:
since 
`While Fabric has never supported RSA for transaction signatures or validation,`
@denyeart 
Could we reopen this RFC to keep track for RSA transaction signature support?

sykesm:
> Could we reopen this RFC to keep track for RSA transaction signature support?

It looks like there is a desire for RSA in the larger community but, IMO, the structure of this RFC was built around assumptions that did not hold. Further, an RFC to add RSA should really be more than pointers to old designs and a plan to restore bits of the BCCSP implementation.

Instead of reopening this RFC, I believe a new RFC should be produced that takes the existing implementation into consideration. Ideally this would be done in a way that it addresses other requirements related to hashing, signing, and validation beyond RSA.

PR #37, opened 2020-11-12
Update channel participation motivation

denyeart:
Expand on the operational issues related to management
and dependence on the system channel.


text/0004-channel-participation-api-without-system-channel.md
```diff
@@ -50,14 +50,21 @@ starts by scanning the system channel in order to discover existing channels and
...
 * Channel creation (using the system channel) utilizes information contained in the system channel - e.g. MSP information
-of the organization - in order to apply it to the newly created channel. This creates the convoluted situation in
+of the organization - in order to apply it to the newly created channel. This creates an unnecessary and convoluted situation in
 which an organization that wants to update its details in the consortium must coordinate that action with third parties
 that have write privileges to the system channel.
 
+* Updating the system channel has been a challenge for users because it works differently than application channels
+(MSP definitions are in different paths, the config structure is slightly different, etc).
+
+* The dependence on the system channel makes creating a channel unnecessarily complex and fragile, for example the need to pick subsets of consenters.
```
jyellick: for example, the channel creation transaction generation must assume the current state of system channel and the template channel configuration it will create.  This leads to relatively simple use cases like specifying a subset of consenters at channel creation being difficult and error prone.

text/0004-channel-participation-api-without-system-channel.md
```diff
@@ -50,14 +50,21 @@ starts by scanning the system channel in order to discover existing channels and
 channels increase, this scan prolongs the process of adding a new orderer.  
 
 ## Operational problems
-There are several operational problems that affect large scale deployment and hinder efficient management of cloud
-environments using automatic tooling.
+The dependence on a system channel causes channel management to become unnecessarily fragile and complicated, impacting
+large scale deployment and hindering efficient management of cloud environments using automatic tooling.
+
+* It is not possible to spin up ordering service nodes without first coordinating on a system channel genesis block.
```
jyellick: , which in turn requires that the other bootstrap parameters of the ordering service nodes have already been generated.

tock-ibm (approved):

text/0004-channel-participation-api-without-system-channel.md
denyeart: @jyellick I made the update, but I don't fully understand it... for my own education, what is an example of a bootstrap parameter that must be generated.

sykesm (approved):

C0rWin (approved):

text/0004-channel-participation-api-without-system-channel.md
jyellick: For instance, it is typical model for deployment is that TLS certificates are generated at startup if they do not exist.  But, the orderer TLS cert is a parameter to the channel configuration consensus parameters.  So, with a system channel, we must first generate all of the orderer TLS artifacts before starting any orderer (because we need those artifacts to build the system channel).  With the channel participation API, we can let orderers start and come up fully before collecting those artifacts to build a channel configuration.

PR #41, opened 2021-01-13
JWT authentication proposal for Fabric

sykesm:


robmurgai:
@sykesm , is it standard practice to have the expires_in value default to 300 seconds or less, or is a judgement call behind this?

sykesm:
> @sykesm , is it standard practice to have the expires_in value default to 300 seconds or less, or is a judgement call behind this?

Assuming you're referring to the statement
```
the `expires_in` value should be some small, configurable duration that defaults to 300 seconds or less
```
It's a judgement call. The goal is to have a sane default that is long enough to be usable and short enough to mitigate concerns about stolen tokens. It should be configurable so people can override it as appropriate.

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+- Start Date: 2021-01-11
+- RFC PR:
+- Fabric Component: orderer, peer
+- Fabric Issue:
+
+# Summary
+
+This document proposes an OAuth2 compatible authentication mechanism for MSP
+identities that produces authorization tokens suitable for use with APIs
+exposed by Fabric.
```
davidkhala, line 17: @sykesm  Dear Matt, you mentioned about channel participation API in motivation and in later section I see `The authentication mechanism described here can also be applied to gRPC services to support activities like chaincode install and query operations.`  If this RFC aims to provide one-off solution to provide ALL mutual TLS APIs with another authentication mechanism, that is more welcomed. Otherwise web developers will face a complex situation that “Sometime we A only, sometime we can B or A, since Fabric is transforming but on the halfway”

sykesm: The target for this work is the channel participation API. It is the only REST (not gRPC) API that we have that influences data processing in the network.

davidkhala: So this work target will not including roadmap about gRPC endpoint auth. Is it？

sykesm: This RFC is explicitly targeted at the channel participation API and I'd like it to remain focused on that. A narrow scope keeps it containable and simple.

While the same technology can be used to authenticate with gRPC services, that is not in scope.

>  aims to provide one-off solution

I also want to say that I disagree with this characterization. We have gone out of our way to build on and around standard protocols to avoid this becoming a "one-off" solution.

jyellick:
All seems good to me.

denyeart:
@sykesm So the traditional client certificate is passed in the first request in order to get a token back? The client certificate is validated by the orderer MSP at that time?
Token based authorization makes sense for chatty interactions so that you don't have to re-validate the client cert each time. But in the case of participation API, it is not chatty, they are usually one and done type interactions. So my one concern is that we still need to pass and validate the client certificate (assuming the answer to my first questions is yes), plus all the additional new infrastructure for the token. Does the JWT approach bring value to the participation api by itself, or does the main benefit come in the future when this is applied to other more chatty interactions?

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+The `exp` time in the token should be checked against the local clock. If the
+expiry is in the past, the assertion must not be accepted. If the expiry time
+is more than some threshold in the future, the assertion must not be accepted.
+
+The `iat` time in the token should be checked against the local clock. If the
+token was issued more than some threshold in the past, the assertion must not
+be accepted. The window between `iat` and `exp` should represent a reasonable
+window of time for authentication - typically less than a minute or two.
+
+The `jti` is generated by the client and is intended to act as a nonce to
```
yacovm, line 157: can you elaborate on the replay attack? Who is replaying to who?
Also, what if the JTI is not a UUID but it's a timestamp and the peer simply rejects if the timestamp is too far (in absolute value) from its clock?

sykesm: > can you elaborate on the replay attack? Who is replaying to who?

The client authentication token is a signed assertion. Without `jti` the same token can be used multiple times to acquire access tokens.

Additionally, the `jti` is a required field for open ID connect client authentication and optional in RFC7523. These are the standards that this proposal builds on.

> jti
> REQUIRED. JWT ID. A unique identifier for the token, which can be used to prevent reuse of the token. These tokens MUST only be used once

>  The JWT MAY contain a "jti" (JWT ID) claim that provides a unique identifier for the token.  The authorization server MAY ensure that JWTs are not replayed by maintaining the set of used "jti" values for the length of time for which the JWT would be considered valid based on the applicable "exp" instant.

---

> Also, what if the JTI is not a UUID but it's a timestamp and the peer simply rejects if the timestamp is too far (in absolute value) from its clock?

The `nbf`, `exp`, and `iat` claims control the time window of validity; `exp` is required, `nbf` and `iat` are optional. `iat` and `exp` are referenced in the proposal already and `nbf` is enforced by token validation libraries.

sykesm:
> @sykesm So the traditional client certificate is passed in the first request in order to get a token back? 

Not a traditional (TLS) client certificate but a signing certificate representing an MSP identity. 

> Token based authorization makes sense for chatty interactions so that you don't have to re-validate the client cert each time. But in the case of participation API, it is not chatty, they are usually one and done type interactions. So my one concern is that we still need to pass and validate the client certificate (assuming the answer to my first questions is yes), plus all the additional new infrastructure for the token. Does the JWT approach bring value to the participation api by itself, or does the main benefit come in the future when this is applied to other more chatty interactions?

I tried to enumerate the value in the motivation section: to move away from TLS mutual auth and build on the current MSP infrastructure.

By using an authentication flow (ie. login) to create an access token, we get the immediate benefit of controlling how long an access token should be considered valid. As we look to future use cases, a separate authentication flow allows us to change the structure of our access token, add more attributes, and reuse the token across nodes. (Take a look at the fallout related to the authorization token changes from FABC-174 to see why this is important.)

text/0000-api-jwt-authentication.md
yacovm: > The client authentication token is a signed assertion. Without jti the same token can be used multiple times to acquire access tokens.


But my question is - what scenario are you trying to protect from? Is it a case where a client had access rights, and later on has been revoked? Or that the token was persisted to disk and compromised?

yacovm:
> Does the JWT approach bring value to the participation api by itself, or does the main benefit come in the future when this is applied to other more chatty interactions?

Using token based access control can be beneficial to service discovery, as it requires you to periodically probe a peer to obtain information.

text/0000-api-jwt-authentication.md
sykesm: My goal is to protect against a compromised authentication token. Even if that weren't the case, the oidc spec requires the `jti` and states that the tokens must be used once.

Just to make sure we're on the same page, the client assertion JWT is used for authentication and not used as the access token...

denyeart:
@sykesm  I was probing relative to the traditional Fabric approach, not the mutual TLS approach.
The traditional Fabric approach is to pass a signing certificate representing an MSP identity.
The proposed approach is to pass a signing certificate representing an MSP identity on the first interaction, and then use a token for subsequent interactions.
At first glance the traditional Fabric approach seems simpler. What are the advantages of the token approach? Obviously there are advantages with chatty interactions. What are the advantages for non-chatty interactions such as those with participation api? That's what I'd like to see in the motivation.

sykesm:
> The traditional Fabric approach is to pass a signing certificate representing an MSP identity.  At first glance the traditional Fabric approach seems simpler.

How so? Instead of signing a message once, the "traditional" approach envelopes and signs every message that is sent. This is not a typical REST API interaction with a negotiated content type of `application/json`.

> What are the advantages of the token approach? Obviously there are advantages with chatty interactions. What are the advantages for non-chatty interactions such as those with participation api? That's what I'd like to see in the motivation.

I feel like I've answered this question. Chattiness is irrelevant. Authentication allows us to control the scope and lifetime of an access token and removes the requirement to sign every message by replacing it with a login flow.

yacovm:
> > The traditional Fabric approach is to pass a signing certificate representing an MSP identity.  At first glance the traditional Fabric approach seems simpler.
> 
> How so? Instead of signing a message once, the "traditional" approach envelopes and signs every message that is sent. This is not a typical REST API interaction with a negotiated content type of `application/json`.
> 
> > What are the advantages of the token approach? Obviously there are advantages with chatty interactions. What are the advantages for non-chatty interactions such as those with participation api? That's what I'd like to see in the motivation.
> 
> I feel like I've answered this question. Chattiness is irrelevant. Authentication allows us to control the scope and lifetime of an access token and removes the requirement to sign every message by replacing it with a login flow.

Let's say you want to join an orderer to a channel. 

-  In the "traditional" approach you would create a config block, add your identity, marshal it, sign it, and send it to the orderer.

- In the new approach you would create a request for privileges, add your identity, marshal it, sign it, and then send it to the orderer from which you obtain an access token. 
Now you can use this access token to do operations, such as join an orderer to a channel. But this means you need 2 phases and not one, and you need to pass information from phase 1 to phase 2, while the "traditional" approach has a single phase and no state on the client side nor the server side. 

I don't think the fact that it's a "typical REST API" interaction has any benefit in its own, since adopting a new flow/pattern needs justification.
However, there is some benefit for adopting this approach and that is - that you're using a protocol and an encoding that is widely used and thus it's easy to integrate it into different programming languages and platforms, while in the "traditional" approach you would need to write custom code that handles protobuf marshaling and unmarshaling, and here a dependency at hand is doing it for you. 

Also - the token based approach is more efficient when you need to query data over and over again like a GET on the channel participation API, or pretty much any discovery query, since in the traditional approach you would need to validate each request (verify a signature) over and over again (discovery has a trick that prevents that but that's another matter).

sykesm:
> encoding that is widely used and thus it's easy to integrate it into different programming languages and platforms, while in the "traditional" approach you would need to write custom code that handles protobuf marshaling and unmarshaling, and here a dependency at hand is doing it for you.

This is exactly what I was trying to say when I referenced the content-encoding. This keeps the data interchange format clean and consistent with the expectations of client authors.

denyeart:
Thanks Yacov, that answers the question and is what I am hoping gets included in the motivation section.

>"create a request for privileges, add your identity, marshal it, sign it, and then send it to the orderer from which you obtain an access token"

From a client perspective, is this part (phase 1) also standard? And therefore easier supported than the single phase in the traditional Fabric interaction pattern.

yacovm:
> Thanks Yacov, that answers the question and is what I am hoping gets included in the motivation section.
> 
> > "create a request for privileges, add your identity, marshal it, sign it, and then send it to the orderer from which you obtain an access token"
> 
> From a client perspective, is this part (phase 1) also standard? And therefore easier supported than the single phase in the traditional Fabric interaction pattern.

In some web services you do something else - you attach an HTTP header to your request which contains an _Authorization_ key and a value which is can vary, for example it can be the _username:password_ or maybe something else. 
Then you might get back an access token in the form of a session ID which is the token in our story.

However in the RFC it says:

> This `client_assertion` can then be presented to a resource server's `/token` endpoint to acquire an access token.

From which I deduce that you first obtain a token and then use it, and you're not expected to send the credential as part of your real intent of joining an OSN to a channel.

In theory, I would say that it's possible to use HTTP pipelining and send first the /token request and then immediately the channel join request and the server should realize they come from the same TLS connection and therefore let the 2nd request in, and respond with a token and then later with the response of the 2nd request.

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+
+## OpenID Connect Authentication
+
+OpenID Connect is an identity layer that sits on top of the OAuth 2.0
+protocol. Part of the specification describes how to verify the identify of an
+end-user by authenticating to an authorization server.
+
+The [`private_key_jwt`][private_key_jwt] client authentication method enables
+a client to use a private key associated with a registered public key to sign
+an authentication token and assert its own identity. This `client_assertion`
```
denyeart, line 53: Let's expand this paragraph and tie it back to Fabric MSP client credentials. For example is the 'registered public key' the client's MSP credentials? I assume it is, and I assume the client certificate will be passed in this request, however this paragraph doesn't mention that.

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+Claims, and the Javascript Web Signature (JWS) signature.
+
+#### JOSE Header
+
+The header in the `client_assertion` token indicates the type of the token,
+the algorithm used to sign the token, and a reference to the public key that
+can be used to verify the token signature. In Fabric, this means that `typ`
+must be set to `JWT`, `alg` must be `ES256` for ECDSA-P256 certificates (the
+only type we'll support initially), and `x5c` must be set to the certificate
+or certificate chain of the client in the form described by
```
denyeart, line 114: Again, let's put in Fabric terms... is this certificate the client MSP certificate?

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+- `jti` - a unique token identifier that helps prevent replay attacks
+
+The `aud` value provided by the client must match the URL of the token
+endpoint. Because of proxying, the expected value(s) may need to be configured
+on the server.
+
+The `iss` and `sub` fields declare the `client_id` that is being asserted.
+Both fields must contain the same value and should be the _MSPID_ that issued
+the certificate. This information allows us to validate the certificate
+against the correct MSP.
```
denyeart, line 146: Wouldn't the participation API always evaluate against the orderer nodes local MSP? Since we're talking about admin actions against a specific orderer node.

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+maintaining conventional REST interaction patterns without explicit message
+hashing and signing.
+
+There are two minor issues with this approach:
+
+1. The client CA certificate used to control access to the channel
+   participation API is disconnected from the orderer's local MSP.
+2. It's often difficult for web applications to connect to servers with TLS
+   mutual authentication.
+
```
denyeart, line 33: The mutual TLS approach was known to be an interim solution.
I think it would be even more valuable to discuss the benefits relative to the traditional Fabric authentication approach (passing a client certificate in every interaction).

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+This goal of this proposal is to describe an authentication flow based on
+[RFC7523][RFC7523] that can be used to obtain an access token for the
+channel participation API from an MSP identity.
+
+# Guide-level Explanation
+
+Fabric supports an authentication mechanism based on OAuth 2.0 and OpenID
```
denyeart, line 40: This makes it sound like Fabric already supports it.

text/0000-api-jwt-authentication.md
```diff
@@ -0,0 +1,322 @@
...
+The header in the `client_assertion` token indicates the type of the token,
+the algorithm used to sign the token, and a reference to the public key that
+can be used to verify the token signature. In Fabric, this means that `typ`
+must be set to `JWT`, `alg` must be `ES256` for ECDSA-P256 certificates (the
+only type we'll support initially), and `x5c` must be set to the certificate
+or certificate chain of the client in the form described by
+[RFC7517][RFC7517].
+
+The following is an example of a valid header that includes an ECDSA-P256
+certificate:
```
denyeart, line 118: Again, is this the client MSP certificate?

text/0000-api-jwt-authentication.md
sykesm: That's because the template says:

> Explain the proposal as if it was already included in Fabric and you were teaching it to another Fabric developer.

sykesm:
After some offline discussions, my understanding is that we do not have the appetite to replace Mutual-TLS with an authentication flow and a JWT credential at this time. Perhaps this can be revived in the future.

PR #42, opened 2021-01-26
Add RFC for OpenTelemetry integration

atoulme:
Signed-off-by: Antoine Toulme <antoine@lunar-ocean.com>

This RFC offers a path to integrate OpenTelemetry, and specifically adding tracing ~and metrics~ reporting capabilities, to client SDKs and chaincode shim frameworks. This RFC relies and references on the work done in a [pull request](https://github.com/hyperledger/fabric-chaincode-java/pull/153) made to the Java shim chaincode repository.

Feedback and comments are welcome and encouraged. Thanks!

### DIFF
--- a/text/0000-opentelemetry.md
+++ b/text/0000-opentelemetry.md
@@ -14,7 +14,7 @@ nav_order: 3
 [summary]: #summary
 
 This request for comments proposes integrating Hyperledger Fabric, its SDKs, core and chaincode components 
-with the OpenTelemetry project. It advocates for relying on OpenTelemetry to record and send metrics.
+with the OpenTelemetry project.
 It introduces the concept of tracing, reporting the execution of chaincode, core components and SDKs to help correlate activities with the chain.
 
 # Motivation
@@ -22,26 +22,22 @@ It introduces the concept of tracing, reporting the execution of chaincode, core
 
 The OpenTelemetry project is a CNCF project that aims to standardize observability, especially in a cloud-native environment.
 OpenTelemetry is working to create a standard around representing metrics, traces and logs.
-We aim to bring full observability of Hyperledger Fabric to allow tracing of transactions all the way from the client
-to each of the chaincode invocations. We also want to capture chaincode interactions.
-We want to be able to trace and correlate actions from chaincode events.
+We aim to bring full observability of Hyperledger Fabric to allow tracing of transactions all the way from the SDK to each of the chaincode invocations.
 
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-In this guide, we will cover the notion of observability. Observability comes from the DevOps world, and is defined in
-[Wikipedia](https://en.wikipedia.org/wiki/Observability) as "a measure of how well internal states of a system 
+Observability comes from the DevOps world, and is defined in [Wikipedia](https://en.wikipedia.org/wiki/Observability) as "a measure of how well internal states of a system 
 can be inferred from knowledge of its external outputs".
 Software that is observable exposes its behaviors by sending traces, which represent the path of execution of a particular request.
 Each trace can be represented as a set of spans that are correlated as child/parent or sequential relationships.
 Observable software is exposing metrics, representing the internal state of the components, as well as logs, emitted from the execution of the software.
 
-In Hyperledger Fabric, we rely on the OpenTelemetry framework to report traces, metrics and logs, and correlate them across services.
+In Hyperledger Fabric, we rely on the OpenTelemetry framework to report traces and correlate them across services.
 
-In practice, this means a request made by a client connected via the Fabric SDK to a node can pass on the context of a trace. 
-When chaincode executes, it can correlate to that trace, or create its own, and report to a collector storing this information.
-Chaincode can also expose traces in chaincode events. Any code triggered from such an event can then use that trace to report its origin.
+In practice, this means a request made by a client connected via the Fabric SDK to a node can pass on the context of a trace.
+Peers and orderers propagate the trace context and create spans indicating their own interaction.
 
 Blockchain operators can reconstitute a graph of the interaction of all the components at play to create a service map.
 This helps uncover trends and issues of performance, as well as shortening the time it takes to investigate problems.
@@ -50,17 +46,9 @@ It offers some security capabilities, such as detecting unexpected executions, o
 
 Hyperledger Fabric developers can take advantage of those techniques with no code changes to their existing chaincode deployments.
 Each chaincode execution generates a top-level trace and will report to an endpoint provided by the environment.
-They may desire to build additional spans to capture additional data, or report errors. The spans they create will 
-correlate to the parent span using the local thread context.
 
 Developers may also create a trace before calling out to Fabric in their client code.
 The current trace information will be sent along with the message to the peer.
-
-Finally, developers may like to report metrics as part of chaincode execution. They can use OpenTelemetry counters,
-gauges and other metrics constructs to report metric data.
-The metric endpoint is configured when the chaincode is deployed. 
-In the case of external chaincode, the developers can set standard [OpenTelemetry variables](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md)
-to point at an OpenTelemetry collector instance.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -69,18 +57,6 @@ to point at an OpenTelemetry collector instance.
 
 The trace information is passed in as an optional gRPC metadata header.
 Trace headers are filled in by the SDK and read by the chaincode execution framework.
-
-## Changes to the chaincode shims
-
-Each shim implementation should be modified to capture the trace ID passed in as a gRPC metadata header.
-If no parent trace information is present, the shim continues and creates a new trace with no parent relationship.
-
-As a demonstration, in this [PR](https://github.com/hyperledger/fabric-chaincode-java/pull/153/files#diff-db5cb40a9966f929cdf9963d606b8961248f94aac09ed150491fdc41e73d82b9R94), the Java shim implementation `call()` method which calls the chaincode execution changes to embed a try/finally statement
-that creates and ends the span once execution is finished. It also captures exceptions thrown during execution to set the status of the trace as an error.
-
-The Java chaincode shim also changes to implement the MetricsProvider interface to [provide metrics through OpenTelemetry](https://github.com/hyperledger/fabric-chaincode-java/pull/153/files#diff-d97166bc281ed3eec5bb83ccd164698c2ee3ae1bd5e8e4ab6a8ea1d86068e951R24).
-
-The same set of changes must be applied to the Go and Javascript shim implementations.
 
 ## Changes to client SDKs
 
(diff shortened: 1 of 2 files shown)


text/0000-opentelemetry.md
```diff
@@ -0,0 +1,153 @@
...
+with the OpenTelemetry project. It advocates for relying on OpenTelemetry to record and send metrics.
+It introduces the concept of tracing, reporting the execution of chaincode, core components and SDKs to help correlate activities with the chain.
+
+# Motivation
+[motivation]: #motivation
+
+The OpenTelemetry project is a CNCF project that aims to standardize observability, especially in a cloud-native environment.
+OpenTelemetry is working to create a standard around representing metrics, traces and logs.
+We aim to bring full observability of Hyperledger Fabric to allow tracing of transactions all the way from the client
+to each of the chaincode invocations. We also want to capture chaincode interactions.
```
yacovm: >  We also want to capture chaincode interactions.

Why is this information helpful?

> We want to be able to trace and correlate actions from chaincode events.

A chaincode event is simply an output of its execution. It is recorded on the blockchain. You can't trace it as it doesn't go to any node.

yacovm:
It seems to me the proposal is focused mainly on shims and SDKs. I'm not sure why tracing these two components is useful. An SDK is simply a software component that has a private key and can orchestrate transaction creation and notify the application about its success. A chaincode shim is nothing but a static process that receives as input a proposal, executes arbitrary logic and communicates with its peer (excluding the constellation where you deploy it as server which no one uses). 

If anything, the main thing that seems worthwhile to me to trace, is transaction identifiers. A transaction identifier flows from a client to peers and then back to the client to the ordering service, there it flows among nodes until it lands in a block and then it flows inside the peer in its stages of validation and commit, and tracing this end to end can be useful for analysis of bottlenecks and troubleshooting. 

I think that the proposal lacks examples of what you would want to trace, and also how this works with the fact that there is already a metric framework in Fabric that integrates with statsd and prometheus.

text/0000-opentelemetry.md
atoulme: This information is helpful as it helps understand the path of execution of code - the equivalent of logging entering methods and exiting methods, but with a simple, lightweight object that can track metadata and tags for the purpose of powering operations.

Chaincode events are typically triggers for oracles that then spawn code execution. Using the trace information associated with the event allows to correlate the execution of arbitrary code back to the event that triggered it and the transaction that originated the event.

atoulme:
The trace information is typically rich, with some trace providers arguing to replace logging and add as much metadata in the form of tags for traces. A transaction identifier would definitely be something to add to the trace. It would then be possible to search by transaction identifier for traces.

The existing metrics framework of Hyperledger Fabric is not discussed here indeed - OpenTelemetry can collect Prometheus metrics, and we have been successful in collecting those metrics with the OpenTelemetry Collector or Fabric Logger. I have been looking at offering OpenTelemetry as a framework for metrics for Fabric as well, but I believe that would be a separate enhancement. OpenTelemetry supports exporting metrics in Prometheus format or using the OTLP protocol, so it would be a shoe-in for the existing metrics framework.

In this enhancement, we propose that developers interested in shoring up metrics as part of their chaincode may do so, and allow for those metrics to be sent to a remote OTLP endpoint. This is a useful capability for chaincode developers.

I will work on providing more samples. Thank you!

text/0000-opentelemetry.md
```diff
@@ -0,0 +1,153 @@
...
+The SDK should use the standard environment variable environment to let users define how and if they want to report
+trace data to an endpoint of their choosing.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+OpenTelemetry has not reached 1.0. It has unequal maturation between metrics, traces and logs, and across languages.
+However, traces and metrics are well ahead of logs, and Java, Go and Javascript are well supported.
+
+The changes require a change in the structure of Protobuf messages, which may require a major upgrade of Fabric.
```
mastersingh24: I don't think we would need to modify the protobuf messages.  We should be able to use gRPC metadata for this.  Have a look at https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/tracing/opentracing .

atoulme: I missed that, I'll try it out and change the RFC.

atoulme: OK, I changed the Java reference implementation PR I provided along this RFC, and I'm able to test out and make sure this all works. I have pushed my changes to the PR and this RFC.

jt-nti:
It seems like this RFC is a good fit for some telemetry that was just optionally getting logged out in the Java chaincode implementation. Good to see that gRPC metadata works instead of making changes to the protobuf messages.

denyeart:
I agree with Yacov that tracing a transaction end-to-end seems like the most valuable use case for Fabric with OpenTelemetry.  I can see the need to instrument each of the SDKs for the end-to-end transaction view, but it seems it would be more efficient to instrument the peer and orderer for the transaction, rather than instrumenting each of the language-specific chaincode shims as is described in this RFC. Why implement in each chaincode language shim when it could be implemented once in the peer?

atoulme:
I have amended the RFC based on the conversation on the maintainers call this morning. In particular, I have removed the notion of chaincode tracing. I will create a separate RFC specifically for this particular aspect, as well as adding trace information to chaincode events.

atoulme:
As we continue the discussion and reviews of this RFC, may I request that a number be associated with this RFC and we may move it along the process?

text/0000-opentelemetry.md
```diff
@@ -0,0 +1,126 @@
...
+Observable software is exposing metrics, representing the internal state of the components, as well as logs, emitted from the execution of the software.
+
+In Hyperledger Fabric, we rely on the OpenTelemetry framework to report traces and correlate them across services.
+
+In practice, this means a request made by a client connected via the Fabric SDK to a node can pass on the context of a trace.
+Peers and orderers propagate the trace context and create spans indicating their own interaction.
+
+Blockchain operators can reconstitute a graph of the interaction of all the components at play to create a service map.
+This helps uncover trends and issues of performance, as well as shortening the time it takes to investigate problems.
+It also allows to quickly iterate when upgrading chaincode and witnessing the effects it has on the system.
```
sykesm: To echo comments from others, there seems to be too much focus on `chaincode` and not enough on the interactions between applications, peers, and orderers. In particular a chaincode invocation is initiated by a peer when an application sends a proposal to the peer.

text/0000-opentelemetry.md
```diff
@@ -0,0 +1,126 @@
...
+- Feature Name: OpenTelemetry Integration
+- Start Date: 2021-01-12
+- RFC PR: (leave this empty)
+- Fabric Component: core, chaincode, sdks
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This request for comments proposes integrating Hyperledger Fabric, its SDKs, core and chaincode components
```
sykesm: I thought the scope was to be reduced to remove the chaincode components?

sykesm (commented):
I'm late to the party but I agree with most of the comments that have been made by the other maintainers.

I also want to point out that the current implementation uses gRPC streams to communicate between the peer and chaincode. This pattern complicates things as messages from many different transactions will share the same stream context. I also want to note that Fabric does not consistently propagate contexts from the transport layer so there is some work to do beyond wiring in the interceptors.

text/0000-opentelemetry.md
atoulme: It is, but there were still mentions or allusions to chaincode tracing. I have removed all mentions of the word chaincode to avoid this trap :)

text/0000-opentelemetry.md
atoulme: I think I don't know that much about Fabric to be helpful there. There's probably some good documentation about the message exchanges in the Fabric doc, would you like me to link to it?

atoulme:
No, the streaming and context propagation might just be fine as the interceptor works at the level of a gRPC service and message exchange over the stream. Don't take my word for it - let's try it out!

sykesm:
> No, the streaming and context propagation might just be fine as the interceptor works at the level of a gRPC service and message exchange over the stream. Don't take my word for it - let's try it out!

I'm trying to point to an area that you may want to spend a bit of time looking at. Obviously the PRs to Fabric will demonstrate if and how things work.

atoulme:
No worries. I built an integration test that shows we can collect traces with the chaincode stream here:
https://github.com/hyperledger/fabric-chaincode-java/pull/153/files#diff-3d1a2998188d8694d38faff7ee0a9a713fe5373927db9fa5eef9962eff319b2eR127

Maybe it can be enhanced to push a bit further, but chaincode is out of scope anyway. I guess I'll get started on a RFC just for chaincode soon.

sykesm (approved):

atoulme:
This request was approved a while back, should it be merged? What is the next step here?

mastersingh24 (approved):

mastersingh24:
@denyeart  - Can you have a look at this as well?  I think we are almost ready for final review on this one

atoulme:
Would it be possible to receive an update on this RFC please?

denyeart:
@atoulme Sorry this got held up, I think we are close. The last thing that I think could be clarified in the RFC is some discussion of tracing versus metrics. This RFC mentions tracing the most, but also mentions metrics, so it is unclear if the focus is the former or both. Would it make sense as a first pass to focus on tracing only in this RFC? Or do tracing and metrics typically go hand-in-hand? Do you envision the metrics support complementing or replacing the existing Fabric metrics? Would you expect users to want to use both existing Fabric metrics and OpenTelemetry metrics or would that be overkill? Can the existing Prometheus service in the peer and orderer be leveraged for the OpenTelemetry metrics?

I see you replied to Yacov with some metrics thoughts, but I think those thoughts either need to be expanded within the RFC text itself (as well as addressing my questions), or removed from the RFC to make it clear that the initial focus is on tracing.

atoulme:
I am happy to slice and dice it however makes most sense, but OpenTelemetry provides support for metrics and traces. Therefore, if it is applied across the board as a standard, you should expect to see metric integration too.
The RFC sells heavily on the traces side because that's a net new feature for Fabric.

Since I introduced this RFC, OpenTelemetry also issued a standard for logs.

atoulme:
@denyeart please let me know what you think?

denyeart:
@atoulme I think we are very close... could you respond to each of my questions from May 11th and update the RFC accordingly? Thanks!

atoulme:
Understood, working on it. Thank you for your patience.

atoulme:
I have amended the scope of this RFC. @denyeart, please see my answers inline:
> @atoulme Sorry this got held up, I think we are close. The last thing that I think could be clarified in the RFC is some discussion of tracing versus metrics. This RFC mentions tracing the most, but also mentions metrics, so it is unclear if the focus is the former or both. 

Sorry about that. I dropped all reference to metrics.

> Would it make sense as a first pass to focus on tracing only in this RFC? Or do tracing and metrics typically go hand-in-hand?

You can just do tracing for now. That's plenty to get started.

> Do you envision the metrics support complementing or replacing the existing Fabric metrics? Would you expect users to want to use both existing Fabric metrics and OpenTelemetry metrics or would that be overkill? Can the existing Prometheus service in the peer and orderer be leveraged for the OpenTelemetry metrics?

The OpenTelemetry SDK is really great to use for metric collection. It also allows to push metrics through the current Prometheus set up with no changes - you just integrate OpenTelemetry with a Prometheus collector. You can combine sending metrics with OpenTelemetry gRPC push and Prometheus. The current Prometheus exports are fine and can be used for now. OpenTelemetry Collector can scrape Prometheus metrics. We don't need to do this at this time or in this RFC.

> I see you replied to Yacov with some metrics thoughts, but I think those thoughts either need to be expanded within the RFC text itself (as well as addressing my questions), or removed from the RFC to make it clear that the initial focus is on tracing.

I have removed metrics from the RFC so the scope is tighter.

denyeart (approved):
@atoulme thanks for addressing the comments. Given three approvals we can enter final comment period. I'll announce on the Fabric mailing list.

ryjones:
this has three approvals and it's been a week - should this be merged? @denyeart

denyeart:
@ryjones Yes, this one is ready to merge!

PR #44, opened 2021-03-23
Add RFC for Orderer ledger pruning/snapshots

dvitas:
The RFC introduces ledger snapshot feature at orderers, similar to the same feature at peers. Additionally the RFC describes the proposed process of pruning the ledger without restarting the orderer.

Here is the relevant mail thread:
https://lists.hyperledger.org/g/fabric/topic/80823191

### DIFF
--- a/text/0000-orderer-snapshot.md
+++ b/text/0000-orderer-snapshot.md
@@ -1,0 +1,316 @@
+---
+layout: default
+title: Orderer ledger snapshots
+nav_order: 3
+---
+
+- Feature Name: OSN_snapshots
+- Start Date: 2021-03-23
+- RFC PR:
+- Fabric Component: orderer, ledger
+- Fabric Issue:
+
+# Summary
+[summary]: #summary
+
+The RFC introduces ledger snapshot feature at orderers, similar to the same feature at peers.
+
+# Motivation
+[motivation]: #motivation
+
+Real world use cases show the necessity of controlling the disk space consumed by block ledger, and this RFC is mainly motivated by the need to reduce space consumption at orderer nodes.
+Additionally this RFC presents a better way to join a channel without pulling the huge number of all historical blocks, which also may take significant time and may pose a problem for scaling of an existing Fabric network.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC introduces a notion of a **ledger snapshot for OSNs** (ordering service nodes). With this notion two new logical operations appear at the orderer REST API and in the CLI tool *osnadmin*:
+- Create a snapshot,
+- Join channel from a snapshot.
+
+Ledger snapshot is a set of files containing all the necessary information for OSN to join a channel from an arbitrary block in the chain without loss of any functionality except the ability to deliver an earlier block.
+
+A **local orderer administrator** is a role that was introduced in RFC-0004 "Channel participation API without system channel" and is currently implemented with the help of "Admin Configuration" section in the OSN YAML config file (<Config.Admin.TLS>). Only the identity associated with this role is able to invoke the API defined in this RFC, that is existing **channel participation API** and newly defined **ledger management API**.
+
+## Guide: Create snapshot
+
+To create a ledger snapshot use the following REST API (optional block_number should be supplied as a parameter in the request body):
+
+    POST /ledger/v1/{channelID}/snapshot
+
+ or the following CLI command:
+
+    $ osnadmin ledger snapshot -c <channelID> [-b <block_number>]
+
+If the optional parameter <block_number> is omitted, snapshot will be created from the latest block, i.e. from the block number equal to ledgerHeight-1.
+
+A process of snapshot creation will start immediately if the block number is omitted, or if the block number is specified and is less than current ledger height. If the block number is specified and is in future (>= ledger height) then the operation will immediately fail with a reasonable error message.
+
+The process of snapshot creation is a synchronous one; upon the successful completion of the "ledger snapshot" API call the snapshot itself is ready and can be accessed in the following directory in the file system:
+```
+<Config.FileLedger.snapshots.rootDir>/completed/<channelID>/<block_number>
+```
+
+## Guide: Join channel from a snapshot
+
+To join a channel from a snapshot use the following REST API (files from the snapshot must be placed in the body, using multipart/form-data content-type):
+
+    POST /participation/v1/joinbysnapshot
+
+ or the following CLI command:
+
+    $ osnadmin channel joinbysnapshot -s <snapshot_dir>
+
+There is no need to specify {channelID} when using the above API, as it will be extracted from the snapshot content.
+
+To check the status of the "join by snapshot" operation use the following REST API:
+
+    GET /participation/v1/channels/{channelID}
+
+ or the following CLI command:
+
+    $ osnadmin channel list -c <channelID>
+
+## Guide: Prune block ledger of an existing channel
+[guide-prune-block-ledger-of-an-existing-channel]: #guide-prune-block-ledger-of-an-existing-channel
+
+To prune an existing channel ledger do the following:
+ - create a new snapshot at the desired block number (using "ledger snapshot" API),
+ - remove the channel from the OSN (using "channel remove" API),
+ - join the channel from the snapshot (using "channel joinbysnapshot" API).
+
+This approach has some drawbacks, see [Drawbacks section](#drawbacks).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Changes to the existing API
+
+This RFC adds a new endpoint to the existing **channel participation API** to allow joining a channel by snapshot. Additionally this RFC defines new **ledger management API** and introduces a new endpoint of this API to allow snapshot creation. In future we plan to add at least one more endpoint to this API which will enable ledger pruning "on the fly" without interruption in OSN service to resolve the known [drawbacks](#drawbacks).
+
+Technically we achieve the above by following these steps:
+1. Introduce new REST API endpoint **/participation/v1/joinbysnapshot** while keeping all existing functionality of the channel participation API, that is keeping "join by join-block", "remove channel", and "list channel(s)" API endpoints.
+2. Add new REST API handler for the ledger management API with prefix **/ledger/v1/**. Define a single endpoint of this API **/ledger/v1/{channelID}/snapshot** to allow snapshot creation.
+3. Add "Ledger Management API Configuration" section in the OSN YAML config file.
+4. Add support of two new snapshot operations to the CLI tool *osnadmin*, specifically the commands *osnadmin channel joinbysnapshot* and *osnadmin ledger snapshot*.
+
+## Snapshot content
+
+We propose the OSN snapshot to follow the same logical structure as peer snapshot, that is to comprise a directory with a number of files in it, one of the files being a metadata JSON file. But here the similarity ends, as the content of the files will be drastically different.
+Here is the content of the OSN snapshot:
+1. Snapshot version - to allow for the future extensions. Current version is 1.
+2. Channel ID.
+3. lastBlockNum - the number of the last pruned block.
+4. last config block before lastBlockNum - this is required to verify the newly pulled blocks. The sequence number of the last config block must be lower or equal to lastBlockNum.
+5. lastBlockHash - the hash of the last pruned block.
+6. prevBlockHash - technically not required, but this will make internal structure BlockchainInfo complete.
+7. ConsenterMetadata of the last pruned block (lastBlockNum)
+
+Version 1 of the OSN snapshot described in this RFC defines three files as the content of the snapshot directory:
+1. Last config block for the channel before lastBlockNum. Content: The block data with block metadata removed completely, in protobuf binary serialization format. Name: "lastConfig.block"
+2. Value of OrdererBlockMetadata.ConsenterMetadata of the last pruned block (note: this is coming not from the last config block, but rather from the last pruned block, that is from the block with sequence number lastBlockNum). Content: ConsenterMetadata value in protobuf binary serialization format. Name: "ConsenterMetadata.pb"
+3. Metadata file containing all the rest of the snapshot content. Name: "metadata.json"
+
+Version 1 of the OSN snapshot defines the following JSON fields in the "metadata.json" file:
+```json
+{
+  "version": 1,
+  "channelID": string,
+  "lastBlockNum": number,
+  "lastBlockHash": string,
+  "prevBlockHash": string | null, // null is valid only when lastBlockNum == 0
+  "parts": [ // array of all the files in the snapshot
+    "lastConfig.block",
+    "ConsenterMetadata.pb"
+  ]
+}
+```
+
+Unlike peer snapshot the OSN snapshot does not contain state DB, private data configs, the list of TxIDs and commit hash - as the OSN is purposefully designed to be unaware of these. So these two kinds of snapshots (peer snapshot and OSN snapshot) bear resemblance only on the conceptual level, being completely different in the content.
+
+## Future-compatibility of the "join by snapshot" API
+
+1. REST API future-compatibility.
+To maintain compatibility with the future modifications of the snapshot format, this RFC mandates the following:
+
+    1.1. The POST request to **/participation/v1/joinbysnapshot** endpoint must use multipart/form-data content-type.
+
+    1.2. First part of the multipart body must have the name "metadata.json" and must always contain "metadata.json" file.
+
+    1.3. API handler must read the "metadata.json" part first, confirm the validity of the JSON format and extract the fields defined for the version 1. Additional fields may be present in future snapshot versions; the presence of the additional fields not defined in version 1 format should not prevent the process from successful completion.
+
+    1.4. API handler checks the validity of all the metadata fields, ignoring the fields not defined in version 1. In case the validity check fails, the API handler must fail immediately with an appropriate error.
+
+    1.5. After the successful check of the metadata, the handler searches for two more parts with the names "lastConfig.block" and "ConsenterMetadata.pb" containing the respective files. Additional body parts may be present in future snapshot versions; the presence of the additional body parts not defined in version 1 format should not prevent the process from successful completion.
+
+    1.6. API handler checks the validity of the found "lastConfig.block" and "ConsenterMetadata.pb" files. In case the validity check fails, the API handler must fail immediately with an appropriate error. Otherwise it proceeds with the "join by snapshot" process.
+
+2. CLI tool *osnadmin* future-compatibility.
+When processing *osnadmin channel joinbysnapshot* command, the following checks must be performed:
+
+    2.1. Firstly, the CLI tool should find the file "metadata.json" in the specified directory <snapshot_dir>.
+
+    2.2. Next it must confirm the validity of the JSON format and the presence of all metadata fields defined in version 1. Additional fields may be present in future snapshot versions; the presence of the additional fields not defined in version 1 format should not prevent the process from successful completion.
+
+    2.3. Next, the CLI tool checks the validity of all the metadata fields, ignoring the fields not defined in version 1. In particular all the files defined in "parts" object must be present. In case the validity check fails, the process must fail immediately with an appropriate error.
+
+    2.4. Lastly, the CLI tool creates POST request to **/participation/v1/joinbysnapshot** endpoint and places all the files defined in the "parts" object of the metadata JSON into the body of the request according to the REST API specification. Name of each part must be the name of the file as defined in "parts" object of the metadata JSON. First part must be the "metadata.json" itself.
+
+## Implementation considerations
+
+When a channel is joined from a snapshot by "channel joinbysnapshot" command, we have to save the following info on the block storage:
+1. Last config block from snapshot
+2. BootstrappingSnapshotInfo
+
+For BootstrappingSnapshotInfo persistence the implementation is already there, but for the last config block a new persistent unit of storage (a file in case of file-based ledger) should be created.
+A special attention should be paid to the use case of **Deliver** service being asked to deliver a single block with the number equal to the block number of the last config block from a snapshot - in this case the delivery request must succeed by returning the last config block from the new persistent unit of storage (from a separate file) and not from the continuous block ledger. In general, the ledger storage interface should be aware of a possibility that aside from a continuous block ledger there is a special case of a single block representing a last config block from a snapshot which should also be successfully returned when requested by block number.
+Additionally all the procedures inside the OSN codebase looking up the last config block in the OSN ledger storage should be modified to return the last config block from the snapshot in case the last config block and/or a last block is not found in the block ledger itself.
+
+The channel bootstrapping from snapshot must be crash-consistent, that is if a crash takes place while bootstrapping a channel, upon restart the orderer, either the bootstrapped channel exists in the usable state or not at all. In particular this means that the creation of the above two pieces of information (last config block and BootstrappingSnapshotInfo) must be an atomic transaction, and there should be no possibility after a crash to have one of them but not another.
+Possible implementations of such an atomic transaction may be keeping both pieces of information in the same file and creating the file atomically, or keeping them in separate files in the same directory and creating the directory atomically.
+
+## System channel co-existence
+
+The following API can work both with system channel and without it:
+- "channel list" - existing API, keep existing behaviour
+
+The following APIs can work only without system channel, will fail in presence of system channel:
+- "channel joinbysnapshot"
+- "channel remove" - existing API, keep existing behaviour
+- "channel join" - existing API, keep existing behaviour
+- "ledger snapshot"
+
+## New channel creation ("genesis" snapshot)
+
+A new channel can be created using the existing API "channel join" with a genesis block as a join-block.
+
+Theoretically a "genesis" snapshot also can be constructed. For this a genesis block should be used as a last config block and lastBlockNum should be set to 0. This RFC does not define any specific tool capable of creation of a "genesis" snapshot, instead this RFC mandates that such a snapshot should be accepted as a valid one. In particular a snapshot should be considered as valid if prevBlockHash is omitted from the metadata JSON file or is set to "null", but only when lastBlockNum is equal to 0.
+
+## A **follower** mode
+
+Exactly as when using "channel join" API the orderer bootstrapping from a snapshot should start in the **follower** mode pulling the blocks without processing them until it finds a config block confirming its membership in the channel, unless it can be established that it is a member of the channel already (a member of the consenter set in case of RAFT).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There are three main drawbacks in the proposed ledger pruning method (see [Guide: Prune block ledger of an existing channel](#guide-prune-block-ledger-of-an-existing-channel)):
+- an interruption in the service of the orderer is required,
+- this works only in the absence of system channel, because "channel joinbysnapshot" works only in the absence of system channel.
+
+To resolve the above problems a new RFC will be issued introducing a separate **Prune** endpoint to the **ledger management API** which will allow to prune the ledger without stopping the service and will work both in presence and in absence of system channel.
+
+## Admin risks
+
+Allowing ledger pruning in OSNs introduces a new risk at the network level. It is possible to end up in a situation when all orderers are pruned at some point and the peer snapshot above that point is not accessible. In this situation it is impossible to bootstrap a new peer.
+We believe that the benefits of the ability to manage disk space outweigh the possible risk; nevertheless the admins should apply extreme caution when pruning OSNs.
+Possible solutions and the network admin level may be:
+ - always keep at least one orderer un-pruned;
+ - ensure the existence of a peer snapshot above the lowest pruning point of all orderers.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+## Alternatives considered
+
+- Implement only pruning without the snapshot functionality.
+  This still requires storing some information about the history of the ledger, as a bare minimum we need the last config block and the hash of the last pruned block. So why should we invent a new mechanism to support pruning, we can as well use the snapshot format to store the required info.
+- Store the information about the first known block in a new field (firstBlockNum) somewhere in the blockStore or fileLedger structure.
+  Again, we already have a code implemented that stores this info in BootstrappingSnapshotInfo structure. It is better to rely on the already tested code to achieve the same functionality.
+
+# Rationale
+
+Strictly speaking to achieve the main goal - orderer ledger pruning - we don't need the snapshot functionality. But due to the implementation reasons the notion of snapshot arises in some or another way. For example we need to store somewhere the last config block and the number of the last pruned block (or the number of the first block existing in the pruned ledger). So we decided that it is better to rely on the already existing code supporting snapshots and slightly modify it to suit the needs of the OSN snapshot.
+
+In addition to the above, joining a channel from a snapshot solves many problems with the existing join-block implementation from RFC-0004 "Channel participation API without system channel", namely:
+- the need to pull potentially huge number of the historical blocks
+- the fact that the config in the join-block may be outdated and not representing the latest config
+
+# Prior art
+[prior-art]: #prior-art
+
+RFC-0000 "Ledger Checkpointing"
+
+RFC-0004 "Channel participation API without system channel"
+
+# Testing
+[testing]: #testing
+
+Integration tests will cover the following scenarios:
+
+1. Snapshot generation
+  - start with a set of OSNs, with a channel, with blocks
+  - create a snapshot from the same non-genesis block on all the OSNs
+  - verify that the snapshot generated across OSNs is exactly same
+2. Bootstrap:
+  - start with a set of empty OSNs
+  - construct a "genesis" snapshot
+  - join all to the application channel from the "genesis" snapshot
+3. Add new channel:
+  - start with a set of OSNs with a channel
+  - construct a "genesis" snapshot for the second app channel
+  - join a subset of OSNs to the second channel from the "genesis" snapshot
+4. Add new OSN to the channel:
+  - start with a set of OSNs, with a channel, with blocks
+  - create a snapshot from some non-genesis block
+  - add the new OSN to the channel from the snapshot:
+    - as a follower (not in consensus set), adding it after catch-up
+    - as a member (in consensus set)
+5. Add new channel without genesis block:
+  - start with a set of OSNs, with a channel, with blocks, with a non-genesis config block among them
+  - create a snapshot from some block after the last config block
+  - remove the OSN from channel
+  - start a set of additional OSNs
+  - add all OSNs to the channel from the created snapshot
+6. Snapshot creation after bootstrapping:
+  - start with a set of OSNs, with a channel, with blocks, with a non-genesis config block among them
+  - create a snapshot from some block after the last config block
+  - start a new OSN, no channels
+  - new OSN joins the channel from the snapshot
+  - add some blocks to the channel without a config block
+  - trigger snapshot creation
+  - check the correctness of the created snapshot
+  - add a config block and some more blocks to the channel
+  - trigger snapshot creation
+  - check the correctness of the created snapshot
+7. System channel restrictions:
+  - start with a set of OSNs with the system channel
+  - make sure that the following operations fail:
+    - "channel joinbysnapshot"
+    - "ledger snapshot"
+  - create an app channel with blocks
+  - make sure that the following operation succeeds:
+    - "channel list"
+  - remove system channel
+  - make sure that two operations that previously failed now succeed
+8. Failure during bootstrapping the OSN from a snapshot
+  - start "channel joinbysnapshot" operation
+  - kill the OSN during the bootstrap
+  - restart the OSN
+  - verify that the OSN can again join the channel with the same snapshot
+  - verify that the OSN can again join the channel with a different snapshot
+
+# Dependencies
+[dependencies]: #dependencies
+
+**RFC-0004 "Channel participation API without system channel"**
+
+This RFC relies on the **channel participation API** introduced in RFC-0004 and extends it for the use case of joining a channel from a snapshot without breaking compatibility with the existing functionality of the channel participation API.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## Questions to be resolved during the implementation
+
+- Should we introduce a new return code to **Deliver** call when a block before the snapshot start is requested?
+
+It appears that the current codebase already returns a reasonable error code “NOT_FOUND“ with the log record “Error reading from channel, cause was: NOT_FOUND“ which indicates that the channel is there, but the block was not found. Because of this we do not include a new error code for this particular use case in the RFC.
+
+## Questions to be resolved in future RFCs
+
+- The problem of channel divergence that was pointed out in RFC-0004 "Channel participation API without system channel" is still unresolved.
+
+It is possible to join the channel with the same channelID by using the snapshots created from different config blocks with the same block number.
+
+- How to resolve a situation when all the orderers are pruned and a peer tries to fetch a block below the pruning point?
+
+The peer will receive an error code “NOT_FOUND“, but the problem is that it will not be able to proceed further and will be basically stuck at this point. This is something that has to be resolved on the network admin level and the exact resolution is unclear at this point.


text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,272 @@
...
+
+When a channel is joined from a snapshot either by "channel joinbysnapshot" or by "channel prune" command, we have to save the following info on the block storage:
+1. Last config block from snapshot
+2. BootstrappingSnapshotInfo
+
+For BootstrappingSnapshotInfo persistence the implementation is already there, but for the last config block a new persistent unit of storage (a file in case of file-based ledger) should be created. Additionally all the procedures looking up the last config block should be modified to return the last config block from the snapshot in case the last config block is not found in the block ledger itself.
+
+## System channel co-existence
+
+The following APIs can work both with system channel and without it:
```
yacovm: I don't think we should mix snapshots and system channel. It's best to simply disable the feature in case we are using a system channel. 

You can't "use" the snapshot with a system channel anyway, so there is no point in listing it or creating snapshots.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,272 @@
...
+
+To create a ledger snapshot use the following REST API (block_number must be supplied as a parameter in the request body):
+
+    POST /v1/mgmt/channels/{channelID}/snapshot
+
+ or the following CLI command:
+
+    $ osnadmin snapshot submitrequest -c <channelID> -b <block_number>
+
+A process of snapshot creation will start immediately if the block number is less than current ledger height. If the block number is in future (>= ledger height) then the snapshot creation will be scheduled and the snapshot will be created when the requested block will appear in the ledger. To check the status of the pending snapshot requests use the following REST API:
```
yacovm: I think it's best to simply fail fast in case the block height is higher. We have tools to track a block height, so this can be addressed with machinery instead of making the implementation and API even more complex.

text/0000-orderer-pruning-and-snapshot.md
dvitas: We included it here for two reasons:
1. It is technically possible, snapshot creation for a channel does not depend on the method of the channel creation (with or without system channel)
2. For the future expansion when snapshot will have more interesting content we could use snapshots from different orderers to verify that they are properly synchronized. This is one of the use cases for peer snapshots.

We understand that both arguments are weak, and if we need this in future we can easily allow snapshot creation in the presence of system channel anytime later.
So I will move all the "snapshot" APIs to the section "will fail in presence of system channel".

text/0000-orderer-pruning-and-snapshot.md
dvitas: This feature (scheduling snapshot in the future) was included only to be consistent with the peer snapshot implementation, where it is the only way of creating snapshots. So the code for snapshot scheduling is already out there.

If general consensus is that the ease of development is more important at this point, we would happily remove this "scheduling" feature, as we understand that the development and testing will be mostly on us.

I would like to collect a second opinion on this though.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+3. Rename "Channel participation API Configuration" section in the OSN YAML config file to "Channel management API Configuration", rename accordingly all the config settings
+4. Add three new snapshot operations to the REST API definitions of **/v1/mgmt/channels** REST API
+5. Add support of three new snapshot operations to the CLI tool *osnadmin*, specifically to the command *osnadmin channel* and the new command *osnadmin snapshot*
+6. Deprecate *osnadmin channel join* command. This will be the only command using deprecated **/v1/participation/channels**, all the rest of the commands including inherited *osnadmin channel remove* and *osnadmin channel list* must use internally new **/v1/mgmt/channels** endpoint.
+
+## Snapshot content
+
+We propose the OSN snapshot to follow the same logical structure as peer snapshot, that is to comprise a directory with a number of files in it, one of the files being a metadata JSON file with the checksums of other files and other metadata. But here the similarity ends, as the content of the files will be drastically different.
```
manish-sethi: > comprise a directory with a number of files in it

Just to clarify, won't it contain just two files.
1) The Most recent config block file
2) A JSON file that contains the metadata such as Version, ChannelD, LastBlockNumber, and LastBlockHash etc.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+Here is the content of the OSN snapshot:
+1. Snapshot version - to allow for the future extensions. Current version is 1.
+2. Channel ID.
+3. Last config block for the channel.
+4. lastBlockNum - the number of the last pruned block. This can be both behind the last config block and ahead of it.
+5. lastBlockHash - the hash of the last pruned block.
+6. prevBlockHash - technically not required, but this will make internal structure BlockchainInfo complete.
+
+Unlike peer snapshot the OSN snapshot completely ignores the state DB, private data configs, the list of TxIDs and commit hash - all of these are not used by orderers. So these two snapshots bear resemblance only on the conceptual level, being completely different in the content.
```
manish-sethi: > completely ignores the state DB, private data configs, the list of TxIDs and commit hash - all of these are not used by orderers

Perhaps a better wording may be -- does not contain state DB, private data configs, the list of TxIDs and commit hash - as the OSN is purposefully designed to be unaware of these.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Implementation considerations
+
+When a channel is joined from a snapshot either by "channel joinbysnapshot" or by "channel prune" command, we have to save the following info on the block storage:
+1. Last config block from snapshot
+2. BootstrappingSnapshotInfo
+
+For BootstrappingSnapshotInfo persistence the implementation is already there, but for the last config block a new persistent unit of storage (a file in case of file-based ledger) should be created. Additionally all the procedures looking up the last config block should be modified to return the last config block from the snapshot in case the last config block is not found in the block ledger itself.
+
```
manish-sethi: As this section talks about implementation consideration, just to take a note, I would suggest to add here a note about crash consistency. That is if a crash takes place while bootsrapping a channel, upon restart the orderer, either the boot strapped channel exists in the usable state or not at all. 

Though, in the case case of orderer, it should be much simpler as compared to the peer, still better to take a note of this. At the implementation for example, a simplest way could be to generate/copy the two files (`BootstrappingSnapshotInfo` and `last config block`) in a tmp folder (in the same filesystem) and then rename the folder to the proper name inside the blockstore.

text/0000-orderer-pruning-and-snapshot.md
manish-sethi: I agree with Yacov that we don't need this scheduling for a future block at the orderer. 

We introduced this mechanism in the peer because we opt for simplified design for the statedb and that limits us to take a snapshot of the data in the past and for a snapshot at a common height across peers, a future block was the only option.

In the case of an orderer, even if it is desired to take a snapshot at a common height, it can still be done for an historical height.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+Because of the drawbacks, as a part of this RFC we propose to implement a separate **Prune** API which allows to prune the ledger without stopping the service. Essentially this API will create a snapshot and will immediately start using it while deleting all the pruned blocks.
+
+To invoke this new **Prune** feature use the following REST API (block_number must be supplied as a parameter in the request body):
+
+    PATCH /v1/mgmt/channels/{channelID}/prune
+
+ or the following CLI command:
+
+    $ osnadmin channel prune -c <channelID> -b <block_number>
```
manish-sethi: In my opinion, since `prune` is an add-on operation, it can always be done in a separate RFC. However, If you want to include `prune` in the scope of this RFC, I would suggest to add details about this in the the next section as this requires additional work. In essence, this operation should leave an orderer in a state as if the channel on the orderer  had been bootstrapped from the snapshot. 

The additional work includes deleting the entries from the blockstore index for the blocks prior to the snapshot and  deleting the files (and possibly, splitting one of the files) that contains the prior blocks. Moreover, these operations need to be done in a manner that offers a crash consistent behavior.

On a side note, there are on-going thoughts on implementing the pruning feature for a peer as well. And that would take advantage of this partial work done in this context here.

About actual outcome of the pruning, I would suggest that we could just leave the unused block files in a separate location and let admin decide whether (s)he wants to delete these or archive them somewhere.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+[prior-art]: #prior-art
+
+RFC-0000 "Ledger Checkpointing"
+
+RFC-0004 "Channel participation API without system channel"
+
+# Testing
+[testing]: #testing
+
+Integration tests will cover the following scenarios:
```
manish-sethi: You may add one more test to the list -- an orderer can produce a snapshot after itself being bootstrapped from a snapshot irrespective of whether a config block has been added afterward.

text/0000-orderer-pruning-and-snapshot.md
```diff
@@ -0,0 +1,275 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC introduces a notion of a **ledger snapshot for OSNs** (ordering service nodes). With this notion three new logical operations appear at the orderer REST API and in the CLI tool *osnadmin*:
+- Create a snapshot,
+- Join channel from a snapshot,
+- Prune block ledger of an existing channel.
+
+Ledger snapshot is a set of files containing all the necessary information for OSN to join a channel from a block later than a genesis block (block #0) without loss of any functionality except the ability to deliver an earlier block.
+OSN ledger snapshot functionality is designed with the same principles as the peer ledger snapshot, in particular it is a full snapshot. We do not implement an incremental snapshot capability.
```
manish-sethi: > as the peer ledger snapshot, in particular it is a full snapshot. We do not implement an incremental snapshot capability.

I am not sure what do you mean by "incremental snapshot" in the context of an OSN. (In the case of a peer, it refers mainly to the contents of the statedb)

manish-sethi (commented):
Thanks @dvitas. Left a few comments.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done.
Snapshot scheduling in future will fail

text/0000-orderer-pruning-and-snapshot.md
dvitas: Text improved.
It now explicitly mentions that in version 1 of the snapshot we have only these two files.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done, text changed.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done.
Crash consistency consideration included in the text.

text/0000-orderer-pruning-and-snapshot.md
dvitas: > I am not sure what do you mean by "incremental snapshot" in the context of an OSN

I mean "any built-in mechanism that allows referring to an info from a previous snapshot instead of explicitly including it into the current snapshot"

Mostly this means that we do not include any mechanism of the possible dependency on a previous snapshot in the snapshot format.
I agree that there is no real reason for having incremental snapshots in OSNs, at least at the moment. So we just mention the absence of any mechanism built into the snapshot format for incremental build-up because we want to highlight the conceptual similarity between peer snapshots and OSN snapshots.

Anyway, I have improved the text to make it absolutely clear in this matter. Please check the latest commit.

Also it is very minor issue, so if you think it creates unnecessary confusion, I am happy to remove this paragraph altogether.

text/0000-orderer-pruning-and-snapshot.md
dvitas: It is in our best interest to streamline the approval and implementation of the RFC, so I will remove pruning from this RFC and create a separate one, for pruning only.
So this RFC will be limited in scope to snapshot feature only.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done, new test case has been added

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done.
Pruning feature removed from the RFC.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+
+    POST /v1/mgmt/channels/{channelID}/snapshot
+
+ or the following CLI command:
+
+    $ osnadmin snapshot submitrequest -c <channelID> -b <block_number>
+
+A process of snapshot creation will start immediately if the block number is less than current ledger height. If the block number is in future (>= ledger height) then the operation will immediately fail with a reasonable error message.
+
+The process of snapshot creation is an asynchronous one, so the success of the "snapshot submitrequest" operation does not indicate that the snapshot itself is ready. The indication of the snapshot readiness is the presence of the appropriate directory in the file system (<Config.FileLedger.snapshots.rootDir>/completed/<channelID>/<block_number>).
```
manish-sethi, line 46: - In my opinion, this can be made a synchronous call as the processing is not expected to take long. It makes the job of the caller a bit easier.
- <block_number> can be made optional (or block zero can be treated as a special value), which could be interpreted as the latest committed block.

text/0000-orderer-pruning-and-snapshot.md
manish-sethi: Thanks!. As we have agreement on this, following are additional points to consider...

- In my opinion, this API be made a synchronous call as the processing is not expected to take long. It makes the job of the caller easier
- <block_number> can be made optional (or block zero can be treated as a special value), which could be interpreted as the latest committed block
If you agree, the term `submitrequest` can be removed from the command.

text/0000-orderer-pruning-and-snapshot.md
manish-sethi: Yes, I feel that this adds an unnecessary diversion. So, better to omit it altogether in my opinion.

manish-sethi: Yes, I feel that it causes unnecessary diversion and can be omitted entirely.

text/0000-orderer-pruning-and-snapshot.md
manish-sethi: One more point that I missed earlier - you may like to add a line about the role that would be authorized to perform this command and add a test for that.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+ - join the channel from the snapshot.
+
+This approach has some drawbacks, see [Drawbacks section](#drawbacks).
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Changes to the existing API
+
+This RFC renames the existing **channel participation API** to **channel management API** and introduces three new logical operations (see [Guide-level explanation](#guide-level-explanation)) at this API while keeping the existing channel participation operations, namely:
```
jyellick: I'm not sure that it's appropriate to do this renaming.

The term "channel management API" was considered and debated during the development of the original feature, with the concern being that "channel management" conveys the idea that the API has an affect on the configuration of the channel, it does not.  This API only has an affect on how the orderer interacts with that channel.  Particularly, whether the orderer participates in the channel or not.

Extending this API to support not only the participation actions, but also the snapshotting/pruning actions might make this old name a bit misleading, but there's certainly a lot of breaking for existing applications and documentation by doing so.

Have you considered simply exposing another rest path specific to the new functions?

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+6. Deprecate *osnadmin channel join* command. This will be the only command using deprecated **/v1/participation/channels**, all the rest of the commands including inherited *osnadmin channel remove* and *osnadmin channel list* must use internally new **/v1/mgmt/channels** endpoint.
+
+## Snapshot content
+
+We propose the OSN snapshot to follow the same logical structure as peer snapshot, that is to comprise a directory with a number of files in it, one of the files being a metadata JSON file with the checksums of other files and other metadata. But here the similarity ends, as the content of the files will be drastically different.
+Here is the content of the OSN snapshot:
+1. Snapshot version - to allow for the future extensions. Current version is 1.
+2. Channel ID.
+3. Last config block for the channel.
+4. lastBlockNum - the number of the last pruned block. This can be both behind the last config block and ahead of it.
```
jyellick: A bit of a nit, but rather than "both .. behind .. and ahead" I believe the intent is "either ... behind ... or ahead".

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+
+Unlike peer snapshot the OSN snapshot does not contain state DB, private data configs, the list of TxIDs and commit hash - as the OSN is purposefully designed to be unaware of these. So these two kinds of snapshots (peer snapshot and OSN snapshot) bear resemblance only on the conceptual level, being completely different in the content.
+
+## Implementation considerations
+
+When a channel is joined from a snapshot by "channel joinbysnapshot" command, we have to save the following info on the block storage:
+1. Last config block from snapshot
+2. BootstrappingSnapshotInfo
+
+For BootstrappingSnapshotInfo persistence the implementation is already there, but for the last config block a new persistent unit of storage (a file in case of file-based ledger) should be created. Additionally all the procedures looking up the last config block should be modified to return the last config block from the snapshot in case the last config block is not found in the block ledger itself.
```
jyellick: > Additionally all the procedures looking up the last config block should be modified to return the last config block from the snapshot in case the last config block is not found in the block ledger itself.

I don't quite follow this.  Typically, to find the last config block from an orderer a client pulls the latest block, looks at its metadata to find a pointer to the latest config block, and then pulls that block.  Are you suggesting we augment the Deliver service with the ability to pull the original config block even if it is not in the ledger?

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+
+    $ osnadmin snapshot submitrequest -c <channelID> -b <block_number>
+
+A process of snapshot creation will start immediately if the block number is less than current ledger height. If the block number is in future (>= ledger height) then the operation will immediately fail with a reasonable error message.
+
+The process of snapshot creation is an asynchronous one, so the success of the "snapshot submitrequest" operation does not indicate that the snapshot itself is ready. The indication of the snapshot readiness is the presence of the appropriate directory in the file system (<Config.FileLedger.snapshots.rootDir>/completed/<channelID>/<block_number>).
+
+## Guide: Join channel from a snapshot
+
+To join a channel from a snapshot use the following REST API (files from the snapshot must be placed in the body, using multipart/form-data content-type):
```
jyellick, line 56: I don't know why we're coupling this API so tightly to the snapshot encoding format.

Why not simply add additional parameters to the existing join? Instead of only a config block, we could also specify a pruned block number and the other information we need to perform the replication.

If there is ever another snapshot format, then won't we run into problems?  If we have to version our APIs whenever the version of our snapshot changes, this seems like a bad situation.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,263 @@
...
+
+The following APIs can work only without system channel, will fail in presence of system channel:
+- "channel joinbysnapshot"
+- "channel remove" - this one is not described in this RFC, it is inherited from **channel participation** API
+- "channel join" - inherited from **channel participation** API, deprecated
+- "snapshot submitrequest"
+
+## New channel creation ("genesis" snapshot)
+
+To be able to create a new channel we need a valid snapshot created with the genesis block as a last config block and lastBlockNum set to 0. We propose to modify *configtxgen* tool so that it can produce this "genesis" snapshot. Then we can add new orderers to the channel from the genesis block using this "genesis" snapshot.
```
jyellick: I return to the question of why we are making new APIs around joining by snapshot.

If we simply accept 'zero' values as defaults, then joining by a genesis block would specify the genesis block, would implicitly set the prune height to 0 (the genesis block), and implicitly set the previous block hash to nil, as there is no previous block.  Everything seems to work without changing output artifact formats.

jyellick (commented):
In addition to the specific comments outlined below, I think it would probably be at least worth the thought experiment of what will happen to existing peers and applications when they encounter an orderer which does not have all of the historic blocks.

I'd point especially to the case where the last config block is below the prune point, and it will not be possible using any of the existing APIs (Deliver) to query _any_ config block from the orderer.  My feeling is that it will be hard to find all of these cases until things are implemented and tried in the field.  Still, I think it is worth putting some thought into how we might report these problems back to the user -- does the Deliver API provide clear guidance when the client requests a block below the lowest block stored in the blockstore?  Do we need a new return type or code for failures which indicate that the resources has been explicitly removed and will not return?  Perhaps those who were more involved with the peer snapshotting work can weigh in here with how these were addressed.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done, here:
27c1fbbadf41bbba9c119d99b6d27fff4ad7ec11

text/0000-orderer-pruning-and-snapshot.md
dvitas: Admin role explicit mention - done, here:
f6e34c0b072d3182a1adddc1e965d08b67310abe

As for the test - the test is to check that with the wrong identity we fail to create a snapshot. Currently Admin identity is implemented as TLS credentials, so the test is basically that with wrong TLS credentials the API call will fail - sounds like a unit test to me, not an integration test, hence we do not mention this test it explicitly in the RFC.

text/0000-orderer-pruning-and-snapshot.md
dvitas: Done, removed here:
b8f85b8ffd241e64a22b8e3dbc9be61aff9f8714

text/0000-orderer-snapshot.md
dvitas: Done, here:
507e634c0711d2668e93ae6aa5f8cc0ad2798973

text/0000-orderer-snapshot.md
dvitas: Done, here:
e93aada9af8665466cc0b5777d4ee62f19661e65

text/0000-orderer-snapshot.md
dvitas: The text in the RFC is about searching for the last config block locally, in the local ledger of an OSN. It is not about the remote block fetching.
Here are the couple of places (not exhaustive list!) where the search in local ledger should be modified to also account for the possibility that the last config block is not in the ledger, but in a bootstrapping storage unit (in a separate file):
https://github.com/hyperledger/fabric/blob/4cfc722d0ffef35a4ac8b0ebe17eea4dcc3b1018/orderer/common/follower/follower_chain.go#L552
https://github.com/hyperledger/fabric/blob/4cfc722d0ffef35a4ac8b0ebe17eea4dcc3b1018/orderer/common/cluster/util.go#L648

Also I have clarified the text in the RFC:
a916490fd6fb66e5c5903de8a062ee4cc3bb488a

text/0000-orderer-snapshot.md
dvitas: 1. We would like to avoid using the same API for both "join by join-block" and "join by snapshot", because these two operations are semantically quite different. "join by join-block" implies that all the blocks starting from block #0 will be pulled into the local ledger, while "join by snapshot" is specifically designed to avoid this. That's why we prefer to define two separate APIs for these two separate operations.
2. Forward compatibility concern - resolved here:
d17c0a01acca9efe8dd3539a2806972330fea785

text/0000-orderer-snapshot.md
dvitas: As we reverted the deprecation of "channel join" API, there is no need anymore in an additional tool support for the "genesis" snapshot. So now the RFC allows for the special case of "genesis" snapshot, but does not mandate any tool support for its creation.
Here:
0737f544deb4383c1d07a8f7368942869b3ba82a

dvitas:
> In addition to the specific comments outlined below, I think it would probably be at least worth the thought experiment of what will happen to existing peers and applications when they encounter an orderer which does not have all of the historic blocks.
> 
> I'd point especially to the case where the last config block is below the prune point, and it will not be possible using any of the existing APIs (Deliver) to query _any_ config block from the orderer. My feeling is that it will be hard to find all of these cases until things are implemented and tried in the field. Still, I think it is worth putting some thought into how we might report these problems back to the user -- does the Deliver API provide clear guidance when the client requests a block below the lowest block stored in the blockstore? Do we need a new return type or code for failures which indicate that the resources has been explicitly removed and will not return? Perhaps those who were more involved with the peer snapshotting work can weigh in here with how these were addressed.

1. There is definitely a risk of ending up in a situation when it is impossible for a peer to fetch a block because all orderers are pruned above that block. So to make this risk explicitly clear, we have described it in the "Drawbacks" section. Additionally we've added one more "Unresolved question". Here:
4edfc6fcf469c0dbd38e4263100c95c654ac49cd

2. Situation when a peer is unable to fetch the last config block from an orderer is impossible with the proposed solution.  An orderer will always have the last config block either in its block ledger, or in the bootstrapping snapshot.

3. How exactly to manage the case when a peer asks for a block below the pruning point - it is an open question. Current codebase returns "NOT_FOUND" to the peer, so at the peer side it is not exactly clear what exactly was not found - it may happen when a requested block is not found, or a channel is not found at the OSN. But if we consult OSN logs, there is a very distinct log message that appears only in case the requested block was not found: "Error reading from channel, cause was: NOT_FOUND"
So maybe the task of creating a separate response error code for this situation does not require an immediate solution as a part of this RFC and can be defined as a standalone JIRA ticket post-RFC implementation?
In any case, this is exactly the last unresolved question in the category "to be resolved during the RFC process".

manish-sethi:
Thanks @dvitas and this looks good to me now overall. Just on a second thought, I feel that it would be good to discuss one remaining point and better if we can resolve.

A peer snapshot for a channel on a given height that is generated on different peers is same. The intended use was that the admins from different orgs could match the snapshots and sign a common snapshot hash for future record (or if some new org wants to use this snapshot to start with). However, in the case of orderer, since it would contain the config block (which could differ at different OSNs, because of signatures field in metadata), this can pose a challenge. 

On the solution part, we can just retain the contents (i.e, the `Value` field) and remove the actual signature before dumping the config block and then include the hash of the serialized config block file into `parts` in the metadata. This will makes it similar to a peer snapshot and easy for admin from different orgs to compare and sign. When you load the config block during bootstrap, the bootstrapping OSN will populate this field with its own signature.

I would like to here from folks who are more knowledgeable in the OSN - @jyellick, @yacovm?

jyellick:
Rather than attempting to prune the block, I would suggest that instead you simply take the hash of the block header.  This is a hash over everything except the block metadata (which may differ from node to node and includes the signatures Manish referenced).  It is also the value used as "previous block hash" for the next block, so it is well known to be deterministic.

There is an open issue somewhere around block replication and validating that the signature set attached by the orderers in the metadata is valid, but that's an open question, outside of the scope of this RFC.

I believe I'm ready as written to approve and to handle the fallout during the implementation.

manish-sethi:
> Rather than attempting to prune the block, I would suggest that instead you simply take the hash of the block header. 

In the context of this RFC, my main concern was that in this approach, we would need a separate command line to compare the two snapshots. Otherwise, like peer snapshots, admins can simply use Linux command for matching the hashes of the files.

denyeart:
I tend to agree with Manish that the on-disk snapshots should be identical and easily comparable across orderer snapshots. Removing the metadata with signature when writing the last config block achieves that objective.  If it is important to retain the block with original orderer signature for audit purpose, that could be written as a second file for reference. But the last config block without orderer-specific metadata would be considered the actual snapshot that orderer administrators come to agreement on.  This way, each orderer admin can take their own snapshot and easily verify that they are consistent.

yacovm:
I agree with Manish's approach. I think the snapshot should contain the OrdererBlockMetadata as encoded in the signature's value field.

Also don't we need to persist the `*etcdraft.BlockMetadata` that is encoded in the consensus metadata? 
```

	// determine raft replica set mapping for each node to its id
	// for newly started chain we need to read and initialize raft
	// metadata by creating mapping between conseter and its id.
	// In case chain has been restarted we restore raft metadata
	// information from the recently committed block meta data
	// field.
	blockMetadata, err := ReadBlockMetadata(metadata, m)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to read Raft metadata")
	}
```

denyeart:
Ok, we are talking about two different things here.

1 - What to retain from the last config block. Manish and I are suggesting that the snapshot should just have the raw config block content, no metadata signature from a specific orderer, in order to keep snapshots orderer-agnostic and easily comparable. Is the config block's (unsigned) metadata important at all, if the actual snapshot is based on a later block?

2 - What to retain from the snapshot block so that Raft can bootstrap. This may require keeping the consensus metadata from the actual snapshot block (not the last config block). 

We'll need orderer experts to weigh in.

@jyellick thoughts?

manish-sethi:
@yacovm - I believe that you are correct that the metadata from the latest block would be needed, as it contains the corresponding raft entry index. However, that brings another question in my mind. What would happen if an OSN bootstrapped from a snapshot becomes a leader in a situation where some other follower is still behind the snapshot block? I guess that now this leader won't be able to serve previous entries. If this is correct, somewhere we need to capture that this situation should be avoided and once one snapshot is used to bootstrap an OSN, only same or some higher snapshot should be used to bootstrap any future OSN bootstrapping.

yacovm:
> What would happen if an OSN bootstrapped from a snapshot becomes a leader in a situation where some other follower is still behind the snapshot block?

Why is that a problem? Just don't prune an OSN until it passed the snapshot. The follower behind obviously cannot replicate blocks that no one has anymore and they are at best lying around in some tape storage picking dust. 

> once one snapshot is used to bootstrap an OSN, only same or some higher snapshot should be used to bootstrap any future OSN bootstrapping.

This is only true if you always prune blocks at every OSN after you checkpoint.

yacovm:
> @yacovm - I believe that you are correct that the metadata from the latest block would be needed, as it contains the corresponding raft entry index. However, that brings another question in my mind. What would happen if an OSN bootstrapped from a snapshot becomes a leader in a situation where some other follower is still behind the snapshot block? I guess that now this leader won't be able to serve previous entries. If this is correct, somewhere we need to capture that this situation should be avoided and once one snapshot is used to bootstrap an OSN, only same or some higher snapshot should be used to bootstrap any future OSN bootstrapping.

A Raft follower can synchronize from a Raft leader either by receiving AppendEntries directly (which are fetched from the memory storage) or via receiving a snapshot. 

When we receive a snapshot, we actually receive a single block, and we replicate to that block height. That replication, pulls blocks from all nodes and not just the leader.
So I am basically saying that if the leader is sufficiently far from the follower, it will send it a snapshot and won't be aware that it is missing blocks in its belly. 

The problem is in the direct append entries synchronization where the leader can be at block 105 where we pruned up to block 100 tries to synchronize a follower at block 95, and our storage size is 20 (by default), what will happen? 

Following the code, the leader will optimistically try to send a snapshot if it fails fetching the required missing entries from the storage (entries 96-->105). 

So, I think we are probably good.... but, we need to test that it really works in reality.

jyellick:
I'm a little perplexed by the emphasis on making snapshots the same across nodes.  If you have two different orderers generate a snapshot for the same block height, then yes, the hash of that snapshot will differ across nodes.  Just like if you fetch two blocks from two different orderers, the hash of those block files will differ.

In the peer case, the statedb does not have a Fabric v0.6 style state hash that's updated with every transaction, and therefore checkpoints are useful to compare across nodes, as they can confirm that the state has not forked.  For orderers, the current "state" is always reflected in the latest block header, so it's already trivial to detect forks.

So, I would ask, what is the true utility of deterministic snapshots at orderers?  It seems that it complicates our life without providing any real value.  Why are admins comparing these across disks?  Clearly it's not to detect forks.  Is it because one org does not trust another to provide a valid snapshot?  Ordering is not yet BFT, so a dishonest admin can always subvert ordering anyway, so what's the point?

I would actually suggest we address this in a different manner.  The real problem is that block files can have different hashes.  This is because the block metadata is not consented over.  But, this actually creates the possibility that blocks with invalid metadata could be disseminated (for instance, what if the admin forgets to authorize some OSN to create blocks?).  Similarly, under BFT, we will need to tolerate a malicious node forging bad metadata (for instance re-ordering node IDs), and will therefore need to have some attestation of the metadata's correctness.  I actually think solving this for BFT is an interesting problem.  In the CFT Raft case, it's as simple as having the leader send the block with a signature attached and having all involved parties use that block in its entirety.  For BFT, you might have different sets of signers, but you should always have the same metadata, it would require some thought and design (and very likely might need some more substantial changes to the block structure).

I guess my TLDR; version would be: Block metadata is a long outstanding problem, and if it makes our lives harder here, let's ignore it, tolerate the 'extra step' to compare snapshots, and come up with a real fix that will handle this use case and others as another RFC on the block metadata structure and construction.

yacovm:
> Similarly, under BFT, we will need to tolerate a malicious node forging bad metadata (for instance re-ordering node IDs), and will therefore need to have some attestation of the metadata's correctness. I actually think solving this for BFT is an interesting problem.

In [SmartBFT](https://smartbft-go.github.io/paper.pdf), the consensus metadata is carried both in the signature value inside the block metadata and also as a separate part (metadata field) of the [proposal](https://github.com/SmartBFT-Go/consensus/blob/master/smartbftprotos/messages.proto#L90-L94) ([pre-prepare](https://github.com/SmartBFT-Go/consensus/blob/master/smartbftprotos/messages.proto#L26-L30) inner part) consented upon (so it is duplicated). 
Then, is verified at the time of pre-prepare twice:
1. The consensus library grabs the consensus metadata from a dedicated field in the proposal that is being consented upon and [verifies it](https://github.com/SmartBFT-Go/consensus/blob/master/internal/bft/view.go#L504-L518).
2. The Fabric layer just [checks](https://github.com/SmartBFT-Go/fabric/blob/release-1.4-BFT-2/orderer/consensus/smartbft/verifier.go#L252-L264) that the metadata in the proposal field is equal to the encoded metadata in the block signature.

After the proposal is prepared, during the commit phase, it is [signed](https://github.com/SmartBFT-Go/fabric/blob/release-1.4-BFT-2/orderer/consensus/smartbft/signer.go#L46-L53) by a quorum of nodes. 

Therefore, having the snapshot include the block metadata will clearly allow you to onboard a SmartBFT ordering node even without signatures (as long as the leader is at least two blocks ahead from the snapshot, but, that's a quirk with an interesting story for another time...) as this block will never be need to be replicated anywhere (because everyone will have it) in pretty much any BFT you can think of, so I say that it's worth to include this field as part of the snapshot.

> I guess my TLDR; version would be: Block metadata is a long outstanding problem, and if it makes our lives harder here, let's ignore it, tolerate the 'extra step' to compare snapshots, and come up with a real fix that will handle this use case and others as another RFC on the block metadata structure and construction.

Can Raft block metadata differ across nodes? I would say that no, since all it contains is Consenter IDs and next consenter ID for that block height, and the index for the Raft entry that should be the same for each block.

If so, then I hope we agree that our consenter metadata (and last config index) are deterministic, right?

> So, I would ask, what is the true utility of deterministic snapshots at orderers? It seems that it complicates our life without providing any real value. 

The utility is that you could compare snapshots without any tooling, just by byte-wise comparison in applications, or check-summing the snapshots and then letting humans compare the digest strings. 
If the snapshot isn't deterministic, then you need to implement (and maintain) a software tool that compares snapshots.


>  Ordering is not yet BFT, so a dishonest admin can always subvert ordering anyway, so what's the point?

It is not yet BFT in the official Fabric, yes, but if we are designing a snapshot and pruning mechanism for Fabric, shouldn't we dedicate some thought to make it work regardless of the consensus protocol the orderer employs? So far, the orderer architecture allowed you to use integrate consensus mechanisms quite well - Solo, Kafka, Raft, and some unofficial BFT implementations, so I don't see a particular reason why we shouldn't devise an archive and pruning solution that is consensus agnostic.

denyeart:
@yacovm 

I still think we need to split the conversation into two parts. Let me see if I understand your opinion on both:

1 - last config block
Only the config itself is needed, therefore the block metadata can be entirely pruned from the config block at time of snapshot, making it easily comparable with standard file utilities.

2 - snapshot block
While the block itself is not needed (beyond the hash which is already planned to be included in the snapshot JSON), the block metadata is critical to retain, since it is needed to bootstrap an orderer in both Raft and BFT scenarios. The orderer-specific signature can be pruned from the block metadata, again making it easily comparable with standard file utilities.

Is that an accurate summary of your view?

denyeart:
Ok, thanks for confirming Yacov. I'm on board with that.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,309 @@
...
+4. Add support of two new snapshot operations to the CLI tool *osnadmin*, specifically the commands *osnadmin channel joinbysnapshot* and *osnadmin ledger snapshot*.
+
+## Snapshot content
+
+We propose the OSN snapshot to follow the same logical structure as peer snapshot, that is to comprise a directory with a number of files in it, one of the files being a metadata JSON file. But here the similarity ends, as the content of the files will be drastically different.
+Here is the content of the OSN snapshot:
+1. Snapshot version - to allow for the future extensions. Current version is 1.
+2. Channel ID.
+3. Last config block for the channel.
+4. lastBlockNum - the number of the last pruned block. This can be either behind the last config block, equal to the last config block number, or ahead of it.
```
denyeart: At first I assumed that the lastBlockNum must be higher than the last config block number, so that the orderer could process blocks using the latest config right after bootstrapping from a snapshot. But perhaps this is not needed, since the orderer could simply replicate blocks instead of process blocks, until it reached the specified config block in the future. Can anybody confirm?

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,309 @@
...
+5. lastBlockHash - the hash of the last pruned block.
+6. prevBlockHash - technically not required, but this will make internal structure BlockchainInfo complete.
+
+Version 1 of the OSN snapshot described in this RFC defines two files as the content of the snapshot directory:
+1. Last config block for the channel. Name: "lastConfig.block"
+2. Metadata file containing all the rest of the snapshot content. Name: "metadata.json"
```
denyeart: Per RFC discussion, we need to add the snapshot block metadata here as well, so that the orderer can bootstrap for Raft processing.

text/0000-orderer-snapshot.md
dvitas: As far as I understand it, this is exactly what the current channel participation code is doing ("channel join" API). It starts the orderer in a "follower" mode where it just pulls the blocks without processing them until it finds a config block confirming its membership in the channel.
I propose to use exactly the same logic (and the same code) in case of snapshot bootstrapping and start the orderer in a "follower" mode in the following cases:
- lastBlockNum is behind the lastConfigBlock, so the membership status of the orderer at the time of  this last pruned block is unclear (unless we can deduce the membership status from the metadata - can we?)
- lastBlockNum is equal or ahead of lastConfigBlock and the orderer is not a member of the channel (not in the consenter set).

dvitas: I've included the explicit description here:
f177408b823ef964cac27fd876db33cd8385ea9d

text/0000-orderer-snapshot.md
dvitas: Done, here:
eda4ecc28ccb19f10faa3c7671c0cf1b2025af9c

I included only consenter-related part of metadata in the snapshot, and not all the block metadata, for the following reasons:
1. To not include the signatures
2. We have "deprecated" items in the BlockMetadata array (BlockMetadataIndex_LAST_CONFIG, BlockMetadataIndex_ORDERER), they may go away or change their semantics, so the metadata bytes produced by a newer orderer will not be identical to the same of older orderers.
3. New items may be added to the BlockMetadata array which in turn will lead again to the differing binary representations between a newer and older orderers.

The same arguments apply to the decision to remove the whole BlockMetadata from the lastConfig.block (as opposed to removing only the signatures), to keep the binary representation of it as future-proof as possible.

text/0000-orderer-snapshot.md
manish-sethi: >lastBlockNum is behind the lastConfigBlock, so the membership status of the orderer at the time of this last pruned block is unclear (unless we can deduce the membership status from the metadata - can we?)

As far as I understand, in the metadata it would only be raft-node-IDs not the actual identities. However, in the "follower" mode to catch up in this scenario, it would need the channel config for verifying the signature. So, most recent channel config below the snapshot block in this case would also be needed. @yacovm and @jyellick certainly are more knowledgeable and would be good to hear from them.

jyellick: In order to pull a block, and properly validate it, you need to have the config block which preceded it.  That config block contains the MSP data (so that you can verify identities), and the BlockValidation policy which allows you take verified sets of identities from the block signatures (in the Raft case, just one), and verify that those identities were authorized to produce the block.

There are some definitely advantages to doing hash chain validation, instead of signature validation, and I believe the code does perform some optimization around this, but fundamentally, unless we modify the block replication to retrieve the entire chain backwards (which, could be done), if the lastBlockNum is sufficiently old, you will need the config before it to validate those blocks.  I would suggest that lastBlockNum must _always_ be greater than or (better yet) equal to the last config block.

dvitas: Done, here:
0a36f369aa3d54a84c99fe2cd7f54d4b0b0cfcb1

A comment:
Config blocks happen not that often, so if we limited lastBlockNum to be always a config block, we would loose the ability to prune the ledger at a desired point. Hence lastBlockNum is allowed to be greater than the last config block sequence number.

denyeart (approved):
Thanks for all the updates @dvitas , I think you can squash your commits now, as we won't need all the history in the final merge. See guidance [here](https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request).

I can go ahead and Approve, and then re-Approve when squashed.

We typically move an RFC to final comment period (one week period) once we get three maintainer approvals.

@manish-sethi @jyellick @yacovm Could you do a final review and add your approval if there are no more comments?

manish-sethi (approved):

dvitas:
> Thanks for all the updates @dvitas , I think you can squash your commits now, as we won't need all the history in the final merge. See guidance [here](https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request).
> 
> I can go ahead and Approve, and then re-Approve when squashed.
> 

All commits squashed, branch re-pushed as a single commit.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+ or the following CLI command:
+
+    $ osnadmin channel list -c <channelID>
+
+## Guide: Prune block ledger of an existing channel
+[guide-prune-block-ledger-of-an-existing-channel]: #guide-prune-block-ledger-of-an-existing-channel
+
+To prune an existing channel ledger do the following:
+ - create a new snapshot at the desired block number,
+ - remove the OSN from the channel,
```
yacovm: I don't understand why we're coupling here a channel scoped operation and a local operation.
If we want to prune an OSN then why not just prune it? 

Removing the OSN from the channel requires config update and joining it "from a snapshot" doesn't actually "bring it back" to the channel. 

I think we should include the prune step that is discussed in the drawback section in this RFC and not in the next one.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+4. Add support of two new snapshot operations to the CLI tool *osnadmin*, specifically the commands *osnadmin channel joinbysnapshot* and *osnadmin ledger snapshot*.
+
+## Snapshot content
+
+We propose the OSN snapshot to follow the same logical structure as peer snapshot, that is to comprise a directory with a number of files in it, one of the files being a metadata JSON file. But here the similarity ends, as the content of the files will be drastically different.
+Here is the content of the OSN snapshot:
+1. Snapshot version - to allow for the future extensions. Current version is 1.
+2. Channel ID.
+3. lastBlockNum - the number of the last pruned block.
+4. last config block before lastBlockNum - this is required to verify the newly pulled blocks. The sequence number of the last config block must be lower or equal to lastBlockNum.
```
yacovm, line 104: > last config block before lastBlockNum

I don't understand. If there is a config block before the "lastBlockNum" then that config block is essentially the last block, no? 

Also, "the number of the last pruned block" - Shouldn't we care about the number of the lowest block that is not pruned?

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+5. lastBlockHash - the hash of the last pruned block.
+6. prevBlockHash - technically not required, but this will make internal structure BlockchainInfo complete.
+7. ConsenterMetadata of the last pruned block (lastBlockNum)
```
yacovm, line 107: again, this should be the consenter metadata of the first block that is not pruned. That is the block that we bootstrap with, not the last block that was pruned.

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+
+Version 1 of the OSN snapshot described in this RFC defines three files as the content of the snapshot directory:
+1. Last config block for the channel before lastBlockNum. Content: The block data with block metadata removed completely, in protobuf binary serialization format. Name: "lastConfig.block"
+2. Value of OrdererBlockMetadata.ConsenterMetadata of the last pruned block (note: this is coming not from the last config block, but rather from the last pruned block, that is from the block with sequence number lastBlockNum). Content: ConsenterMetadata value in protobuf binary serialization format. Name: "ConsenterMetadata.pb"
```
yacovm, line 111: > Value of OrdererBlockMetadata.ConsenterMetadata of the last pruned block

Shouldn't this be the first non-pruned block?

text/0000-orderer-snapshot.md
```diff
@@ -0,0 +1,316 @@
...
+- "channel joinbysnapshot"
+- "channel remove" - existing API, keep existing behaviour
+- "channel join" - existing API, keep existing behaviour
+- "ledger snapshot"
+
+## New channel creation ("genesis" snapshot)
+
+A new channel can be created using the existing API "channel join" with a genesis block as a join-block.
+
+Theoretically a "genesis" snapshot also can be constructed. For this a genesis block should be used as a last config block and lastBlockNum should be set to 0. This RFC does not define any specific tool capable of creation of a "genesis" snapshot, instead this RFC mandates that such a snapshot should be accepted as a valid one. In particular a snapshot should be considered as valid if prevBlockHash is omitted from the metadata JSON file or is set to "null", but only when lastBlockNum is equal to 0.
```
yacovm, line 187: what's so special about this genesis snapshot? why do we need it?

text/0000-orderer-snapshot.md
dvitas: In a previous discussion with @manish-sethi it was decided that it is better to limit the scope of this RFC to snapshots only. On-the-fly prune operation is not a trivial one (mostly because of the concurrency issues), and snapshot functionality does not really depend on it.
https://github.com/hyperledger/fabric-rfcs/pull/44#discussion_r603650437

> Removing the OSN from the channel requires config update and joining it "from a snapshot" doesn't actually "bring it back" to the channel.

Snapshot functionality works only when the system channel does not exist, so if an admin is following these steps when system channel is present in the setup, the very first step (snapshot creation) will fail immediately.
(note: probably I should have mentioned this explicitly, that the described sequence of commands works only when the system channel is not present)

Also by removing OSN from channel I mean using the channel participation API "channel remove". This API does not produce new config block, it simply deletes the local ledger for the channel on the local filesystem of the OSN.
(note: again, probably I should have mentioned this explicitly to avoid confusion)

So if we use newly defined "channel joinbysnapshot" API in the last step, the OSN will pick up the last known config block which has this orderer listed as one of the consenters.

-------

In general this section of the *Guide-level explanation* is there only to demonstrate how this new snapshot feature can be used to prune the block ledger on the disk. This section does not mandate anything, it is just a usage example.

text/0000-orderer-snapshot.md
dvitas: > I don't understand. If there is a config block before the "lastBlockNum" then that config block is essentially the last block, no?

No. It may happen that the current block has sequence number 1,000,000 and the last config block has the sequence number 100.
We want to be able to prune the ledger at 1 million sequence number to save the disk space.

> Also, "the number of the last pruned block" - Shouldn't we care about the number of the lowest block that is not pruned?

"the number of the lowest block that is not pruned" =  lastBlockNum + 1

it does not matter what number to store

text/0000-orderer-snapshot.md
dvitas: > again, this should be the consenter metadata of the first block that is not pruned. That is the block that we bootstrap with, not the last block that was pruned.

ConsenterMetadata of the first block that is not pruned we will get from the block itself when it will be pulled - actually it will be the first block pulled after joining by snapshot. Also it may happen that we pruned the ledger exactly at the top, so "the first block that is not pruned" has not been created yet.

text/0000-orderer-snapshot.md
dvitas: The same as in the reply above.
We will get this info from the block itself when it will be pulled. Also this block even may not exist yet, if we created the snapshot at the very top of the chain.

text/0000-orderer-snapshot.md
dvitas: This is a nice feature to have. We do not strictly need it - as explained in the text, "channel join" API can be used to implement the same functionality.
But it makes the snapshot API self-sufficient and allows admins to operate with snapshots without worrying about an exceptional case of block number 0. And the implementation of it should be trivial.

text/0000-orderer-snapshot.md
yacovm: Oh I see I thought you wanted to do a config update that removes the OSN from the channel.
What you really meant was - _remove the channel from the OSN_, not _remove the OSN from the channel_. 

However let me ask you something - what if that node is a leader? What do we do then? It may destabilize the cluster. Perhaps we should try to [transfer the leadership](https://github.com/etcd-io/etcd/blob/main/raft/node.go#L180) in case that node is a leader for the channel?

yacovm (approved):

text/0000-orderer-snapshot.md
dvitas: > What you really meant was - remove the channel from the OSN, not remove the OSN from the channel.

Yes, you are right.
So I improved the wording here to avoid confusion and re-pushed.

> However let me ask you something - what if that node is a leader? What do we do then? It may destabilize the cluster. Perhaps we should try to transfer the leadership in case that node is a leader for the channel?

Unfortunately an admin does not have an API to transfer the leadership. So they will have to wait a bit for a new re-election to occur by timeout.

All in all, this is one more argument in favour of the future "pruning on-the-fly" RFC.

yacovm: I was trying to say that the node itself may transfer the leadership if it is the leader and only then proceed with deleting the channel from itself.

dvitas: Ah, I see.
Yes, I agree, it would be a nice improvement to the implementation of RemoveChannel() API.

I personally think that this is not in the scope of the RFC, but a separate JIRA ticket makes sense.

denyeart:
No additional comments during Final Comment Period, merging!

riki95:
Hi, working on a project for Siemens, we are in need of this feature as well. Glad to see that there is already an RFC. Let me know if you need help in some way and we can contribute.

PR #46, opened 2021-10-18
RFC for private data purge

manish-sethi:
This RFC proposes a new operation *purge* on private data via chaincode shim that allows purging of all the historical versions of the supplied keys from the private data store permanently so as to enable applications to implement right-to-forget behavior for private data.

Signed-off-by: manish <manish.sethi@gmail.com>

### DIFF
--- a/text/0000-private_data_purge.md
+++ b/text/0000-private_data_purge.md
@@ -1,0 +1,179 @@
+---
+layout: default
+title: Private Data Purge
+nav_order: 3
+---
+
+- Feature Name: private_data_purge)
+- Start Date: 2021-10-01)
+- RFC PR: (leave this empty)
+- Fabric Component: chaincode, ledger
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The private data collections feature in Fabric allows the dissemination of the sensitive data to a subset of organizations in a channel. A peer that is eligible to receive the private data of a collection within a namespace, maintains all the versions of the private data in its private data store forever. This RFC proposes a new operation *purge* on private data via chaincode shim that allows the deletion of one or more private data keys from the current state (statedb), and in addition, purging of all the historical versions of the supplied keys from the private data store permanently. In addition, it lays down a basic groundwork for enabling history-related queries on private data.
+
+# Motivation
+[motivation]: #motivation
+
+The primary motivation of this feature is to enable users to implement right to forget in the context of private data in a Fabric application. This may help users to meet certain business compliance about data privacy. At present, there exists two ways to retrieve historical versions of private data. An end user can retrieve historical versions of private data from a peer using the block Deliver service (`DeliverWithPrivateData` API). Additionally, other peers, potentially from other organizations, can access to the historical versions via the private data reconciliation process. Further, in future, we may expose an API for users to access the history of a specific private key in a collection, similar to the existing API for accessing the history of a key in the chaincode namespace. By enabling the purging of desired private data, the proposed feature makes peer prevent sending private data out from these APIs and eventually deletes it permanently from its persistent stores. An added advantage of this feature is that it enables users to free up some storage space by purging the historical versions of private data.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+To enable private data purge feature, we introduce a new API in chaincode shim. Below is the function signature of this new API.
+
+**`PurgePrivateData(collection, key string) error`**
+
+As far as the latest state of a channel, maintained in statedb, is concerned, this new operation, `PurgePrivateData`, is no different from existing operation, `DeletePrivateData`. Hence, an application or chaincode would not notice any difference in behavior. The additional semantics associated with `PurgePrivateData` primarily impact private data store and reconciliation of private data. Both of these are not exposed to end users as of now (beyond the APIs mentioned above).
+
+
+For understanding the usage of this API, lets assume a scenario where a chaincode defines a private data collection and the collection contains personally identifiable information for customers such that the chaincode maintains PII info of a customer in the form of three keys, namely {CustomerID}_PII_Name, {CustomerID}_PII_DOB, and {CustomerID}_PII_Address. While name and date-of-birth may not change, a customer may have updated his address a few times and hence peers of eligible orgs would have persisted all the historical values for the address. Now, the customer may want the Fabric network to purge any of its PII data.
+
+In order to achieve this, the application may invoke a chaincode transaction in which the chaincode in turn invokes the *PurgePrivateData* API three times - once for each of the above mentioned PII keys. Finally, the application would get enough endorsements to satisfy the endorsement policies for each of the above keys and submits the transaction to the orderers. Once the transaction is committed on a peer, the peer stops sending out any version of any of these private data keys and eventually deletes from the persistent stores.
+
+Note that we will gate this new API behind a capability tag so that this new operation is interpreted on all the peers in a channel and assures the user that the data will be purged if the transaction is committed as a valid transaction. In the absence of the capability, the endorser will return error during the invocation of `PurgePrivateData`. As this is a newly added function, it is sufficient to gate it at the endorsing time that ensures that no transaction with purge operation gets committed before enabling the capability.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Before diving into the proposed design, we describe some background, challenges, and broad requirements for the desired behavior for implementing private data purge.
+
+## Background, challenges, and Requirements
+We divide this section into two subsections.
+
+### Deletion of private data from statedb and private data store
+In regard of actual deletion of private data keys specified in a purge transaction, following is the desired behavior of a peer.
+- Deletion of the private data key from the state should be done atomically with the other state updates caused by the transaction. In other words, as far as state updates are concerned from the perspective of future transaction simulation, the *purge* operation is no different from *delete* and merely deleting from hashed state is sufficient.
+- In addition, in order to permanently deleting all traces of the private data, the key should eventually be deleted from the private state (maintained in statedb). However, it is not required to perform this action atomically with the commit of the transaction and can be performed in the background.
+- The deletion of all the historical versions from the private data store is expected to be time consuming. Moreover, it does not have any bearing on the transaction execution. So, in the interest of not directly impacting the transaction throughput, it is desired to perform this operation in the background
+- Despite performing the deletion from private data store in the background, the peer should behave for all external communication as if the data is actually purged, i.e., not send out the data in API responses after the purge transaction has been committed.
+- In some scenarios, when an org is removed from a collection, the peer of the org stops receiving future private data operations for the collection. This very well includes the private write-set that includes the purge operation itself. However, such peers may already have the historical versions of private data and latest of those versions in its private state. So, it poses a requirement for performing the deletion of private key versions both from the private data store and from the private key, while working with the hashed write-set, which is delivered to all the peers, as part of transaction inside the block.
+
+### Impact and desired changes on private data reconciliation
+Most impacted function from this feature would be [private data reconciliation](https://hyperledger-fabric.readthedocs.io/en/latest/private-data-arch.html#private-data-reconciliation). As a recap, the private data reconciliation enables an eligible peer to retrieve latest and historical versions of private data from other eligible peers. The reconciliation is triggered by the requesting peer when the peer has some versions of private data missing in its private data store. This missing data situation is caused by one of the following situations.
+
+1) Private data could not be delivered to the peer along with the block at the block commit time
+2) A peer (org) is added to an existing collection, so peer catches up by pulling all historical versions of private data
+
+In the current implementation of reconciliation, the basic unit of data transfer in the reconciliation is a tuple <blockNumber, transactionNumber, chaincodename, collection> i.e., the entire private write-set of a collection within a transaction is what the requesting peer asks for from another eligible peer. The requesting peer, before accepting the private write-set from the replying peer, verifies the contents by matching the hash of the entire private write-set with the expected hash for this collection that is included in the block. However, when we implement the private data purge feature, the replying peer may send a subset of the private write-set as some of the keys may have been purged. This change in behavior motivates the following changes in the reconciliation process.
+
+- The replying peer filters the already purged keys before sending out the response to reconciliation request
+  - For example, if a private key purge transaction is committed at block number 100, the replying peer will remove the private key from the response if the version of the key being sent was committed in a block prior to block number 100.
+  
+- The requesting peer filters the already purged keys
+  - In some scenarios, the replying peer of private data historical version may be behind the requesting peer in the terms of block chain height. For example, assume that a private key purge transaction is committed at block number 100. The requesting peer is at height 101 and seek to retrieve private data corresponding to block 50. The replying peer is at height 60, i.e., it can deliver the requested private data but has not yet seen the purge transaction. In this case, the requesting peer needs to trim the received write-set by removing the purged keys that it is aware of before persisting the write-set.
+
+- The requesting peer accepts and verifies subset of the collection
+  - Contrary to the above situation, the requesting peer may be behind and may not be aware of the purge transaction yet. While, the replying peer sends the subset of the collection write-set. In this scenario, the requesting peer needs to verify the subset by comparing the hashes of the individual keys (as opposed to the hash of the entire collection write-set). Moreover, the requesting peer cannot remove the collection from its list of still missing data, until the requesting peer itself sees the purge transaction, as the replying peer may be playing malicious by deliberately sending a selective subset.
+
+
+Keeping the above requirements and desired behavior in consideration, below is the the proposed design. First, we describe the new key types that we would maintain in the private data store and then, we discuss various operations in Fabric that would create, update, read, or delete these keys to implement this feature.
+
+## Proposed Design
+We divide the proposed design in four subsections. First, we describe the changes in the Write-set Proto message. Second, we introduce two new key types that we would maintain in the private data store. Third, we describe the proposed changes in the existing functions and finally, we describe the new functions that we would add.
+
+### Changes in Write-set Proto
+We capture the Hashed write-set for a particular key in the proto message [`KVWriteHash`](https://github.com/hyperledger/fabric-protos/blob/9c69228417592158899cb0dd4a77a3eedf25225f/ledger/rwset/kvrwset/kv_rwset.proto#L57). In the proposed design, We would extend this proto to capture the purge operation. We do this by adding an additional field `is_purge`. 
+- Note that unlike existing fields in the hashed-write-set, this field will not have any corresponding field in the raw private-write-set.
+- Also, when we set `is_purge` to true, as a result of invocation of the newly introduced chaincode shim API, we also set the existing field `is_delete` set to true both in the hashed-write-set as well as in the private-write-set. In the terms of semantics, we interpret `is_delete` as the instruction for deleting the key from the latest state while `is_purge` indicates purging from the private data store.
+
+### New Keys in Private data store
+ In this design, we maintain two types of additional keys in the private data store. We refer to these new types of keys as *PurgeMarker* and *HashedIndex*. Below, is the format and intended use of these keys.
+1) PurgeMarker
+   A PurgeMarker key is intended to record a purge operation for a key. The design of this marker key would be `Ns_Coll_Hash(PrivateKey)_CommittingBlkNum_CommittingTxNum`. For illustration, assume that a purge transaction appears in block number 100, at position 20 that specifies for purging the keys `key1` and `key2` under collection `myColl` for chaincode `myCC`. When processing this transaction, we insert the two keys into private data store as `myCC_myColl_hash(key1)_100_20` and `myCC_myColl_hash(Key2)_100_20`. Note that the separation of different components of a key shown by an underscore is just for the illustration. For the reasons discussed latter, we intend to keep these marker keys forever. However, the assumption is that the purge operation is expected to be relatively less frequent, and hence, it should not be a concern from storage space point of view.
+
+2) HashedIndex
+   A HashedIndex key is intended to act as an index for the data keys in private data store, so that when a key is eventually purged we know from where to purge it in the private data store. The format of this key would be `Ns_Coll_Hash(PrivateKey)_blkNum_txNum`. For each version of a private key in the private data store, we would maintain such an index key. For illustration, assume that a key `key1` under collection `myColl` in chaincode `myCC` is inserted at block 100, transaction 10. Later this key is updated in block 200, transaction 20 and finally, the key is deleted in block 300, transaction 30. In this case, today, we maintain in the private data store all three private write-sets that contain, potentially along with other, the above three mentioned operations. In this example, for `key1`, we would maintain three HashedIndex keys as `myCC_myColl_hash(key1)_100_10`, `myCC_myColl_hash(key1)_200_20`, and `myCC_myColl_hash(key1)_300_30`. The value for the HashedIndex key will be the private key itself. For instance, in the above two examples, the values would be `key1` and `key2` respectively.
+   
+   - Note that the format of this key look same as the format of the PurgeMarker key - however, we always prepend a unique byte for the key type so, no two keys of different type ever clash.
+   
+   - Unlike the PurgeMarker key, we delete an HashedIndex key when we delete the corresponding data, either because of newly introduced explicit purge operation in this RFC or because of an implicit purge operation caused by [expiry of private data](https://hyperledger-fabric.readthedocs.io/en/release-2.2/private_data_tutorial.html#purge-private-data).
+   
+   - An important point to note here is that we would need to create these index keys retroactively - i.e., when a peer is upgraded to a version that would introduce private data purge feature, as a part of the upgrade, we would create these index keys.
+
+### Changes in existing operations
+
+#### Block commit
+In the current code, when computing the state updates for a block, for each updated key, we take the latest value (i.e., the value set by the highest valid transaction in the block for the key). With the purge private data feature, while we still compute the state updates as is but, in addition, we perform following two data-set for committing to the private data store.
+- Compute a set of PurgeMarker keys such that the set contains one entry for each hashed-key that is marked for purge by at least one transaction in the block. Further, if there exist more than one transactions that purge the same key, we ignore the lower transaction for the key. For example, assume block 100 contains transaction 10 that purges the key `key1` and `key2` and transaction 15 that purges the key `key1` and `key3` for collection `myColl` in chaincode `myCC`. In this case the set of PurgeMarker keys would look like `{myCC_myColl_hash(Key1)_100_15, myCC_myColl_hash(Key2)_100_10, myCC_myColl_hash(Key3)_100_15}`.
+- Compute HashedIndex keys for each private key-value present in the private write-set.
+
+
+#### Private data commit via reconciliation
+As mentioned previously, private data reconciliation is going to be most affected of the existing functions. Below is a list of changes that we would make in the reconciliation.
+
+1) When we commit private data via reconciliation, the requesting peer may receive the keys that have already been purged. This would be the case when the replying peer has not yet seen the purge transaction. However, the requesting peer may have seen the purge transaction. In order to not save the purged keys, we would first compute the expected private write-set by removing the already purged keys on the requesting peer from the hashed write-set from block, by intersecting with PurgeMarker keys. This implicitly means that we need to maintain PurgeMarker keys forever (theoretically, at least, until no more missing data, including the missing data for ineligible collections, exists below the committing block of PurgeMarker key).
+
+2) In the current code, during reconciliation, a requesting peer verifies the data by computing the hash of the private write-set received against the hash of the collection present in the corresponding transaction in the block. On a side note, if a peer has been bootstrapped via a snapshot, this is true only for the private data that has been added after the height of the bootstrapping snapshot. Now, in the presence of the private data purge feature at a key level, a peer may receive a partial private write-set, because one or more keys in the private write-set may have been purged by the replying peer. Hence, it is required for the requesting peer to perform the verification of hashes at the level of individual key-value pairs.
+
+3) During reconciliation, a requesting peer may receive a subset of the expected write-set, as computed in (1) above, including an empty write-set, because the requesting peer may yet not have seen the purge transaction while the replying peer has seen it. To handle this, the requesting peer accepts the subset of the expected write-set and puts the missing data request into the deprioritized list and is attempted again at a low frequency interval. Eventually, the requesting peer catches up to block height and executes the purge transaction and gets aware of the purge transaction. Finally, when the expected keys, as computed in (1) above, becomes the empty-set, the peer updates the missing keys info as fully reconciled.
+
+#### Queries on Private Data Store
+The queries need to filter the private write-sets that it sends out as a response to a query. This is intended to be achieved with the help of PurgeMarker key, if the version of the key is lower than the version present in one of the corresponding PurgeMarkerKey keys. This is required because the actual data purge may take some time after having created the marker keys.
+
+
+### New Operations
+
+#### Purging from Private Data Store and private state
+As mentioned in section [Background, challenges, and Requirements](#background-challenges-and-requirements), we intend to perform the actual deletions of the keys in the background. Therefore, at the time of processing a private data purge transaction, we only insert the marker key at the time of block commit, as described in the section [Block commit](#block-commit). In order to perform the actual purge of the data, we run a background goroutine that reads the PurgeMarker keys and scans all the relevant HashedIndex keys such that the index keys contain lower blkNum_TxNum than what is present in the PurgeMarker key. Finally, we would use the index key to locate the actual private write-set and we would trim the write-set by purging the intended key. While we purge the historical versions from the private data store, we would also delete the private key from the private state maintained in the statedb. It would be possible to derive the actual private key from the key-hash with the use of HashedIndex key. In the implementation, we intend to add an info field in the Marker key that would indicate whether the particular marker key has been processed. Also, if another purge operation is specified on the same key in the future, we delete the existing PurgeMarker for the key
+
+#### Creating the HashedIndex keys retroactively
+As mentioned in the section [New Keys in Private data store](#new-keys-in-private-data-store), when starting the peer version that contains purge support for the first time, we would need to retroactively create the HashedIndex keys for already committed private data in the private data store. We intend to do this automatically at the start of peer using the framework we introduced in data format upgrade in version 2.0
+
+[drawbacks]: #drawbacks
+# Drawbacks
+If historical versions of private data are purged, it will not be possible to rollback the peers to a previous height without loosing the purged private keys forever. However, this is not a big drawback as reset/rollback is not a strategic direction for Fabric.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- Feature Name: private_data_purge)
+- Start Date: 2021-10-01)
+- RFC PR: (leave this empty)
+- Fabric Component: chaincode, ledger
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Private data collections feature in Fabric allows the dissemination of the sensitive data limitimg to a subset of organzations in a channel. A peer that is eligible to receive the private data of a collection within a namespace, maintains all the versions of the private data in priavte data store forever. This RFC proposes a new operation *purge* on private data via chaincode shim that allows the deletion of one or more private data keys from the current state (statedb), and in addition, purging of all the historical versions of the supplied keys from the private data store permanently. In addition, it lays down a basic ground work for enabling history related queries on private data.
```
denyeart:
```suggestion
Private data collections feature in Fabric allows the dissemination of the sensitive data to a subset of organizations in a channel. A peer that is eligible to receive the private data of a collection within a namespace, maintains all the versions of the private data in its private data store forever. This RFC proposes a new operation *purge* on private data via chaincode shim that allows the deletion of one or more private data keys from the current state (statedb), and in addition, purging of all the historical versions of the supplied keys from the private data store permanently. In addition, it lays down a basic ground work for enabling history related queries on private data.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The primary motivation of this feature is to enable the users to implement right to forget in the context of private data in a Fabric application. This may help users to meet certain business compliance about data privacy. At present, there exists two ways to retrieve historical versions of private data. An end user can retrieve historical versions of private data from a peer using Deliver service APIs and other peers, potentially from other organizations, can access to these historical versions via reconcilation process. Further, in future, we may expose an API for users to access the history of a specific private key in a collection, similar to the existing API for accessing the history of a key in the chaincode namespace. By enabling the purging of desired private data, the proposed feature makes peer prevent sending private data out from these APIs and evetually deletes it permanently from the persistent stores. An added advantage of this feature is that it enables users to free up some storage space by purging the historical versions of private data.
```
denyeart:
```suggestion
The primary motivation of this feature is to enable users to implement right to forget in the context of private data in a Fabric application. This may help users to meet certain business compliance about data privacy. At present, there exists two ways to retrieve historical versions of private data. An end user can retrieve historical versions of private data from a peer using the block Deliver service (DeliverWithPrivateData API). Additionally, other peers, potentially from other organizations, can access to the historical versions via the private data reconcilation process. Further, in future, we may expose an API for users to access the history of a specific private key in a collection, similar to the existing API for accessing the history of a key in the chaincode namespace. By enabling the purging of desired private data, the proposed feature makes peer prevent sending private data out from these APIs and eventually deletes it permanently from its persistent stores. An added advantage of this feature is that it enables users to free up some storage space by purging the historical versions of private data.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+To enable private data purge feature, we introduce a new API in chaincode shim. Below is the function signature of this new API.
+
+**`PurgePrivateData(collection, key string) error`**
+
+As far as the latest state of a channel, maintained in statedb, is concerned, this new operation, `PurgePrivateData`, is no different from existing operation, `DeletePrivateData`. Hence, an application or chaincode would not notice any difference in behavior. The additional semantics associated with `PurgePrivateData` primarily impact private data store and reconcilation of private data. Both of these are not exposed to end users directly as of now.
```
denyeart:
```suggestion
As far as the latest state of a channel, maintained in statedb, is concerned, this new operation, `PurgePrivateData`, is no different from existing operation, `DeletePrivateData`. Hence, an application or chaincode would not notice any difference in behavior. The additional semantics associated with `PurgePrivateData` primarily impact private data store and reconcilation of private data. Both of these are not exposed to end users as of now (beyond the APIs mentioned above).
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order to achive this, the application may invoke a chaincode transaction in which the chaincode in turn invokes the *PurgePrivateData* API three times - once for each of the above mentioned PII keys. Finally, the application would get enough endorsements to satisfy the endorsement policies for each of the above keys and submits the transaction to the orderers. Once the transaction is committed on a peer, the peer stops sending out any version of any of these private data keys and eventually deletes from the persistent stores.
+
+Note that we will gate this new API behind a capability tag so that this new operation is interpreted on all the peers in a channel and assures the user that the data will be purged if the transaction is committed as a valid transaction. In the absence of the capability, the effect of the API will be exactly same as invoking the API `DeletePrivateData`.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Before diving into the proposed design, we describe some background, challenges, and braod requirements for the desired behavior for implementing private data purge.
```
denyeart:
```suggestion
Before diving into the proposed design, we describe some background, challenges, and broad requirements for the desired behavior for implementing private data purge.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Background, challenges, and Requirements
+We divide this section into two subsections.
+
+### Deletion of private data from statedb and private data store
+In regard of actual deletion of private data keys specified in a purge transaction, following is the desired behaviour of a peer.
+- Deletion of the private data key from the state should be done atomically with the other state updates caused by the transaction. In other words, as far as state updates are concerned from the prespective of future transaction simulation, the *purge* operation is no different from *delete* and merely deleting from hashed state is sufficient.
+- In addition, for permanently deleting all traces of the private data, the key should eventually be deleted from the private state. However, this not a require to perform atomically with the commit of the transaction and can be performed in the background.
```
denyeart:
```suggestion
- In addition, for permanently deleting all history of the private data, the key should eventually be deleted from the private data store. However, this not a require to perform atomically with the commit of the transaction and can be performed in the background.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- The deletion of all the historical versions from the private data store is expected to be time consuming. Moreover, it does not have any bearing on the transaction execution. So, in the interest of not directly impacting the transaction throughput, it is desired to perform this operation in the background
+- Despite performing the deletion from private data store in the background, the peer should behave for all external comminacation as if the data is actually purged, i.e., not send out the data after the purge transaction has been committed.
```
denyeart:
```suggestion
- Despite performing the deletion from private data store in the background, the peer should behave for all external communication as if the data is actually purged, i.e., not send out the data in API responses after the purge transaction has been committed.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- In some scenarios, when an org is removed from a collection, the peer of the org stops receiving furture priavte data operation for the collection. This very well includes the private write-set that includes the purge operation itself. However, such peers may already have the historical versions of private data and latest of those versions in its private state. So, it poses a requirement for performing the deletion of private key versions both from the private data store and from the private key, while working with the hashed write-set, which is delived to all the peers, as part of transaction inside the block.
```
denyeart: I'm not following the main point of this bullet.

```suggestion
- In some scenarios, when an org is removed from a collection, the peer of the org stops receiving future private data operations for the collection. This very well includes the private write-set that includes the purge operation itself. However, such peers may already have the historical versions of private data and latest of those versions in its private state. So, it poses a requirement for performing the deletion of private key versions both from the private data store and from the private key, while working with the hashed write-set, which is delivered to all the peers, as part of transaction inside the block.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+1) Private data could not be delivered to the peer along with the block at the block commit time
+2) A peer (org) is added to an existing collection, so peer catches up by pulling all historical versions of private data
+
+In the current implementation of reconcilation, the basic unit of data transfer in the reconcilation is a tuple <blockNumber, transactionNumber, chaincodename, collection> i.e., the entire private write-set of a collection with in a transaction is what the requesting peer asks for from another eligible peer. The recieving peer, before accepting the private write-set, verifies the contents by matching the hash of the entire private write-set with the expected hash for this collection that is included in the block. However, when we implement the private data purge feature, the sending peer may send a subset of the private write-set as some of the keys may have been purged. This set forth broadly following requirements for the changes in the reconcilation.
```
denyeart: "Sending peer" and "receiving peer" names are ambiguous (are you talking about sending/receiving the request for private data, or sending/receiving the private data itself)? I suggest using terms "requesting peer" and "replying peer" throughout, as this makes it clear which peer you are talking about.


```suggestion
In the current implementation of reconcilation, the basic unit of data transfer in the reconcilation is a tuple <blockNumber, transactionNumber, chaincodename, collection> i.e., the entire private write-set of a collection within a transaction is what the requesting peer asks for from another eligible peer. The requesting peer, before accepting the private write-set from the replying peer, verifies the contents by matching the hash of the entire private write-set with the expected hash for this collection that is included in the block. However, when we implement the private data purge feature, the replying peer may send a subset of the private write-set as some of the keys may have been purged. This change in behavior motivates the following changes in the reconcilation process.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+- The sending peer filters the already purged keys before sending out the response to reconcilation request
```
denyeart:
```suggestion
- The replying peer filters the already purged keys before sending out the response to reconciliation request
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+  - For example, if a private key purge transaction is committed at block number 100, the sending peer will remove the private key from the response if the version of the key being sent was committed in a block prior to block number 100.
```
denyeart:
```suggestion
  - For example, if a private key purge transaction is committed at block number 100, the replying peer will remove the private key from the response if the version of the key being sent was committed in a block prior to block number 100.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+  
+- The receiving peer filters the already purged keys
```
denyeart:
```suggestion
- The requesting peer filters the already purged keys
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+  - In some scenarios, the sending peer of private data historical version may be behind the receiving peer in the terms of block chain height. For example, assume that a private key purge transaction is committed at block number 100. The requesting peer is at height 101 and seek to retrieve private data corresponding to block 50. The peer that is selected for delivering the requested private data is at height 60, i.e., it can deliver the requested private data but has not yet seem the purge transaction. In this case, the receving peer needs to trim the write-set by removing the purged keys that it is aware of before persisting the write-set.
```
denyeart:
```suggestion
  - In some scenarios, the replying peer of private data historical version may be behind the requesting peer in the terms of block chain height. For example, assume that a private key purge transaction is committed at block number 100. The requesting peer is at height 101 and seek to retrieve private data corresponding to block 50. The replying peer is at height 60, i.e., it can deliver the requested private data but has not yet seen the purge transaction. In this case, the requesting peer needs to trim the received write-set by removing the purged keys that it is aware of before persisting the write-set.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+- The receiving peer accepts and verifies subset of the collection
```
denyeart:
```suggestion
- The requesting peer accepts and verifies subset of the collection
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+  - Contarary to the above situation, the receiving peer may be behind and may not be aware of the purge transaction yet. While, the sending peer sends the subset of the collection write-set. In this scenario, the receiving peer needs to verify the subset by comparing the hashes of the indivial keys (as opposed to the hash of the entire collection write-set). Moreover, the peer cannot remove the collection from its list of still missing data, until the receiving peer itself sees the purge transaciton, as the sending peer may be playing malicious by delibrately sending a selective subset.
```
denyeart:
```suggestion
  - Contarary to the above situation, the requesting peer may be behind and may not be aware of the purge transaction yet. While, the replying peer sends the subset of the collection write-set. In this scenario, the requesting peer needs to verify the subset by comparing the hashes of the individual keys (as opposed to the hash of the entire collection write-set). Moreover, the requesting peer cannot remove the collection from its list of still missing data, until the requesting peer itself sees the purge transaction, as the replying peer may be playing malicious by deliberately sending a selective subset.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+2) HashedIndex
+   A HashedIndex key is intended to act as an index for the data keys in private data store that would be purged. The format of this key would be `Ns_Coll_Hash(PrivateKey)_blkNum_txNum`. For each version of a private key in the private data store, we would maintain such an index key. For illustration, assume that a key `key1` under collection `myColl` in chaincode `myCC` is inserted at block 100, transaction 10. Later this key is updated in block 200, transaction 20 and finally, the key is deleted in block 300, transaction 30. In this case, today, we maintain all three private write-sets that contain, potentially along with other, the above three mentioned operations. In this example, for `key1`, we would maintain three HashedIndex keys as `myCC_myColl_hash(key1)_100_10`, `myCC_myColl_hash(key1)_200_20`, and `myCC_myColl_hash(key1)_300_30`. The value for the HashedIndex key will be the private key itself. For instance, in the above two examples, the values would be `key1` and `key2` respectively.
```
denyeart:
```suggestion
   A HashedIndex key is intended to act as an index for the data keys in private data store, so that when a key is eventually purged we know from where to purge it in the private data store. The format of this key would be `Ns_Coll_Hash(PrivateKey)_blkNum_txNum`. For each version of a private key in the private data store, we would maintain such an index key. For illustration, assume that a key `key1` under collection `myColl` in chaincode `myCC` is inserted at block 100, transaction 10. Later this key is updated in block 200, transaction 20 and finally, the key is deleted in block 300, transaction 30. In this case, today, we maintain in the private data store all three private write-sets that contain, potentially along with other, the above three mentioned operations. In this example, for `key1`, we would maintain three HashedIndex keys as `myCC_myColl_hash(key1)_100_10`, `myCC_myColl_hash(key1)_200_20`, and `myCC_myColl_hash(key1)_300_30`. The value for the HashedIndex key will be the private key itself. For instance, in the above two examples, the values would be `key1` and `key2` respectively.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+   
+   - Note that the format of this key look same as the format of the PurgeMarker key - however, we always prepend a unique byte for the key type so, no two keys of differnt type ever clash.
+   
+   - Unlike the PurgeMarker key, we delete an HashedIndex key when we delete the corresponding data, either because of newly introduce explicit purge operation in this RFC or because of an implicit purge operation caused by [expiry of private data](https://hyperledger-fabric.readthedocs.io/en/release-2.2/private_data_tutorial.html#purge-private-data).
+   
+   - An important point to note here is that we would need to create these index keys retroactively - i.e., when a peer is upgraded to a version that would introduce private data purge feature, as a part of the upgrade, we would create these index keys.
```
denyeart, line 94: We'll create both index entries retroactively upon upgrade? Or just the HashedIndex entries?

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- Compute HashedIndex keys for each private key-value present in the private write-set.
+
+
+#### Private data commit via reconciliation
+As mentioned previously, private data reconcilation is going to be most affected of the existing functions. Below is a list of changes that we would make in the reconcilation.
+
+1) When we commit private data via reconciliation, the peer may recieve the keys that have already been purged. This would be the case when the sending peer has not yet seen the purge transaction. However, the receiving peer may have seen the purge transaction. In order to not save the purged keys, we would first compute the expected private write-set by removing the already purged keys on the recieving peer from the hashed write-set from block, by intersecting with PurgeMarker keys. This implicitly means that we need to maintain PurgeMarker keys forever (theoretically, at least, until no more missing data, including the missing data for ineligible collections, exists below the committing block of PurgeMarker key).
```
denyeart:
```suggestion
1) When we commit private data via reconciliation, the requesting peer may receive the keys that have already been purged. This would be the case when the replying peer has not yet seen the purge transaction. However, the requesting peer may have seen the purge transaction. In order to not save the purged keys, we would first compute the expected private write-set by removing the already purged keys on the requesting peer from the hashed write-set from block, by intersecting with PurgeMarker keys. This implicitly means that we need to maintain PurgeMarker keys forever (theoretically, at least, until no more missing data, including the missing data for ineligible collections, exists below the committing block of PurgeMarker key).
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+2) In the current code, during reconciliation, a peer verifies the data by computing the hash of the private write-set received against the hash of the collection present in the corresponding transaction in the block. On a side note, if a peer has been bootstrapped via a snapshot, this is true only for the private data that has been added after the height of the bootstrapping snapshot. Now, in the presence of the private data purge feature at a key level, a peer may receive a partial private write-set, because one or more keys in the private write-set may have been purged by the supplier peer. Hence, it is required for the receiving peer to perform the verification of hashes at the level of individual key-value pairs.
```
denyeart:
```suggestion
2) In the current code, during reconciliation, a requesting peer verifies the data by computing the hash of the private write-set received against the hash of the collection present in the corresponding transaction in the block. On a side note, if a peer has been bootstrapped via a snapshot, this is true only for the private data that has been added after the height of the bootstrapping snapshot. Now, in the presence of the private data purge feature at a key level, a peer may receive a partial private write-set, because one or more keys in the private write-set may have been purged by the replying peer. Hence, it is required for the requesting peer to perform the verification of hashes at the level of individual key-value pairs.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+3) During reconciliation, a peer may receive a subset of the expected write-set, as computed in (1) above, including an empty write-set, because the receiving peer may yet not have seen the purge transaction. To handle this, the receiving peer accepts the subset of the expected write-set and puts the missing data request into the deprioritized list and is attempted again at a low frequency interval. Eventually, the peer executes the purge transaction and gets aware of the purge transaction. finally, when the expected keys, as computed in (1) above, becomes the empty-set, the peer updates the missing keys info as fully reconciled.
```
denyeart:
```suggestion
3) During reconciliation, a requesting peer may receive a subset of the expected write-set, as computed in (1) above, including an empty write-set, because the requesting peer may yet not have seen the purge transaction while the replying peer has seen it. To handle this, the requesting peer accepts the subset of the expected write-set and puts the missing data request into the deprioritized list and is attempted again at a low frequency interval. Eventually, the requesting peer catches up to block height and executes the purge transaction and gets aware of the purge transaction. Finally, when the expected keys, as computed in (1) above, becomes the empty-set, the peer updates the missing keys info as fully reconciled.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Queries on Private Data Store
+The queries need to filter the private write-sets that it sends out as a response to a query. This is intended to be achieved with the help of PurgeMarker key, if the version of the key is lower than the version present in one of the corresponding PurgeMarkerKey keys. This is required because the actual data purge may take some time after having created the marker keys.
```
denyeart, line 114: Which queries are you talking about? Both the reconciliation queries, as well as the DeliverWithPrivateData queries?

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
...s-and-requirements), we intend to perform the actual deletions of the keys in the backgound. Therefore, at the time of processing a private data purge transaction, we only insert the marker key at the time of block commit, as described in the section [Block commit](#block-commit). In order to perform the actual purge of the data, we run a background goroutine that reads the PurgeMarker keys and scans all the relevant HashedIndex keys such that the index keys contain lower blkNum_TxNum than what is present in the PurgeMarker key. Finally, we would use the index key to locate the actual private write-set and we would trim the write-set by purging the intended key. While we purge the historical versions from the private data store, we would also delete the private key from the private state maintained in the statedb. It would be possible to derive the actual private key from the key-hash with the use of HashedIndex key. In the implementation, we intend to add an info field in the Marker key that would indicate whether the particular marker key has been processed. Also, if another purge operation is specified on the same key in the future, we delete the existing PurgeMarker for the key
```
denyeart: "While we purge the historical versions from the private data store, we would also delete the private key from the private state maintained in the statedb."
Wasn't the private state already deleted at time of block commit?

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Creating the HashedIndex keys retroactively
+As mentioned in the section [New Keys in Private data store](#new-keys-in-private-data-store), for the proposed feature, we would need to create the HashedIndex keys for allready committed private data in the private data store. We intend to do this atomatically at the start of peer using the framework we introduced in data format upgrade in version 2.0
```
denyeart:
```suggestion
As mentioned in the section [New Keys in Private data store](#new-keys-in-private-data-store), when starting the peer version that contains purge support for the first time, we would need to retroactively create the HashedIndex keys for already committed private data in the private data store. We intend to do this automatically at the start of peer using the framework we introduced in data format upgrade in version 2.0
```

text/0000-private_data_purge.md
manish-sethi: `PurgeMarker` is created upon processing a purge transaction, which will be the case only after peer has been upgraded. So, there won't be anything to create for the `PurgeMarker` retroactively.

text/0000-private_data_purge.md
manish-sethi: Yes, both. The overall goal is not to leak data between `having processed the purge transaction`  and `physically deleted the data`

text/0000-private_data_purge.md
manish-sethi: In general, we need access to raw private key in order to delete it from private state and the raw key may not necessarily arrive with the block. In one particular scenario that I mentioned in line 52, it may never arrive. So, we need to rely upon HashedIndexEntry even for deleting from the private state. As mentioned in line 48 and 49, the intention is to delete from only hashed state during block commit and delete from private state in the background along with updating the private data store.

text/0000-private_data_purge.md
manish-sethi: Here, I indeed actually meant private state not the private data store.

denyeart (approved):

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- Feature Name: private_data_purge)
+- Start Date: 2021-10-01)
+- RFC PR: (leave this empty)
+- Fabric Component: chaincode, ledger
+- Fabric Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Private data collections feature in Fabric allows the dissemination of the sensitive data to a subset of organizations in a channel. A peer that is eligible to receive the private data of a collection within a namespace, maintains all the versions of the private data in its private data store forever. This RFC proposes a new operation *purge* on private data via chaincode shim that allows the deletion of one or more private data keys from the current state (statedb), and in addition, purging of all the historical versions of the supplied keys from the private data store permanently. In addition, it lays down a basic ground work for enabling history related queries on private data.
```
cendhu:
```suggestion
Private data collections feature in Fabric allows the dissemination of the sensitive data to a subset of organizations in a channel. A peer that is eligible to receive the private data of a collection within a namespace, maintains all the versions of the private data in its private data store forever. This RFC proposes a new operation *purge* on private data via chaincode shim that allows the deletion of one or more private data keys from the current state (statedb), and in addition, purging of all the historical versions of the supplied keys from the private data store permanently. In addition, it lays down a basic groundwork for enabling history related queries on private data.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The primary motivation of this feature is to enable users to implement right to forget in the context of private data in a Fabric application. This may help users to meet certain business compliance about data privacy. At present, there exists two ways to retrieve historical versions of private data. An end user can retrieve historical versions of private data from a peer using the block Deliver service (`DeliverWithPrivateData` API). Additionally, other peers, potentially from other organizations, can access to the historical versions via the private data reconciliation process. Further, in future, we may expose an API for users to access the history of a specific private key in a collection, similar to the existing API for accessing the history of a key in the chaincode namespace. By enabling the purging of desired private data, the proposed feature makes peer prevent sending private data out from these APIs and eventually deletes it permanently from its persistent stores. An added advantage of this feature is that it enables users to free up some storage space by purging the historical versions of private data.
```
cendhu, line 21:
```suggestion
The primary motivation of this feature is to enable users to implement right to forget in the context of private data in a Fabric application. This may help users to meet certain business compliance about data privacy. At present, there exists two ways to retrieve historical versions of private data. An end user can retrieve historical versions of private data from a peer using the block Deliver service (`DeliverWithPrivateData` API). Additionally, other peers, potentially from other organizations, can access to the historical versions via the private data reconciliation process. Further, in future, we may expose an API for users to access the history of a specific private key in a collection, similar to the existing API for accessing the history of a key in the chaincode namespace. By enabling the purging of desired private data, the proposed feature prevents peers from sending private data out from these APIs and eventually deletes it permanently from its persistent stores. An added advantage of this feature is that it enables users to free up some storage space by purging the historical versions of private data.
```

makes peer prevent -> prevents peers from

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+For understanding the usage of this API, lets assume a scenario where a chaincode defines a private data collection and the collection contains personally identifiable information for customers such that the chaincode maintains PII info of a customer in the form of three keys, namely {CustomerID}_PII_Name, {CustomerID}_PII_DOB, and {CustomerID}_PII_Address. While name and date-of-birth may not change, a customer may have updated his address a few times and hence peers of eligible orgs would have persisted all the historical values for the address. Now, the customer may want the Fabric network to purge any of its PII data.
+
+In order to achive this, the application may invoke a chaincode transaction in which the chaincode in turn invokes the *PurgePrivateData* API three times - once for each of the above mentioned PII keys. Finally, the application would get enough endorsements to satisfy the endorsement policies for each of the above keys and submits the transaction to the orderers. Once the transaction is committed on a peer, the peer stops sending out any version of any of these private data keys and eventually deletes from the persistent stores.
```
cendhu:
```suggestion
In order to achieve this, the application may invoke a chaincode transaction in which the chaincode in turn invokes the *PurgePrivateData* API three times - once for each of the above mentioned PII keys. Finally, the application would get enough endorsements to satisfy the endorsement policies for each of the above keys and submits the transaction to the orderers. Once the transaction is committed on a peer, the peer stops sending out any version of any of these private data keys and eventually deletes from the persistent stores.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+Before diving into the proposed design, we describe some background, challenges, and broad requirements for the desired behavior for implementing private data purge.
+
+## Background, challenges, and Requirements
+We divide this section into two subsections.
+
+### Deletion of private data from statedb and private data store
+In regard of actual deletion of private data keys specified in a purge transaction, following is the desired behaviour of a peer.
+- Deletion of the private data key from the state should be done atomically with the other state updates caused by the transaction. In other words, as far as state updates are concerned from the prespective of future transaction simulation, the *purge* operation is no different from *delete* and merely deleting from hashed state is sufficient.
```
cendhu:
```suggestion
- Deletion of the private data key from the state should be done atomically with the other state updates caused by the transaction. In other words, as far as state updates are concerned from the perspective of future transaction simulation, the *purge* operation is no different from *delete* and merely deleting from hashed state is sufficient.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+- In addition, in order to permanently deleting all traces of the private data, the key should eventually be deleted from the private state (maintained in statedb). However, it not required to perform this action atomically with the commit of the transaction and can be performed in the background.
```
cendhu:
```suggestion
- In addition, in order to permanently deleting all traces of the private data, the key should eventually be deleted from the private state (maintained in statedb). However, it is not required to perform this action atomically with the commit of the transaction and can be performed in the background.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+- The requesting peer accepts and verifies subset of the collection
+  - Contarary to the above situation, the requesting peer may be behind and may not be aware of the purge transaction yet. While, the replying peer sends the subset of the collection write-set. In this scenario, the requesting peer needs to verify the subset by comparing the hashes of the individual keys (as opposed to the hash of the entire collection write-set). Moreover, the requesting peer cannot remove the collection from its list of still missing data, until the requesting peer itself sees the purge transaction, as the replying peer may be playing malicious by deliberately sending a selective subset.
```
cendhu:
```suggestion
  - Contrary to the above situation, the requesting peer may be behind and may not be aware of the purge transaction yet. While, the replying peer sends the subset of the collection write-set. In this scenario, the requesting peer needs to verify the subset by comparing the hashes of the individual keys (as opposed to the hash of the entire collection write-set). Moreover, the requesting peer cannot remove the collection from its list of still missing data, until the requesting peer itself sees the purge transaction, as the replying peer may be playing malicious by deliberately sending a selective subset.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+Keeping the above requirements and desired behavior in consideration, below is the the proposed design. First, we describe the new key types that we would maintain in the private data store and then, we discuss various operations in Fabric that would create, update, read, or delete these keys to implement this feature.
+
+## Proposed Design
+We divide the proposed design in four subsections. First, we describe the changes in the Write-set Proto message. Second, we introduce two new key types that we would maintain in the private data store. Third, we describe the proposed changes in the existing functions and finaly, we describe the new functions that we would add.
```
cendhu:
```suggestion
We divide the proposed design in four subsections. First, we describe the changes in the Write-set Proto message. Second, we introduce two new key types that we would maintain in the private data store. Third, we describe the proposed changes in the existing functions and finally, we describe the new functions that we would add.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+### New Keys in Private data store
+ In this design, we maintain two types of additional keys in the private data store. We refer to these new types of keys as *PurgeMarker* and *HashedIndex*. Below, is the format and intended use of these keys.
+1) PurgeMarker
+   A PurgeMarker key is intended to record a purge operation for a key. The design of this marker key would be `Ns_Coll_Hash(PrivateKey)_CommittingBlkNum_CommittingTxNum`. For illustration, assume that a purge transaction appears in block number 100, at position 20 that specifies for purging the keys `key1` and `key2` under collection `myColl` for chaincode `myCC`. When processing this transaction, we insert the two keys into private data store as `myCC_myColl_hash(key1)_100_20` and `myCC_myColl_hash(Key2)_100_20`. Note that the separation of different components of a key shown by an understore is just for the illustration. For the reasons discussed latter, we intend to keep these marker keys forever. However, the assumption is that the purge operation is expected to be relatively less frequent, and hence, it should not be a concern from storage space point of view.
```
cendhu:
```suggestion
   A PurgeMarker key is intended to record a purge operation for a key. The design of this marker key would be `Ns_Coll_Hash(PrivateKey)_CommittingBlkNum_CommittingTxNum`. For illustration, assume that a purge transaction appears in block number 100, at position 20 that specifies for purging the keys `key1` and `key2` under collection `myColl` for chaincode `myCC`. When processing this transaction, we insert the two keys into private data store as `myCC_myColl_hash(key1)_100_20` and `myCC_myColl_hash(Key2)_100_20`. Note that the separation of different components of a key shown by an underscore is just for the illustration. For the reasons discussed latter, we intend to keep these marker keys forever. However, the assumption is that the purge operation is expected to be relatively less frequent, and hence, it should not be a concern from storage space point of view.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+   
+   - Note that the format of this key look same as the format of the PurgeMarker key - however, we always prepend a unique byte for the key type so, no two keys of differnt type ever clash.
```
cendhu:
```suggestion
   - Note that the format of this key look same as the format of the PurgeMarker key - however, we always prepend a unique byte for the key type so, no two keys of different type ever clash.
```

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+   
+   - Unlike the PurgeMarker key, we delete an HashedIndex key when we delete the corresponding data, either because of newly introduce explicit purge operation in this RFC or because of an implicit purge operation caused by [expiry of private data](https://hyperledger-fabric.readthedocs.io/en/release-2.2/private_data_tutorial.html#purge-private-data).
```
cendhu:
```suggestion
   - Unlike the PurgeMarker key, we delete an HashedIndex key when we delete the corresponding data, either because of newly introduced explicit purge operation in this RFC or because of an implicit purge operation caused by [expiry of private data](https://hyperledger-fabric.readthedocs.io/en/release-2.2/private_data_tutorial.html#purge-private-data).
```

cendhu (commented):
Design part looks good to me.

cendhu (approved):

text/0000-private_data_purge.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order to achieve this, the application may invoke a chaincode transaction in which the chaincode in turn invokes the *PurgePrivateData* API three times - once for each of the above mentioned PII keys. Finally, the application would get enough endorsements to satisfy the endorsement policies for each of the above keys and submits the transaction to the orderers. Once the transaction is committed on a peer, the peer stops sending out any version of any of these private data keys and eventually deletes from the persistent stores.
+
+Note that we will gate this new API behind a capability tag so that this new operation is interpreted on all the peers in a channel and assures the user that the data will be purged if the transaction is committed as a valid transaction. In the absence of the capability, the effect of the API will be exactly same as invoking the API `DeletePrivateData`.
```
jt-nti: If the private data purge capability is not enabled for a channel, calls to `PurgePrivateData` should fail.

jt-nti (changes requested):
I understand that there may be some edge cases which make it difficult but there should not be any doubt in the chaincode that private data is being purged rather than deleted.

text/0000-private_data_purge.md
manish-sethi: Yes, we can very well return error during endorsement however, irrespective of this, the important point is to define the behavior at commit path. During the commit, as mentioned above in line 80, if we don't treat this as a delete, the older version of the peer would cause `panic`.

manish-sethi: @jt-nti - I gave it a bit more thoughts and we can indeed gate this at the time of endorsement only as it's an entirely new operation. I have updated the text - Thanks for the feedback.

denyeart:
Final comment period has ended, I'll go ahead and merge with three maintainers in agreement.

PR #49, opened 2022-02-03
Gossip block dissemination re-architecture

yacovm:


text/0011-block-dissemination.md
```diff
@@ -0,0 +1,215 @@
...
+# The proposed new block dissemination method
+[Proposed]: #Proposed
+
+An alternative block dissemination method is proposed to replace the current block dissemination.
+
+**Note:** Whenever an "orderer" is mentioned, we mean some orderer among the available orderers via some selection policy that is orthogonal to this proposal.
+
+The proposal is to be implemented in two phases:
+
+**Phase I**: Gossip block replication among peers will cease to operate, and the relevant code will be removed. This means that both gossip block replication layer and leader election layers will be removed and each peer will always pull only from the ordering service.
```
SamYuan1990, line 106: Question here:
Assuming we have 5 peers(from different org) and 5 orderers(running in raft).
Will those 5 peers pull block from raft leader or those 5 peers able to pull block from one of orderer nodes randomly?

yacovm: The peers don't know who is the leader anyway, so they have to use an arbitrary node

hxx258456:
Will newly joined nodes be affected when they perform block synchronization?

PR #50, opened 2022-03-07
Ordering service framework

yacovm:
This RFC suggests to decouple the Fabric ordering service codebase into
a framework that can be used for implementing various consensus mechanisms
without changing the Fabric codebase itself, in contrast to the current state of affairs.

It also proposes an alternative authentication mechanism for inter-orderer communication.

Signed-off-by: Yacov Manevich <yacovm@il.ibm.com>

### DIFF
--- a/text/orderer-v3.md
+++ b/text/orderer-v3.md
@@ -1,0 +1,320 @@
+---
+layout: default
+title: Fabric ordering service consensus agnostic framework 
+nav_order: 3
+---
+
+- Feature Name: A consensus agnostic framework for building a Fabric ordering service  
+- Start Date: (fill me in with today's date, 2022-03-01)
+- RFC PR: (leave this empty)
+- Fabric Component: orderer
+- Fabric Issue: (leave this empty)
+
+
+# Table of Contents
+1. [Summary](#summary)
+2. [Motivation](#motivation)
+3. [Ordering service framework building blocks](#identifying-building-blocks)
+4. [Ordering service node construction](#connecting-the-pieces-together)
+5. [Orderer to orderer communication and authentication](#orderer-to-orderer-authentication-and-communication)
+6. [Testing](#testing)
+
+# Summary
+[summary]: #summary
+
+The Fabric ordering service node is designed to support multiple consensus protocols through an abstraction of consensus plugins.
+Consensus plugins order transactions into blocks on behalf of the Fabric ordering node, and the latter provides consensus plugins with 
+networking, storage and cryptography software infrastructure.
+Unfortunately, the consensus plugin approach not only makes it hard to integrate new types of consensus protocols into the Fabric ordering 
+node, but also induces un-needed semantic dependency of the rest of the orderer codebase on the consensus plugins.  
+
+This RFC proposes a paradigm for structuring the ordering service node that can be seen as the opposite one to the consensus plugin one.
+Simply put, instead of having the ordering service node contain multiple consensus protocol implementations, it proposes to decouple 
+the ordering service node into discrete software packages and then to have the consensus plugins tie them together, forming ordering 
+service node binaries per consensus protocol. 
+
+Additionally, the RFC discusses the current approach for orderer to orderer authentication and proposes an alternative design
+which is more friendly to consume and deploy.
+
+This RFC aims to:
+
+- Define building blocks that consensus protocols may require to be integrated to Fabric.
+- Outline how the aforementioned building blocks can be used to construct Fabric ordering service nodes.
+- Design a more consumer friendly orderer to orderer authentication scheme.
+
+# Motivation
+[motivation]: #motivation
+
+
+The Hyperledger Fabric ordering service is designed to support various types of consensus protocols.
+The support for each consensus protocol is manifested in the form of a "plugin", a package that wraps around a consensus protocol implementation
+and connects the rest of the ordering service code base to the consensus protocol.
+
+The Fabric orderer codebase currently embeds the following consensus plugins:
+
+- **Solo**: A single node orderer, useful for development, testing and proof of concepts.
+- **Kafka**: Several ordering nodes relay transactions into an [Apache Kafka](https://kafka.apache.org/) system,
+  pull transactions back in the same order and then cut them into identical blocks via synchronization messages also sent through Kafka.
+- **Raft**: Several ordering nodes run the [Raft](https://raft.github.io/) protocol by embedding the [Raft library of etcd](https://github.com/etcd-io/etcd/tree/main/raft).
+
+Apart from the official Hyperledger Fabric consensus implementation, there is also a [Fabric fork](https://github.com/SmartBFT-Go/fabric) of a [BFT library](https://github.com/SmartBFT-Go/consensus) called SmartBFT.
+
+A BFT consensus protocol might have a different approach for onboarding and state snapshot replication compared to a CFT consensus protocol.
+In particular, a state snapshot sent between nodes in CFT is always assumed to be trusted, while in BFT it is not the case. 
+While it is tempting to have common code that performs state snapshot transfer that caters to both adversarial models, 
+such code is most likely to be complex and hard to follow and test.
+
+Each orderer type requires Fabric to understand how to initialize instances of it from the consensus metadata in the configuration blocks.
+However, this makes it impossible for external contributors to use their own consensus implementation without either forking Fabric entirely, 
+or making intrusive changes in the official Fabric orderer and peer code base.
+
+Undoubtedly, an alternative architecture which doesn't require Fabric codebase changes to add support for new 
+consensus protocols would be a sought-after property.
+
+# Identifying building blocks
+[building-blocks]: #identifying-building-blocks
+
+Apart from consensus, which agrees on orders of transaction, the orderer contains infrastructure that manages various aspects of its lifecycle and operations:
+
+ - **Chain management**: Spawns instances of the consensus protocols per channel and dispatches messages and operations to the right 
+   consensus instance.
+ - **Block replication service**: Server-side API that sends blocks upon request. 
+ - **Block synchronization utility**: Pulls blocks from other ordering service nodes and appends them to the local ledger. 
+ - **Communication service**: Authenticates client connections, receives messages from remote nodes and sends message to remote nodes.
+ - **Ledger subsystem**: Appends blocks to persistent storage and retrieves blocks by their index.
+ - **Channel participation service**: Manages local addition and removal of ledgers per channel.
+ - **Transaction filtering**: Enforces access control and validity checks on transactions both received from clients or from nodes.
+ - **Membership Service Provider subsystem**: Classifies identities into principals and drives policy authorization. 
+
+In this RFC, it is proposed that these building blocks be made consensus independent, and the building blocks that cannot be consensus independent will be 
+part of the consensus plugin that wraps around the consensus independent building blocks. 
+
+Therefore, the new organization of the components in the new ordering service node of a given type will look as follows:
+
+```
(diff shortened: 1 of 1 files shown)


text/orderer-v3.md
```diff
@@ -0,0 +1,320 @@
...
+   consensus instance.
+ - **Block replication service**: Server-side API that sends blocks upon request. 
+ - **Block synchronization utility**: Pulls blocks from other ordering service nodes and appends them to the local ledger. 
+ - **Communication service**: Authenticates client connections, receives messages from remote nodes and sends message to remote nodes.
+ - **Ledger subsystem**: Appends blocks to persistent storage and retrieves blocks by their index.
+ - **Channel participation service**: Manages local addition and removal of ledgers per channel.
+ - **Transaction filtering**: Enforces access control and validity checks on transactions both received from clients or from nodes.
+ - **Membership Service Provider subsystem**: Classifies identities into principals and drives policy authorization. 
+
+In this RFC, it is proposed that these building blocks be made consensus independent, and the building blocks that cannot be consensus independent will be
```
Param-S, line 128: > In this RFC, it is proposed that these building blocks be made consensus independent,

I think, some of these building blocks map to the orderer specific common code which is present under https://github.com/hyperledger/fabric/tree/main/orderer/common and fabric core specific common code(deliver service) https://github.com/hyperledger/fabric/tree/main/common/deliver which is common for both peer & orderer node. 

The common implementation(like deliver service) for both core fabric & orderering service where we need to handle the inter dependencies.

yacovm: That is correct, a lot of the code is indeed independent, not not all of it. 

For example, onboarding which is in **orderer/common**, currently [imports Raft](https://github.com/hyperledger/fabric/blob/main/orderer/common/onboarding/onboarding.go#L110).

denyeart: I think it is worth highlighting which of the building blocks can largely be re-used from v2.x. And for the building blocks that require significant changes, what is the nature of those changes.

text/orderer-v3.md
```diff
@@ -0,0 +1,320 @@
...
+
+
+# Connecting the pieces together
+[connecting]: #connecting-the-pieces-together
+
+Now, having common and consensus independent building blocks that perform various functions,
+an ordering service for a given consensus protocol can be implemented without changing Fabric codebase. 
+
+The developers who implement the ordering service node will just import the needed packages from the Fabric ordering service framework,
+and will implement a binary that services the [Atomic Broadcast gRPC API](https://github.com/hyperledger/fabric-protos/blob/main/orderer/ab.proto#L80-L86).
```
denyeart, line 183: Isn't the part that implements the client facing Atomic Broadcast API in the common building blocks rather than in the consensus-specific code? I assumed at least the Atomic Broadcast Deliver service would be implemented by the common Block Replication Service.

Also I think how to implement a binary deserves a whole section, answering questions like:
- What exactly is necessary to implement it?
- What does the consensus plugin look like (I assume it is not a Go plugin per se)?
- Is there a specific consensus interface in mind yet? (I realize it may change during implementation phase)
- What assumptions are made about the types of consensus libaries that can be used? (I assume not any BFT library could be used)
- Is it expected that some integration code would be required to tie the consensus library into the ordering node consensus interface?

text/orderer-v3.md
```diff
@@ -0,0 +1,320 @@
...
+ - **Ledger subsystem**: Appends blocks to persistent storage and retrieves blocks by their index.
+ - **Channel participation service**: Manages local addition and removal of ledgers per channel.
+ - **Transaction filtering**: Enforces access control and validity checks on transactions both received from clients or from nodes.
+ - **Membership Service Provider subsystem**: Classifies identities into principals and drives policy authorization. 
+
+In this RFC, it is proposed that these building blocks be made consensus independent, and the building blocks that cannot be consensus independent will be 
+part of the consensus plugin that wraps around the consensus independent building blocks. 
+
+Therefore, the new organization of the components in the new ordering service node of a given type will look as follows:
+
```
denyeart, line 132: Can you be more explicit about an "ordering service node of a given type". That is, highlight that an ordering service node of type Solo, Raft, and SmartBFT would be three entirely different binaries.

When first looking at the diagram I thought Solo, Raft, and SmartBFT were each consensus instances. But then I realized there would be a single consensus type per binary and the instances were per channel instances of that consensus type. This should be clarified to avoid confusion.

text/orderer-v3.md
```diff
@@ -0,0 +1,320 @@
...
+In case an attacker manages to obtain the TLS private key of the responder node and performs a MITM attack, the function `ExportKeyingMaterial`
+on the initiator node and the real responder node would yield different values, because the `master_secret` will differ.
+The `master_secret` will differ because Fabric orderers will be restricted to use ECDHE key exchange. With ECHDE key exchange, a MITM with the same `master_secret` 
+across both the initiator and responder means that the initiator and responder each picked the public key share, but the attacker managed to compute the shared key,
+thus breaking the [CDH assumption](https://en.wikipedia.org/wiki/Computational_Diffie%E2%80%93Hellman_assumption).
+
+In case the responder node is located behind a TLS terminating reverse proxy, it can be configured to ignore the `session_binding` check, 
+but the check will be enabled by default.
+
+# Testing
```
denyeart: For a change of this scale, I think this section should include "Implementation Plan and Testing".

For example will the existing fabric repo be updated to support both the old v2.x orderer node and new v3.0 orderer node in parallel? And then remove the old v2.x code before 3.0 release?

Will the existing fabric repo support Solo and Raft and SmartBFT orderer binaries? Or will each of the orderer implementations be delivered in different repositories that import the base Fabric orderer Go modules?

Will the TLS pinning approach continue to be supported to enable rolling upgrades, perhaps in a deprecated mode so that it can be retired in a future release?

denyeart (commented):
I think this is a solid plan. I like the direction, just would like to see some more details as highlighted in the comments.

text/orderer-v3.md
yacovm: Addressed the comments modulo BFT which is not discussed in that RFC.

text/orderer-v3.md
yacovm: > Isn't the part that implements the client facing Atomic Broadcast API in the common building blocks rather than in the consensus-specific code? 

It doesn't matter if you can use a package that implements it. The orderer can be externally seen as a gRPC service for the atomic broadcast API. How you implement it, is up to you.


>  I assumed at least the Atomic Broadcast Deliver service would be implemented by the common Block Replication Service.

It already is, see the latest revision to the RFC.


> Also I think how to implement a binary deserves a whole section, answering questions like:
>  -   What exactly is necessary to implement it?
>  -  What does the consensus plugin look like (I assume it is not a Go plugin per se)?
>  -  Is there a specific consensus interface in mind yet? (I realize it may change during implementation phase)
>  -   What assumptions are made about the types of consensus libaries that can be used? (I assume not any BFT library could be used)
>  -   Is it expected that some integration code would be required to tie the consensus library into the ordering node consensus interface?


That's the thing - there is no "consensus plugin" in the new architecture because each orderer binary now only has a single type of consensus. Hence, there is no consensus interface anymore because no one is expected to invoke it. 
It is the other way around - the orderer is the consensus and it uses the non-consensus specific building blocks as it needs, according to its consensus implementation.

text/orderer-v3.md
yacovm: > and the instances were per channel instances of that consensus type. This should be clarified to avoid confusion.

Well it makes little sense to have an orderer that participates in CFT in one channel and in BFT in another channel. Either you trust the organizations in your network or you don't. 

And if you want to connect to two different and mutually exclusive networks you can always just run another orderer. 

It's not like we have 100 different consensus implementations and you need to use each of them.

text/orderer-v3.md
yacovm: I did in the revised version.

C0rWin (dismissed):

C0rWin:
Really thrilled to see, Fabric finally having BFT support.

text/orderer-v3.md
```diff
@@ -0,0 +1,394 @@
...
+
+message Auth {
+ ...
+ string channel = 7;
+ ...
+}
+
+```
+
+The `ConsensusRequest` and `SubmitRequest` will be defined similarly as the [v2.x gRPC API](https://github.com/hyperledger/fabric-protos/blob/release-2.4/orderer/cluster.proto),
```
Param-S, line 256: I hope, we don't remove the `channel` field from message `ConsensusRequest` for backward compatibility, otherwise it requires the downtime for version upgrade.

yacovm: If we would need a downtime, it wouldn't be because of switching to a different protobuf schema.

text/orderer-v3.md
```diff
@@ -0,0 +1,394 @@
...
+However, this approach has several drawbacks: 
+1. **TLS keys might not be stored in HSM**: The Go TLS stack doesn't provide HSM integration capabilities, and therefore TLS private keys of orderer nodes are more susceptible for theft. 
+A theft of a TLS private key means that an adversary may send messages on behalf of the victim node, which is catastrophic in a crash fault tolerant (CFT) setting.
+Although it is possible to override the `PrivateKey` reference of the `tls.Certifciate` with a signer implementation that communicates with an HSM, existing users which already have key management systems 
+   built will be reluctant to make the required changes.
+   
+
+
+2. **Mutual TLS authentication is challenging when combined with reverse proxies**: In case the organization hosting the orderer has a TLS terminating reverse proxy that routes ingress traffic into the organization,
+it is challenging for that organization to host an ordering service node that relies on mutual TLS authentication.
```
manish-sethi, line 296: Wouldn't the `ExportKeyingMaterial` approach mentioned below also has this limitation?

text/orderer-v3.md
manish-sethi: I was expecting a lot more details in this section. Primarily, the interfaces that the non-consensus specific building blocks would expose and how the orderer is expected to consume these with an example - say, how a `raft` based orderer would consume those interfaces at different occasions for achieving the consensus - mainly covering the important functionalities such as transaction submission, reconfiguration, and block cutting - as the primary motivation of this RFC is for others to understand the functions provided by these services so they can build orderer using their preferred consensus algorithm.

text/orderer-v3.md
yacovm: I address that in the quote below:


> In case the responder node is located behind a TLS terminating reverse proxy, it can be configured to ignore the `session_binding` check,  but the check will be enabled by default.

text/orderer-v3.md
yacovm: @manish-sethi I'm not sure explicit interfaces for building blocks should be in an RFC. I also don't want to bind us to a specific contract should we later discover we need to make changes. 

Furthermore, we currently have the existing subcomponents in our code base, and they have existing APIs. I do not expect them to change too much so I'm not sure why it's important to specify in an RFC that a block replication API gets as input a block number and returns a block, and a communication API receives a message and a destination, etc. 

As for transaction submission, reconfiguration and block cutting - these are all completely orthogonal to this RFC, since they deal with the consensus itself. Here, we deal with what components a consensus needs. 
It is up to the consumer of the Fabric ordering service framework to decide how transactions are submitted, how nodes are being re-configured and how transactions are cut into batches.

andrew-coleman (approved):

C0rWin (approved):
I think this is a good approach to assuring Fabric with a more flexible mechanism to enable various consensus modules.

denyeart (approved):

denyeart:
Three approvals...merging.

PR #52, opened 2022-09-04
Byzantine Fault Tolerant ordering service

yacovm:
This commit details the addition of a BFT ordering service based on the SmartBFT consensus library.

It builds upon two previous RFCs:

- An [RFC](https://hyperledger.github.io/fabric-rfcs/text/0010-bft-signatures.html) that describes Byzantine block signature verification and replication.

- An [RFC](https://hyperledger.github.io/fabric-rfcs/text/orderer-v3.html) that describes inversion of control between the consensus and the rest of the ordering service codebase.

Signed-off-by: Yacov Manevich <yacovm@il.ibm.com>

### DIFF
--- a/text/006-bft-based-ordering-service.md
+++ b/text/006-bft-based-ordering-service.md
@@ -127,7 +127,7 @@ Each node collects a quorum of `commit` messages and their attached signatures.
 These signatures are later used to notarize the proposal, essentially creating a checkpoint. 
 Upon receiving a quorum of validated `commit` messages, the node delivers its proposal and the signatures to the application layer for storage and processing.
 
-The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((N + F + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.  
+The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((n + f + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.  
 
 **View change**:
 The view change protocol is inspired by the _Synchronization Phase_ in this [BFT-SMaRt paper](http://www.di.fc.ul.pt/~bessani/publications/edcc12-modsmart.pdf) (illustrated in the attached figure taken from the BFT-SMaRt paper).
@@ -136,7 +136,7 @@ The view change protocol is comprised of three messages: `ViewChange`, `ViewData
 ![alt text](https://raw.githubusercontent.com/wiki/SmartBFT-Go/consensus/images/viewChange.png)
 
 The `ViewChange` message (the first message) is sent to all nodes by some node who suspects the leader is faulty. This message is very lightweight and includes only the next view sequence number. 
-Once collecting `2f+1` `ViewChange` messages, a non faulty node is convinced that enough nodes consider the leader as faulty, and it will send a `ViewData` message to the next view's leader.
+Once collecting `2f+1` `ViewChange` messages, a non faulty node is convinced that enough nodes consider the leader as faulty, and it will send a `ViewData` message to the next view's leader (in practice, it's the view number modulo the total number of nodes).
 
 The `ViewData` message is similar to the `ViewChange` message suggested by [PBFT](http://pmg.csail.mit.edu/papers/osdi99.pdf). It is signed, as it will serve as proof in the next phase. The message contains the last checkpoint (last decided proposal) and the next proposal, if it exists, with its state (proposed or prepared). 
 However, as opposed to the PBFT `ViewChange` message, the `ViewData` message is sent only to the next leader.
@@ -186,7 +186,7 @@ SubmitRequest(req []byte) error
 The library offloads communication to the application that embeds it, hence it is the responsibility of the application to feed the messages from other nodes, or requests forwarded from other nodes into the library instance:
 
 ````
-HandleMessage(sender uint64, m *Message)
+HandleMessage(sender uint64, m *bft.Message)
 
 HandleRequest(sender uint64, req []byte)
 ````
@@ -200,8 +200,7 @@ Deliver(proposal bft.Proposal, signature []bft.Signature) Reconfig
 The application is free to implement the callback as it sees fit in order to realize its desired functionality. 
 When embedded inside a blockchain application, the proposal would often denote a block constructed by the leader of that round, and the signatures array would contain a quorum of signatures collected by the library from the various nodes. The Reconfig is an object that allows the application to convey to the library to reconfigure itself (for example, to expand or reduce the number of nodes).
 
-**Dependencies**: The library interacts with the external world through the application, by delegating capabilities such as cryptography, communication and storage to the application. 
-The full list of dependencies can be found [here](https://smartbft-go.github.io/godoc/pkg/github.com/SmartBFT-Go/consensus/pkg/api/index.html), and below the important ones are listed:
+**Dependencies**: 
 
 The SmartBFT library delegates interaction with the external world (communication and storage) to the application that consumes it. It also doesn't come with its own signature scheme and also assumes its consumer
 implements it. All such abilities it delegates to the consumer application are termed "dependencies" and their full list can be found [here](https://smartbft-go.github.io/godoc/pkg/github.com/SmartBFT-Go/consensus/pkg/api/index.html).
@@ -286,13 +285,13 @@ Upon a config change, the BFT orderer will ensure that the block verification po
 While a malicious orderer cannot forge a block, it can deprive a peer from blocks. Hence, receiving the stream of blocks from a single orderer exposes the peer to what is called a censorship attack, or a block withholding attack.
 
 A [previous RFC](https://github.com/hyperledger/fabric-rfcs/pull/48) explains at length how this censorship attack is mitigated. 
-In brief, if an orderer suspects it is being censored, it reaches to other orderers and pulls block headers and their corresponding signatures.
-Optionally, once in a while, an orderer may also change the orderer from which it pulls blocks.
-This approach is different than the approach taken in the SmartBFT Fabric fork, where the orderer constantly pulls headers and signatures from several orderers.
+In brief, if an orderer or a peer suspects it is being censored, it reaches to other orderers and pulls block headers and their corresponding signatures.
+Optionally, once in a while, an orderer or a peer may also change the orderer from which it pulls blocks.
+This approach is different than the approach taken in the SmartBFT Fabric fork, where the peer constantly pulls headers and signatures from several orderers.
 The deviation is in order to be computationally efficient, as in overwhelming cases, the orderers will not withhold blocks.
 
 
-## Changes to client SDK
+## Changes to client SDK and Gateway service
 Since orderers can be malicious and thus either drop transactions or never include them in a block when assembling proposals, any client that interacts with the BFT ordering service needs to submit the transaction to all orderers.
 such a change is considered to be trivial and minimal and it is argued that it should be made in every SDK with ease.
 
--- a/text/006-bft-based-ordering-service.md
+++ b/text/006-bft-based-ordering-service.md
@@ -129,6 +129,12 @@ Upon receiving a quorum of validated `commit` messages, the node delivers its pr
 
 The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((n + f + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.  
 
+For example, with `7` nodes, `n=7` and therefore `f=2` because `(7-1)/3` is `2`. 
+Notice that with `8` nodes, we still have `f=2`, because the division (`/`) is in the integers and not in the real numbers.
+
+With `7` nodes we have an optimal quorum of `5` because it satisfies `2f+1`, but with `8` nodes the quorum is `6`, because `6` is the smallest group size for which two groups intersect with at least a single honest node (because they intersect by four nodes).
+Two groups of only five nodes, may intersect in only two nodes and they may both be malicious since `f=2`. 
+
 **View change**:
 The view change protocol is inspired by the _Synchronization Phase_ in this [BFT-SMaRt paper](http://www.di.fc.ul.pt/~bessani/publications/edcc12-modsmart.pdf) (illustrated in the attached figure taken from the BFT-SMaRt paper).
 The view change protocol is comprised of three messages: `ViewChange`, `ViewData`, and `NewView` (`Stop`, `StopData`, and `Sync` in BFT-SMaRt).
@@ -178,12 +184,12 @@ so overall the storage overhead of the WAL can be kept constant and not depend o
 
  The [SmartBFT library](https://github.com/SmartBFT-Go/consensus) presentes a minimal interface to the application that uses it:
 
-**Input**: To submit requests to be ordered, the application invokes `SubmitRequest` and passes a byte representation of the request.
+**Input**: To submit requests to be ordered, the application that embeds the library (in our case, the Fabric orderer node) invokes `SubmitRequest` and passes a byte representation of the request.
 ````
 SubmitRequest(req []byte) error
 ````
 
-The library offloads communication to the application that embeds it, hence it is the responsibility of the application to feed the messages from other nodes, or requests forwarded from other nodes into the library instance:
+The library offloads communication to the application that embeds it (the Fabric orderer node, in our case), hence it is the responsibility of the application to feed the messages from other nodes, or requests forwarded from other nodes into the library instance:
 
 ````
 HandleMessage(sender uint64, m *bft.Message)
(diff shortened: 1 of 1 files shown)


text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+
+Next each node waits for a quorum of `prepare` messages with the same digest, view `v`, and sequence number. 
+The quorum size should be big enough to ensure that in the intersection of any two quorums there is at least one correct node. So for example, if `n=3f+1` then the optimal quorum size is `2f+1`. 
+
+After the node is "prepared" it sends a `commit` message with the same content as the `prepare` message, as well as a signature on the proposal.
+Each node collects a quorum of `commit` messages and their attached signatures. 
+These signatures are later used to notarize the proposal, essentially creating a checkpoint. 
+Upon receiving a quorum of validated `commit` messages, the node delivers its proposal and the signatures to the application layer for storage and processing.
+
+The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((N + F + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.
```
andrew-coleman:
```suggestion
The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((n + f + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.  
```

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+
+**View change**:
+The view change protocol is inspired by the _Synchronization Phase_ in this [BFT-SMaRt paper](http://www.di.fc.ul.pt/~bessani/publications/edcc12-modsmart.pdf) (illustrated in the attached figure taken from the BFT-SMaRt paper).
+The view change protocol is comprised of three messages: `ViewChange`, `ViewData`, and `NewView` (`Stop`, `StopData`, and `Sync` in BFT-SMaRt).
+
+![alt text](https://raw.githubusercontent.com/wiki/SmartBFT-Go/consensus/images/viewChange.png)
+
+The `ViewChange` message (the first message) is sent to all nodes by some node who suspects the leader is faulty. This message is very lightweight and includes only the next view sequence number. 
+Once collecting `2f+1` `ViewChange` messages, a non faulty node is convinced that enough nodes consider the leader as faulty, and it will send a `ViewData` message to the next view's leader.
```
andrew-coleman: How do the non-faulty nodes agree on which node should be the next leader?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+
+**Input**: To submit requests to be ordered, the application invokes `SubmitRequest` and passes a byte representation of the request.
+````
+SubmitRequest(req []byte) error
+````
+
+The library offloads communication to the application that embeds it, hence it is the responsibility of the application to feed the messages from other nodes, or requests forwarded from other nodes into the library instance:
+
+````
+HandleMessage(sender uint64, m *Message)
```
andrew-coleman:
```suggestion
HandleMessage(sender uint64, m *bft.Message)
```

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+
+**Dependencies**: The library interacts with the external world through the application, by delegating capabilities such as cryptography, communication and storage to the application. 
+The full list of dependencies can be found [here](https://smartbft-go.github.io/godoc/pkg/github.com/SmartBFT-Go/consensus/pkg/api/index.html), and below the important ones are listed:
+
+The SmartBFT library delegates interaction with the external world (communication and storage) to the application that consumes it. It also doesn't come with its own signature scheme and also assumes its consumer
+implements it. All such abilities it delegates to the consumer application are termed "dependencies" and their full list can be found [here](https://smartbft-go.github.io/godoc/pkg/github.com/SmartBFT-Go/consensus/pkg/api/index.html).
+The dependencies can be categorized into two sets:
```
andrew-coleman: These two paragraphs say, more or less, the same thing.  Could they be combined?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+A Fabric block has to be signed by a quorum of consenters, and only by the specific consenters that were defined as consenters during the agreement on that block.
+When a channel will be created, the `configtxgen` tool will ensure that the policy is configured to be a Byzantine majority of the orderers defined the channel config.
+Upon a config change, the BFT orderer will ensure that the block verification policy is derived from the orderers [defined in the channel config](https://github.com/hyperledger/fabric-protos/blob/main/common/configuration.proto#L35-L49).  
+
+## Changes to the peer
+
+While a malicious orderer cannot forge a block, it can deprive a peer from blocks. Hence, receiving the stream of blocks from a single orderer exposes the peer to what is called a censorship attack, or a block withholding attack.
+
+A [previous RFC](https://github.com/hyperledger/fabric-rfcs/pull/48) explains at length how this censorship attack is mitigated. 
+In brief, if an orderer suspects it is being censored, it reaches to other orderers and pulls block headers and their corresponding signatures.
```
andrew-coleman:
```suggestion
In brief, if a peer suspects it is being censored, it reaches to other orderers and pulls block headers and their corresponding signatures.
```
?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+Optionally, once in a while, an orderer may also change the orderer from which it pulls blocks.
```
andrew-coleman:
```suggestion
Optionally, once in a while, a peer may also change the orderer from which it pulls blocks.
```

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+This approach is different than the approach taken in the SmartBFT Fabric fork, where the orderer constantly pulls headers and signatures from several orderers.
```
andrew-coleman:
```suggestion
This approach is different than the approach taken in the SmartBFT Fabric fork, where the peer constantly pulls headers and signatures from several orderers.
```

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,405 @@
...
+The deviation is in order to be computationally efficient, as in overwhelming cases, the orderers will not withhold blocks.
+
+
+## Changes to client SDK
+Since orderers can be malicious and thus either drop transactions or never include them in a block when assembling proposals, any client that interacts with the BFT ordering service needs to submit the transaction to all orderers.
+such a change is considered to be trivial and minimal and it is argued that it should be made in every SDK with ease.
```
andrew-coleman, line 302: This change also needs to be made to the gateway service.

text/006-bft-based-ordering-service.md
yacovm: In the classical BFT there is no "agreement" on the next leader, but it rotates to the next node in accordance to its numerical identifier. I added a clarification.

yacovm: Also, any BFT protocol that lets the nodes "elect" the leader in a non-randomized process would lead to an attack where an adversary that controls several byzantine nodes constantly makes its nodes be re-elected again as leaders. 
Unless you take some sort of ["black-listing" approach](https://patents.google.com/patent/US11343313B1/en), or a real [randomized approach](https://eprint.iacr.org/2020/025) (which is computationally expensive)

andrew-coleman (approved):

text/006-bft-based-ordering-service.md
sergefdrv: > any BFT protocol that lets the nodes "elect" the leader in a non-randomized process would lead to an attack where an adversary that controls several byzantine nodes constantly makes its nodes be re-elected again as leaders.

Could you please share a literature reference or elaborate a little more yourself to support this claim?

How rotation according to nodes' numerical identifier is different from the non-randomized leader selection process?

yacovm: >  Could you please share a literature reference or elaborate a little more yourself to support this claim?
> 
> How rotation according to nodes' numerical identifier is different from the non-randomized leader selection process?

So, take for example what happens with Raft. And I suppose this is also why Andrew asked this question.
In Raft, the leader election doesn't use rotation like in BFT, but the candidates send out requests to other candidates to vote for them, and eventually they gather a majority of votes. 
A Byzantine candidate may always act faster than everyone else and send quickly requests for votes from everyone else. 

In a (truly) randomized process (where the randomness is verifiable and not just randomized in the honest setting) we can think of the the leader election process as a deterministic function where the inputs are random and come from the candidates.

Param-S:
Reviewed this RFC. It looks fine to me.

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+
+Next each node waits for a quorum of `prepare` messages with the same digest, view `v`, and sequence number. 
+The quorum size should be big enough to ensure that in the intersection of any two quorums there is at least one correct node. So for example, if `n=3f+1` then the optimal quorum size is `2f+1`. 
+
+After the node is "prepared" it sends a `commit` message with the same content as the `prepare` message, as well as a signature on the proposal.
+Each node collects a quorum of `commit` messages and their attached signatures. 
+These signatures are later used to notarize the proposal, essentially creating a checkpoint. 
+Upon receiving a quorum of validated `commit` messages, the node delivers its proposal and the signatures to the application layer for storage and processing.
+
+The quorum size is calculated in the following manner: Let `n` be the total number of nodes in the algorithm and denote `f = ((n - 1) / 3)`. Then, the quorum is: `Ceil((n + f + 1) / 2)` where `Ceil` is the ceiling function. In layman's terms, the quorum is the minimal set of nodes such that every two such sets intersect in at least 1 honest node.
```
denyeart, line 130: An example of a typical deployment would help here. e.g. with 7 nodes, 2 can fail, with a quorum of 5.

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+
+**View change**:
+The view change protocol is inspired by the _Synchronization Phase_ in this [BFT-SMaRt paper](http://www.di.fc.ul.pt/~bessani/publications/edcc12-modsmart.pdf) (illustrated in the attached figure taken from the BFT-SMaRt paper).
+The view change protocol is comprised of three messages: `ViewChange`, `ViewData`, and `NewView` (`Stop`, `StopData`, and `Sync` in BFT-SMaRt).
+
+![alt text](https://raw.githubusercontent.com/wiki/SmartBFT-Go/consensus/images/viewChange.png)
+
+The `ViewChange` message (the first message) is sent to all nodes by some node who suspects the leader is faulty. This message is very lightweight and includes only the next view sequence number. 
+Once collecting `2f+1` `ViewChange` messages, a non faulty node is convinced that enough nodes consider the leader as faulty, and it will send a `ViewData` message to the next view's leader (in practice, it's the view number modulo the total number of nodes).
```
denyeart, line 145: Could you add some more detail here to answer these questions:
- How is the "next view's leader" chosen?
- Won't all the honest nodes potentially select a different "next view's leader" at the same time? Which one 'wins'?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+## Embedding the library in a distributed application
+
+ The [SmartBFT library](https://github.com/SmartBFT-Go/consensus) presentes a minimal interface to the application that uses it:
+
+**Input**: To submit requests to be ordered, the application invokes `SubmitRequest` and passes a byte representation of the request.
+````
+SubmitRequest(req []byte) error
+````
+
+The library offloads communication to the application that embeds it, hence it is the responsibility of the application to feed the messages from other nodes, or requests forwarded from other nodes into the library instance:
```
denyeart: I'm not following this sentence, I think because 'application' is unclear. At first I was thinking the 'client application' would be calling 'SubmitRequest' on the Fabric orderer node. But upon getting to this sentence I think you mean the orderer node is the application that consumes the library. Maybe it would help to set some context at the beginning of this section. Are the Fabric orderer node grpc APIs sitting above these APIs? And the orderer node is the application calling these APIs? Since this is a Fabric RFC, it would help to use language specific to Fabric orderer instead of discussing a generic 'application', and discuss the flow between Fabric orderer code and the library.

Further, what is the difference between SubmitRequest() and HandleRequest(). I guess SubmitRequest() would be ultimately (indirectly) called from client application to every node, while HandleRequest() would only be called in the timeout scenario where a follower doesn't see a request in a block and therefore forwards the request to the leader?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+Similarly to the existing ordering service, the new ordering service node will posses a chain object per channel.
+However, it will not use the consenter approach, but rather it would initialize its chains directly as a result of an invocation of the channel participation API. 
+The latter exposes a callback that is invoked when joining a channel, and that callback will be implemented by a component that will fill a role similar to the [registrar](https://github.com/hyperledger/fabric/blob/main/orderer/common/multichannel/registrar.go).
+
+
+
+## Implementing the dependencies required by the library
+
+In order to integrate the library, the orderer must implement all needed [dependencies](https://smartbft-go.github.io/godoc/pkg/github.com/SmartBFT-Go/consensus/pkg/api/index.html) required by the library.
+Below we enumerate how the dependencies were realized in the SmartBFT orderer:
```
denyeart, line 274: I'm getting confused by the context switching between 'old' orderer approach (v2.x), and the 'new' orderer approach (v3.x). The last section talked about the 'old' approach and then the 'new' approach. Is this section talking about the 'old' orderer approach. Or are these the same across the 'old' and 'new' approaches?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+
+From an operational perspective, the upgrade process from Raft to BFT will be as follows:
+
+1. The channel is put in [maintenance mode](https://hyperledger-fabric.readthedocs.io/en/release-2.4/kafka_raft_migration.html), to allow only administrative tasks and forbid regular transactions.
+
+2. A representative of the channel gathers signatures on a configuration update from sufficient organizations. The configuration update changes the [consensus type](https://github.com/hyperledger/fabric-protos/blob/main/orderer/configuration.proto#L20) to 'smartBFT', and sets the [metadata](https://github.com/hyperledger/fabric-protos/blob/main/orderer/configuration.proto#L22) to be empty.
+
+3. The configuration update is wrapped into a transaction and is sent to the ordering service.
+
+4. The Raft chain realizes the consensus type has changed, and the Raft chain goes into standby mode which only allows migration backward to BFT but nothing else.
```
denyeart: Don't you mean 'forward to BFT' rather than 'backward to BFT'?

text/006-bft-based-ordering-service.md
```diff
@@ -0,0 +1,404 @@
...
+
+# Possible questions and their answers
+
+-  **Q**: Why SmartBFT? There are more performant consensus protocols out there. SmartBFT is essentially a non pipelined PBFT, which isn't great for performance.  
+
+   **A**: By following the approaches explained in the previous [two](https://github.com/hyperledger/fabric-rfcs/pull/48), [RFCs ](https://github.com/hyperledger/fabric-rfcs/pull/50) in the series, integrating SmartBFT into Fabric does not block any other protocol to be integrated as well.  As a matter of fact, it only makes integration of other Byzantine Fault Tolerant libraries easier, as it will surely create re-usable code.
+
+-  **Q**: Who authored and maintains the SmartBFT library why should we trust it? 
+
+   **A**: The SmartBFT consensus library was authored by [a group of people](https://github.com/SmartBFT-Go/consensus/graphs/contributors) from IBM Research, Israel. All of them being contributors to Fabric core, and some of them being maintainers. In mid 2021 its maintenance was transferred to [Ideas](https://scientificideas.org), the Interdisciplinary Blockchain Research and Expertise center.
```
denyeart, line 395: Why was it 'transferred' to Ideas? Seems it would make sense to jointly maintain it between the original authors AND Ideas.

denyeart (commented):
Nice RFC as usual from @yacovm.
Just some minor questions... feel free to update the RFC itself to clarify, rather than answer them in github comments.

text/006-bft-based-ordering-service.md
yacovm: It says: 

> and it will send a `ViewData` message to the next view's leader (in practice, it's the view number modulo the total number of nodes).



So, it's the node with the identifier that is the result of the (next) view number % the number of nodes. 

> Won't all the honest nodes potentially select a different "next view's leader" at the same time? Which one 'wins'?

If there are some nodes in view `v` and some nodes in view `v'` that is different than `v` then either it's possible to collect enough votes for the next leader (be it `v` or `v'`) or it's impossible. 

If it's impossible, the nodes increase their timeout and view number and try again. Eventually, the timeouts are big enough such that they reach an agreement.

If it's possible, then a byzantine majority moves to the next view and the new leader takes over. The nodes that tried to view change to the incorrect view occasionally "probe" the rest of the nodes, and this is a chance for them to learn about the new successful view change.

text/006-bft-based-ordering-service.md
yacovm: This section talks about how the library is instantiated and integrated. 
Integrating the library is the same, of course. 
What differs is what we do in the Fabric side, and it's orthogonal to the library.

text/006-bft-based-ordering-service.md
yacovm: yes

text/006-bft-based-ordering-service.md
yacovm: I added clarifications.

text/006-bft-based-ordering-service.md
yacovm: Added an example

denyeart (approved):

denyeart:
With 3 maintainers in agreement let's proceed to Final Comment Period. I have sent a notification to Fabric mailing list to see if there are any final comments or concerns.

denyeart:
@yacovm @andrew-coleman No other comments, looks like we are good to merge this one, go ahead and create an Epic in fabric github and we'll start accepting PRs.

PR #53, opened 2022-10-14
New Capability for Updated IDEMIX

mbwhite:
Signed-off-by: Matthew B White <whitemat@uk.ibm.com>


yacovm:
Why do you need an RFC? The idemix library upgrade doesn't carry a functionality change.

PR #57, opened 2023-05-16
BFT signatures: update

tock-ibm:
The RFC currently calls for adding a gRPC service for getting the Header+Signature stream (Block attestation).

After reviewing the current (SmartBFT fabric v2.3) and proposed solution, it is clear that adding a gRPC service and multiple new proto messages needlessly duplicates code and bloats the code base. The task is realised in the current implementation by 6 lines of code - 3 in the orderer (set nil to the block payload, depending on the content type requested) and 3 in the peer (set the required content type).

We therefore revert this part of the proposal and will simply migrate the code from SmartBFT fabric v2.3 for this functionality, as described in the current RFC (see https://github.com/hyperledger/fabric-rfcs/blob/8ebea2f8fb54bdc6444229049e75eef98240cd0a/text/0010-bft-signatures.md?plain=1#L132).

### DIFF
--- a/text/0010-bft-signatures.md
+++ b/text/0010-bft-signatures.md
@@ -216,38 +216,49 @@ Therefore, Fabric nodes will pick the BFT implementation if either the block val
 Note that overriding the local configuration will only affect protection against block withholding attacks, and cannot impact policy validation.
 Therefore, the in the worst scenario of a misconfiguration, malicious ordereres still cannot forge blocks.
 
-Unlike the SmartBFT implementation, where the `SeekContentType` is expanded to accommodate sending back block attestations, we shall define a new gRPC service that will be serviced by orderers:
+### Block attestation
 
-````
-    // BlockAttestations receives an Envelope of type DELIVER_SEEK_INFO , then sends back a stream of BlockAttestations.
-    rpc BlockAttestations(Envelope) returns (stream BlockAttestationResponse);
-````
+We follow the approach of the SmartBFT implementation. The `SeekInfo` message is augmented with type `enum SeekContentType` and field `SeekContentType content_type = 5;`, as shown below: 
 
-
-
-Put simply, the consumer of the API sends to the service the same message as the block deliver API, but the messages
-received are `BlockAttestationResponse` messages.
-
-
-```
-message BlockAttestationResponse {
-    oneof Type {
-        common.Status status = 1;
-        BlockAttestation block_attestation = 2;
+```protobuf
+message SeekInfo {
+    enum SeekBehavior {
+        BLOCK_UNTIL_READY = 0;
+        FAIL_IF_NOT_READY = 1;
     }
+
+    enum SeekErrorResponse {
+        STRICT = 0;
+        BEST_EFFORT = 1;
+    }
+
+    // <NEW> SeekContentType indicates what type of content to deliver in response to a request. If BLOCK is specified,
+    // the orderer will stream blocks back to the peer. This is the default behavior. If HEADER_WITH_SIG is  specified, the
+    // orderer will stream only a the header and the signature(s), and the payload field will be set to nil. This allows
+    // the requester to ascertain that the respective signed block exists in the orderer (or cluster of orderers).
+    enum SeekContentType {
+        BLOCK = 0;
+        HEADER_WITH_SIG =1;
+    }
+    
+    SeekPosition start = 1;               // The position to start the deliver from
+    SeekPosition stop = 2;                // The position to stop the deliver
+    SeekBehavior behavior = 3;            // The behavior when a missing block is encountered
+    SeekErrorResponse error_response = 4; // How to respond to errors reported to the deliver service
+    SeekContentType content_type = 5;     // <NEW> Defines what type of content to deliver in response to a request
 }
 ```
 
-The `status` is the regular [common.Status](https://github.com/hyperledger/fabric-protos/blob/main/common/common.proto#L15-L25) and the `BlockAttestation` is a block without a block data, and is defined as follows:
-
-```
-message BlockAttestation {
-       BlockHeader header = 1;
-       BlockMetadata metadata = 2;
+When a stream of Header+Signatures is needed from the orderer, the peer (or any other client) will simply set the `ContentType` in the `SeekInfo` message to `HEADER_WITH_SIG`.
+In this case, the orderer will simply set the block data to `nil`.
+```go
+if seekInfo.ContentType == ab.SeekInfo_HEADER_WITH_SIG {
+    block.Data = nil
 }
 ```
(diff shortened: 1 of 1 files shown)


yacovm:
> The task is realised in the
> current implementation by 6 lines of code - 3 in the orderer (set nil to the block payload, depending on the
> content type requested) and 3 in the peer (set the required content type).
> 

You need to define the new content type no? Where in the change is this defined?

yacovm:
> We therefore revert this part of the proposal and will simply migrate the code from SmartBFT fabric v2.3 for this functionality.

Please, can we be more explicit? I understand what this means and you understand but not everyone does

tock-ibm:
> > We therefore revert this part of the proposal and will simply migrate the code from SmartBFT fabric v2.3 for this functionality.
> 
> Please, can we be more explicit? I understand what this means and you understand but not everyone does

The extension of the `SeekInfo` with content type is defined 2 paragraphs above in the RFC.... do you think we need more?

yacovm:
> > > We therefore revert this part of the proposal and will simply migrate the code from SmartBFT fabric v2.3 for this functionality.
> > 
> > 
> > Please, can we be more explicit? I understand what this means and you understand but not everyone does
> 
> The extension of the `SeekInfo` with content type is defined 2 paragraphs above in the RFC.... do you think we need more?

I think we need to at least explicitly state that we are going to use the same protobuf definitions, otherwise it's not clear. 
It only says "... is similar to the approach of SmartBFT" . 

Additionally I think it makes sense to put somewhere in the RFC a history of edits to it.

tock-ibm:
> > > > We therefore revert this part of the proposal and will simply migrate the code from SmartBFT fabric v2.3 for this functionality.
> > > 
> > > 
> > > Please, can we be more explicit? I understand what this means and you understand but not everyone does
> > 
> > 
> > The extension of the `SeekInfo` with content type is defined 2 paragraphs above in the RFC.... do you think we need more?
> 
> I think we need to at least explicitly state that we are going to use the same protobuf definitions, otherwise it's not clear. It only says "... is similar to the approach of SmartBFT" .
> 
> Additionally I think it makes sense to put somewhere in the RFC a history of edits to it.

Done.

C0rWin:
> multiple new proto messages needlessly duplicates code and bloats the code base.

Is this really so? And another question is, does it really make sense to couple block attestation requests with blocks delivery service? At the first revision, there is a clear segregation between how you can implement block attestation and how you retrieve blocks. For instance, if someone provides a more efficient attestation scheme or a way to prevent censorship by sending a succinct proof, you can change it without touching the delivery service. 

While I understand the intent, I am not sure the suggested change will indeed end up with less code. However, it will introduce coupling, which by the end of the day, the initial proposal was trying to avoid.

tock-ibm:
> > multiple new proto messages needlessly duplicates code and bloats the code base.
> 
> Is this really so? And another question is, does it really make sense to couple block attestation requests with blocks delivery service? At the first revision, there is a clear segregation between how you can implement block attestation and how you retrieve blocks. For instance, if someone provides a more efficient attestation scheme or a way to prevent censorship by sending a succinct proof, you can change it without touching the delivery service.
> 
> While I understand the intent, I am not sure the suggested change will indeed end up with less code. However, it will introduce coupling, which by the end of the day, the initial proposal was trying to avoid.

There will most definitely be less code. In the orderer, there are just 3 lines of code. In the peer, the code of the block delivery client is reused. The benefit of this solution is that it reuses the delivery client/server code and saves a lot of duplication. What is saved here (reused) is the code that sends requests, streams back blocks, receives blocks, etc. That whole part has to be recreated for a separate attestation service. 

The thing is, that currently a block header with signatures is the only proof we have in Fabric. If someone had proposed a new type of proof then we may have to reconsider, but realistically, there isn't right now. The only additional type of attestation I can think of right now is just a header without signatures. If you collect headers from enough orderers (>= f+1) you can reason about the block height as well. This scheme also aligns well with the proposal - a content type of HEADER_ONLY does the trick.

If there would ever be a need or an ability to supply the peer/client with proofs that are not already contained in the existing block structure, then it would be justified to add a separate service. Right now, in my opinion, it is a premature optimization.

C0rWin:
> > > multiple new proto messages needlessly duplicates code and bloats the code base.
> > 
> > 
> > Is this really so? And another question is, does it really make sense to couple block attestation requests with blocks delivery service? At the first revision, there is a clear segregation between how you can implement block attestation and how you retrieve blocks. For instance, if someone provides a more efficient attestation scheme or a way to prevent censorship by sending a succinct proof, you can change it without touching the delivery service.
> > While I understand the intent, I am not sure the suggested change will indeed end up with less code. However, it will introduce coupling, which by the end of the day, the initial proposal was trying to avoid.
> 
> There will most definitely be less code. In the orderer, there are just 3 lines of code. In the peer, the code of the block delivery client is reused. The benefit of this solution is that it reuses the delivery client/server code and saves a lot of duplication. What is saved here (reused) is the code that sends requests, streams back blocks, receives blocks, etc. That whole part has to be recreated for a separate attestation service.
>

Well, I think it still could be solved with just refactoring legacy to introduce better abstractions, which could be plugged into signature verification code.

> The thing is, that currently a block header with signatures is the only proof we have in Fabric. If someone had proposed a new type of proof then we may have to reconsider, but realistically, there isn't right now. The only additional type of attestation I can think of right now is just a header without signatures. If you collect headers from enough orderers (>= f+1) you can reason about the block height as well. This scheme also aligns well with the proposal - a content type of HEADER_ONLY does the trick.

I guess my only concern with suggested approach is that it takes the initial @yacovm proposal of having IoC based pluggable consensus component. Moreover, there is nothing wrong if you will have some duplication at price of having better segregation of different components.

yacovm:
> I guess my only concern with suggested approach is that it takes the initial @yacovm proposal of having IoC based pluggable consensus component.

I'm not sure I follow... IoC doesn't have an opinion about ABI.

C0rWin:
Ok, I guess the reflection of the suggested change is here? https://github.com/hyperledger/fabric/pull/4232/files Right?

tock-ibm:
> Ok, I guess the reflection of the suggested change is here? https://github.com/hyperledger/fabric/pull/4232/files Right?

Yes, this is a reflection of the requested change in the orderer. The part in the peer is similarly simple.

Note that I am not against adding an attestation service in general, If the attestations were carrying something that is different than just a block with nil data, then yes, it makes sense to add one. Also, if the attestations were to be communicated in a different fashion, i.e. different that the request<->stream gRPC style (of the delivery service), then also yes. Upon inspecting the current RFC proposal and  starting to implement it, one realizes that this is a duplication of code that doesn't really help anything. In particular, I don't think it contradicts the RFC that calls for inversion of control (IoC). Our goal is to make the delivery service and the "unified" block puller independent components that can be used when moving forward with the IoC plan.

C0rWin (approved):

PR #58, opened 2024-10-27
Peer to chaincode optimization proposal RFC

C0rWin:
- Added a new RFC proposing a batching mechanism to reduce communication overhead.
- Defined new `ChangeStateBatch` message type to encapsulate multiple state changes.
- Outlined the process flow, including negotiation and batch handling logic.
- Included sample chaincode illustrating the issue with current communication protocol.
- Provided benchmarks comparing performance before and after applying batching.
- Listed required repository changes in the correct order:
  - github.com/hyperledger/fabric-protos
  - github.com/hyperledger/fabric-chaincode-go
  - github.com/hyperledger/fabric
- Ensured backward compatibility by using a negotiation mechanism in `Ready` messages.
- Added example handler code for processing batched state changes.

### DIFF
--- a/text/0011-peer-chaincode-optimization.md
+++ b/text/0011-peer-chaincode-optimization.md
@@ -198,12 +198,20 @@ During the RFC discussion, the following concerns were raised about the proposed
 	•	Solution: Introduce explicit APIs for developers to enable and manage batching in their chaincode. This gives developers control over when and how batching is applied, allowing for easier debugging and incremental adoption.
 	•	Implementation: APIs like StartBatch and FinishBatch can provide clear demarcation points for batching operations, ensuring developers understand and control the batching process.
 
-4. Optimizing Interleaved Operations
-
-	•	Solution: Allow partial batching where consecutive writes are batched, and reads or other operations trigger immediate processing of pending batches. This minimizes communication overhead without significantly altering the behavior for interleaved read-write scenarios.
-	•	Implementation: The shim can dynamically manage batches and flush them when necessary, ensuring optimal performance while maintaining behavior consistency.
-
 ## Final Solution
+
+After careful consideration of the various approaches, option #3 - Explicit Developer Control - was selected as the final solution. This choice was made for several key reasons:
+
+1. Safety and Predictability: By giving developers explicit control over batching through dedicated APIs, we minimize the risk of unexpected behavior that could arise from implicit batching mechanisms. Developers can clearly understand and control when batching occurs.
+
+2. Debugging and Observability: The explicit nature of the batching APIs makes it easier to debug issues by providing clear entry and exit points for batched operations. This visibility is crucial for troubleshooting in production environments.
+
+3. Gradual Adoption Path: This approach allows for incremental adoption of the batching feature. Organizations can update their chaincodes at their own pace, testing and validating the performance benefits without forcing a system-wide change.
+
+4. Preservation of Existing Semantics: The explicit APIs ensure that the current chaincode behavior remains unchanged unless specifically opted into batching. This maintains compatibility while still offering performance improvements where desired.
+
+5. Simplified Implementation: Having clear batching boundaries reduces the complexity of managing state consistency and validation checks, as the system knows exactly when batching begins and ends.
+
 
 ### Extend Chaincode Stub API with Explicit Batching APIs
 
@@ -217,7 +225,7 @@ func (stub *ChaincodeStub) StartWriteBatch() {
 
 // FinishWriteBatch flushes the current batch and sends all collected write operations to the peer.
 // Returns a serialized set of results for each operation with the corresponding status.
-func (stub *ChaincodeStub) FinishWriteBatch() ([]BatchResult, error) {
+func (stub *ChaincodeStub) FinishWriteBatch() error {
     // Implementation to process the batched operations.
 }
 ```
@@ -247,17 +255,11 @@ func (t *OnlyPut) put(stub shim.ChaincodeStubInterface, args []string) *pb.Respo
         }
     }
     // Finish batching and send all operations to the peer
-    results, err := stub.FinishWriteBatch()
+    err := stub.FinishWriteBatch()
     if err != nil {
         return shim.Error(err.Error())
     }
 
-    // Process results if necessary
-    for _, result := range results {
-        if !result.Success {
-            return shim.Error(fmt.Sprintf("Operation failed for key: %s", result.Key))
-        }
-    }
     return shim.Success(nil)
 }
 ```
--- a/text/0011-peer-chaincode-optimization.md
+++ b/text/0011-peer-chaincode-optimization.md
@@ -1,0 +1,265 @@
+
+---
+layout: default
+title: Optimization of Communication Protocol between Hyperledger Fabric Endorsing Nodes and Chaincodes
+nav_order: 3
+---
+
+- Feature Name: Peer to Chaincode Communication Optimization
+- Start Date: (fill me in with today's date, 2022-03-01)
+- RFC PR: (leave this empty)
+- Fabric Component: peer, chaincode
+- Fabric Issue: (leave this empty)
+
+
+## Abstract
+This RFC proposes an enhancement to the communication protocol between Hyperledger Fabric endorsing nodes and chaincodes by introducing a batching mechanism. The current protocol requires each state modification command, such as `PutState` or `DelState`, to be sent individually, creating unnecessary overhead. This optimization minimizes the frequency and volume of communication between the chaincode and peer nodes, improving performance without changing the transaction semantics or requiring modification.
+
+---
+
+## Motivation
+In the current Hyperledger Fabric architecture, every state-changing operation triggers a distinct message from the chaincode to the peer node. This design causes several inefficiencies:
+
+1. **Communication Overhead:** The need to send individual messages for each state change operation increases the total number of messages transmitted, especially for workloads involving many updates. Each message exchange introduces latency and network overhead.
+
+2. **Performance Degradation:** Applications that perform multiple state modifications in loops experience high latency due to the large number of message exchanges. As a result, transaction throughput decreases, and system resources are inefficiently utilized.
+
+### Objectives of the Optimization
+The proposed batching mechanism aims to address these challenges by:
+- **Reducing Message Volume:** Grouping multiple state changes into a single `ChangeStateBatch` message reduces the number of messages exchanged between chaincode and peers.
+- **Improving Resource Utilization:** By transmitting fewer messages, the network load is reduced, and processing overhead on both the peer and chaincode is minimized.
+- **Ensuring Backward Compatibility:** Existing applications can continue to function as before without modification. If the batching feature is not supported by a component, the system will gracefully fall back to the original communication model.
+
+---
+
+## Problem Illustration with Sample Chaincode
+The following chaincode example demonstrates the inefficiencies caused by the current communication protocol. Each iteration of the loop sends an individual `PutState` message to the peer node, resulting in unnecessary overhead:
+
+```go
+type OnlyPut struct{}
+
+// Init initializes the chaincode.
+func (t *OnlyPut) Init(_ shim.ChaincodeStubInterface) *pb.Response {
+    return shim.Success(nil)
+}
+
+// Invoke - Entry point for invocations.
+func (t *OnlyPut) Invoke(stub shim.ChaincodeStubInterface) *pb.Response {
+    function, args := stub.GetFunctionAndParameters()
+    switch function {
+    case "invoke":
+        return t.put(stub, args)
+    default:
+        return shim.Error("Received unknown function invocation")
+    }
+}
+
+// Both params should be marshaled JSON data and base64 encoded.
+func (t *OnlyPut) put(stub shim.ChaincodeStubInterface, args []string) *pb.Response {
+    if len(args) != 1 {
+        return shim.Error("Incorrect number of arguments. Expecting 1")
+    }
+    num, _ := strconv.Atoi(args[0])
+    for i := 0; i < num; i++ {
+        key := "key" + strconv.Itoa(i)
+        if err := stub.PutState(key, []byte(key)); err != nil {
+            return shim.Error(err.Error())
+        }
+    }
+    return shim.Success(nil)
+}
+```
+
+---
+
+## Proposed Solution
+
+The proposed solution introduces a **batching mechanism** where multiple state changes are grouped into a single `ChangeStateBatch` message. This new message type will encapsulate all state changes and transmit them as a batch to the peer node. During the `Ready` message exchange, the peer and chaincode will negotiate the capability to use batching, ensuring backward compatibility.
+
+### New Message Format
+```proto
+message ChangeStateBatch {
+    repeated StateKV kvs = 1;
+}
+
+message StateKV {
+    enum Type {
+        UNDEFINED = 0;
+        PUT_STATE = 9;
+        DEL_STATE = 10;
+        PUT_STATE_METADATA = 21;
+        PURGE_PRIVATE_DATA = 23;
+    }
+
+    string key = 1;
+    bytes value = 2;
+    string collection = 3;
+    StateMetadata metadata = 4;
+    Type type = 5;
+}
+```
+
+---
+
+### Process Flow with ASCII Diagram
+The diagram below illustrates the message exchange between the chaincode and peer, showcasing the integration of the batching mechanism:
+
+```
++------------+                                            +-----------+
+| Chaincode  |                                            |    Peer   |
++------------+                                            +-----------+
+       |                                                        |
+       |--- Ready (UsePutStateBatch, MaxSizePutStateBatch) -->  |
+       |                                                        |
+       |<---- Acknowledgement Ready --------------------------  |
+       |                                                        |
+       |  Batch multiple commands                               |
+       |                                                        |
+       |--- ChangeStateBatch (Batch of Commands) -------------> |
+       |                                                        |
+       |<--- Acknowledgement (Success/Failure) ---------------- |
+       |                                                        |
+```
+
+---
+
+## Benchmark Results
+
+### Before Optimization
+```
+Only PUT State
+Name                                | N   | Min     | Median  | Mean    | StdDev  | Max    
+===========================================================================================
+invoke N-5 cycle-1 [duration]       | 5   | 62.9ms  | 64.2ms  | 437.8ms | 458.4ms | 999.9ms
+invoke N-200 cycle-1000 [duration]  | 200 | 297ms   | 649.5ms | 651ms   | 335.6ms | 1.0139s
+invoke N-200 cycle-10000 [duration] | 200 | 2.3774s | 2.4011s | 2.4037s | 16.8ms  | 2.5557s
+```
+
+### After Optimization
+```
+Only PUT State
+Name                                | N   | Min     | Median  | Mean    | StdDev  | Max    
+===========================================================================================
+invoke N-5 cycle-1 [duration]       | 5   | 62.2ms  | 62.7ms  | 437.9ms | 459.7ms | 1.0013s
+invoke N-200 cycle-1000 [duration]  | 200 | 71.8ms  | 541.3ms | 539.4ms | 427.6ms | 1.0085s
+invoke N-200 cycle-10000 [duration] | 200 | 176.2ms | 586.4ms | 588.9ms | 336ms   | 1.002s
+```
+The cycle of 10000 updates rounds clearly demonstrates the performance improvement of the batching mechanism.
+
+
+## Order of Repository Changes
+
+To implement the proposed batching mechanism, modifications need to be made across multiple repositories. Below is the order in which the repositories should be updated to ensure smooth integration:
+
+	1.	github.com/hyperledger/fabric-protos
+	    •	Define the new ChangeStateBatch message format.
+	    •	Ensure the protobuf schema includes all relevant message fields such as StateKV and Type.
+	2.	github.com/hyperledger/fabric-chaincode-go
+	    •	Update the chaincode framework to support batching logic.
+	    •	Add negotiation mechanisms for the Ready message exchange to determine if the batching feature is enabled.
+	    •	Modify internal functions to collect state modifications and prepare ChangeStateBatch messages for transmission.
+	3.	github.com/hyperledger/fabric
+	    •	Implement the handler logic to process the new ChangeStateBatch messages.
+	    •	Update the peer’s transaction context to support batched state changes.
+	    •	Ensure that any failure within a batch rolls back all changes, preserving consistency with existing transactional behavior.
+
+This sequence ensures that all dependencies are resolved correctly, avoiding integration issues when introducing the batching mechanism.
+
+---
+
+## Conclusion
+The proposed batching mechanism offers a simple yet effective way to optimize communication between Hyperledger Fabric chaincodes and peers. By reducing the number of messages exchanged, the system achieves better performance and resource utilization while maintaining transactional integrity and backward compatibility.
+
+## Amendment
+
+### Concerns Raised
+
+During the RFC discussion, the following concerns were raised about the proposed batching mechanism: maintaining the consistency of read-write order, managing existing read-write relationships, ensuring compatibility with current chaincodes, and addressing the impact on interleaved operations[1][2][3]. Altering the sequence of reads and writes during batching could lead to unexpected behavior or different error messages, particularly for applications that rely on the immediate execution of commands. Additionally, operations like SetState involve implicit reads or validations, such as metadata checks or collection name verification, which are currently handled interactively. Transitioning these operations to a batched model introduces complexity and risks compromising correctness. To mitigate these risks, a thorough review of the transaction simulator code is essential, as certain write operations, such as SetState, depend on reading the current state or metadata to function as expected.
+
+[1]: https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2458605718
+[2]: https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2460109295
+[3]: https://github.com/hyperledger/fabric-rfcs/pull/58#issuecomment-2473626315
+
+
+### Proposed Solutions
+
+1. Maintaining Read-Write Order
+
+	•	Solution: Ensure that the batching mechanism processes operations in the same order as they are invoked in the chaincode. This approach can be achieved by batching only consecutive write operations and sending read operations interactively.
+	•	Implementation: The chaincode shim can collect PutState or DelState operations into a batch, but any GetState operation will immediately flush the current batch and process the read interactively. This guarantees consistency with the current behavior.
+
+2. Handling Dependencies and Validation
+
+	•	Solution: Retain validation logic (e.g., metadata reads and collection name checks) on the peer side to ensure correctness. Avoid shifting these responsibilities to the shim, as it increases complexity and risk.
+	•	Implementation: The peer should process batched operations in a way that preserves existing checks and validations, ensuring no functional changes to current read-write dependencies.
+
+3. Explicit Developer Control
+
+	•	Solution: Introduce explicit APIs for developers to enable and manage batching in their chaincode. This gives developers control over when and how batching is applied, allowing for easier debugging and incremental adoption.
+	•	Implementation: APIs like StartBatch and FinishBatch can provide clear demarcation points for batching operations, ensuring developers understand and control the batching process.
+
+4. Optimizing Interleaved Operations
+
+	•	Solution: Allow partial batching where consecutive writes are batched, and reads or other operations trigger immediate processing of pending batches. This minimizes communication overhead without significantly altering the behavior for interleaved read-write scenarios.
+	•	Implementation: The shim can dynamically manage batches and flush them when necessary, ensuring optimal performance while maintaining behavior consistency.
+
+## Final Solution
+
+### Extend Chaincode Stub API with Explicit Batching APIs
+
+Introduce new APIs in the chaincode stub to explicitly control batching operations:
+
+```go
+// StartWriteBatch initializes a new batch for write operations.
+func (stub *ChaincodeStub) StartWriteBatch() {
+    // Implementation to begin batching write operations.
+}
+
+// FinishWriteBatch flushes the current batch and sends all collected write operations to the peer.
+// Returns a serialized set of results for each operation with the corresponding status.
+func (stub *ChaincodeStub) FinishWriteBatch() ([]BatchResult, error) {
+    // Implementation to process the batched operations.
+}
+```
+
+Advantages of Explicit Batching
+
+	1.	Developer Awareness: Developers can explicitly enable batching, ensuring they understand the impact on transaction behavior.
+	2.	Incremental Adoption: Existing chaincodes can operate without modification, while developers can selectively enable batching for specific chaincodes.
+	3.	Ease of Debugging: The explicit demarcation of batched operations simplifies debugging, as developers can correlate logs and behavior directly with batching commands.
+	4.	Backward Compatibility: Systems without batching support will continue to operate as before, ensuring seamless integration.
+
+### Example Usage
+
+```go
+func (t *OnlyPut) put(stub shim.ChaincodeStubInterface, args []string) *pb.Response {
+    if len(args) != 1 {
+        return shim.Error("Incorrect number of arguments. Expecting 1")
+    }
+    num, _ := strconv.Atoi(args[0])
+
+    // Start batching write operations
+    stub.StartWriteBatch()
+    for i := 0; i < num; i++ {
+        key := "key" + strconv.Itoa(i)
+        if err := stub.PutState(key, []byte(key)); err != nil {
+            return shim.Error(err.Error())
+        }
+    }
+    // Finish batching and send all operations to the peer
+    results, err := stub.FinishWriteBatch()
+    if err != nil {
+        return shim.Error(err.Error())
+    }
+
+    // Process results if necessary
+    for _, result := range results {
+        if !result.Success {
+            return shim.Error(fmt.Sprintf("Operation failed for key: %s", result.Key))
+        }
+    }
+    return shim.Success(nil)
+}
+```
+
+This approach balances performance optimization with consistency, developer control, and backward compatibility.


yacovm:
I think the reason that a series of `PutState`s is so slow is because the chaincode shim waits for a response from the peer, which makes no sense, as the peer doesn't really do anything smart with PutState commands, except from aggregating them in the write-set. 

Have you investigated the performance of a shim that simply defers all `PutState`s to the end of its execution, and sends them all one by one to the peer, without expecting an acknowledgement? I think this should be considered, as this probably will speedup things as well, and will not require a new message, and will probably be more backward compatible (a new shim would still work with an old peer, because it would not wait for the response, and can just probably ignore it)

C0rWin:
> I think the reason that a series of `PutState`s is so slow is because the chaincode shim waits for a response from the peer, which makes no sense, as the peer doesn't really do anything smart with PutState commands, except from aggregating them in the write-set.
> 
> Have you investigated the performance of a shim that simply defers all `PutState`s to the end of its execution, and sends them all one by one to the peer, without expecting an acknowledgement? I think this should be considered, as this probably will speedup things as well, and will not require a new message, and will probably be more backward compatible (a new shim would still work with an old peer, because it would not wait for the response, and can just probably ignore it)

Yes, this will achieve better results compared to the current implementation, while doing it with a single message is still better, given you avoid sending multiple messages.

pfi79:
> I think the reason that a series of `PutState`s is so slow is because the chaincode shim waits for a response from the peer, which makes no sense, as the peer doesn't really do anything smart with PutState commands, except from aggregating them in the write-set.
> 
> Have you investigated the performance of a shim that simply defers all `PutState`s to the end of its execution, and sends them all one by one to the peer, without expecting an acknowledgement? I think this should be considered, as this probably will speedup things as well, and will not require a new message, and will probably be more backward compatible (a new shim would still work with an old peer, because it would not wait for the response, and can just probably ignore it)

1. We tried sending everything at the end, but only with a waiting for a response. The result is the same as it is now.
2. In fabric-chaincode-go, [there is protection against parallel operation](https://github.com/hyperledger/fabric-chaincode-go/blob/main/shim/handler.go#L50-L55) (no waiting). Why this is done is not quite clear to me, so I am wary of getting into this mechanism.
3. Judging by the code, the prohibition of parallel requests in one transaction is necessary for parallel requests and distribution of responses from different transactions to work.

C0rWin:
After all, it is not very hard to implement RFC, which could bring significant gains without breaking current functionality and transaction semantics, not even the programming model.

denyeart:
There was an old Jira item to implement SetStateMultipleKeys() and GetStateMultipleKeys() to set and get keys in batch for precisely the reason you mention. It was even implemented at ledger level, just not at the chaincode programming model level. I don't think there was any reason it didn't get implemented, it simply wasn't prioritized for development.

But let's also check with @manish-sethi to see if there were any other considerations that delayed the implementation...

manish-sethi:
> 2. In fabric-chaincode-go, [there is protection against parallel operation](https://github.com/hyperledger/fabric-chaincode-go/blob/main/shim/handler.go#L50-L55) (no waiting). Why this is done is not quite clear to me, so I am wary of getting into this mechanism.

The main reason for this is to maintain the sanity of a transaction as it is intended at the application layer as a transaction is defined as a sequence of reads and writes. For instance, if a transaction sets different values to the same key, on different peers the result could differ if parallelism is to be allowed. Even if the result on different peers ends up the same, unless we prohibit setting the value of the same key more than once in a transaction, in the worst case, parallelism at shim could create a different result than the chaincode intended.

manish-sethi:
> There was an old Jira item to implement SetStateMultipleKeys() and GetStateMultipleKeys() to set and get keys in batch for precisely the reason you mention. It was even implemented at ledger level, just not at the chaincode programming model level. I don't think there was any reason it didn't get implemented, it simply wasn't prioritized for development.
> 
> But let's also check with @manish-sethi to see if there were any other considerations that delayed the implementation...

Yes, that was the intention. It was always a matter of priority. If we expose SetStateMultipleKeys() and GetStateMultipleKeys() to the chaincode, the simple cases should benefit easily.

text/0011-peer-chaincode-optimization.md
```diff
@@ -0,0 +1,171 @@
...
+        PUT_STATE_METADATA = 21;
+        PURGE_PRIVATE_DATA = 23;
+    }
+
+    string key = 1;
+    bytes value = 2;
+    string collection = 3;
+    StateMetadata metadata = 4;
+    Type type = 5;
+}
```
pfi79, line 99: Please note that this RFC suggests sending anything that changes state in batches.
I think it's cool.

C0rWin:
> There was an old Jira item to implement SetStateMultipleKeys() and GetStateMultipleKeys() to set and get keys in batch for precisely the reason you mention. It was even implemented at ledger level, just not at the chaincode programming model level. I don't think there was any reason it didn't get implemented, it simply wasn't prioritized for development.
> 
> But let's also check with @manish-sethi to see if there were any other considerations that delayed the implementation...

Well, we can reveal this line of work, this RFC precisely about that. Once we will get in agreement, it's we can just do it, I do think there is general consensus to proceed, given in the past there was even plans to do it not to mention ledger already supports it.

C0rWin:
@denyeart, what should the following steps be? Would you like this RFC to be presented at the community call? Would you like us to vote? Something else? Note, this is a straightforward and not intrusive change, not breaking any APIs while bringing impact.

denyeart:
@C0rWin 

I think the original vision for GetStateMultipleKeys() and SetStateMultipleKeys() was for the chaincode developer to be able to control getting and setting multiple keys at once. Whereas I believe your proposal would be transparent to the chaincode developer, that is, the updates would still be one at a time from the chaincode developer perspective, but at runtime they would be cached in the shim until the end of chaincode execution, at which point they would all be 'flushed' to the peer all at once right before the peer builds the final read/write set. Is my understanding correct?

I do agree the transparent batching makes more sense for ledger updates. For gets I still think the original intent makes sense, allow the chaincode developer to get N keys at once, so that they can use the values in the subsequent chaincode logic. However I think batch get is lower priority and not the intent of this RFC, so I'm ok with just doing the transparent batching for the updates as part of this RFC.

Question - do we need to consider splitting the batch into smaller pieces? For example if chaincode attempts to update 100000 keys should we send 10 batches of 10000 each? This may help to stay under limits such as grpc max message size (default 100MB), but on the other hand such large transactions would likely cause problems regardless due to the ultimate large write set. I'd prefer to keep things as simple as possible, just checking your thought.

In terms of next steps, RFCs require approval from 3 maintainers. For relatively straightforward RFCs like this one, we can just utilize github. If you think there are more considerations to discuss we could add an agenda item to the November 20th contributor meeting.

pfi79:
> Is my understanding correct?

Yeah, that's right.

pfi79:
> Question - do we need to consider splitting the batch into smaller pieces? For example if chaincode attempts to update 100000 keys should we send 10 batches of 10000 each? This may help to stay under limits such as grpc max message size (default 100MB), but on the other hand such large transactions would likely cause problems regardless due to the ultimate large write set. I'd prefer to keep things as simple as possible, just checking your thought.

This behavior will be configurable in core.yaml, via the new parameters UsePutStateBatch and MaxSizePutStateBatch.

The thing is, I know this will be an improvement in my case. But I don't know all the fabric use cases, so through the UsePutStateBatch parameter it will be possible to disable the new behavior. Or through the MaxSizePutStateBatch parameter - 1000000000 could be set so that the post is never split into batchs.

A benchmark test will also be added, which you can use as an example to research which parameters will be better in your particular case.

manish-sethi:
I agree with the overall proposal. However, if the goal is to keep the behavior consistent with the current implementation, I recommend an exercise of reviewing the [transaction simulator](https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go) code carefully, as some write operations involve reading the current state as well. For example, `SetState` causes a [read for the existing metadata](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L43). Private data write operations validate the existence of a collection name [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L105). Similarly, in the current code, certain write operations and read queries are mutually exclusive within a single transaction, as seen [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) and [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242).

In summary, we should assess whether it is feasible to maintain the current behavior in all cases, or alternatives should be discussed. A simple alternative could be to limit this optimization to straightforward cases or to allow for modifying the behavior so that any error is propagated at the end, rather than at the specific step where it occurs -- Also in some cases, we would have to limit ourselves to return a more general error as we would have lost the actual sequence of reads and writes.

pfi79:
> I agree with the overall proposal. However, if the goal is to keep the behavior consistent with the current implementation, I recommend an exercise of reviewing the [transaction simulator](https://github.com/hyperledger/fabric/blob/main/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go) code carefully, as some write operations involve reading the current state as well. For example, `SetState` causes a [read for the existing metadata](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L43). Private data write operations validate the existence of a collection name [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L105). Similarly, in the current code, certain write operations and read queries are mutually exclusive within a single transaction, as seen [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) and [here](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242).

That's an excellent point. 
We don't go that deep into change.
Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed
All those checks you're talking about are deeper, so there's no way to bypass them.

manish-sethi:
> That's an excellent point. We don't go that deep into change. Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed All those checks you're talking about are deeper, so there's no way to bypass them.

Yes, and hence it changes the error handling behavior from the current implementation. For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query. So I suggested going through the lower level code and capturing all behavior changes, and necessary changes in the error handling at the lower level that may be needed. IMO, this should be part of the RFC.

pfi79:
> this error

Doesn't it work both ways?

it seems to me that operations that cannot be performed in one transaction cause an error without necessarily following the order. 
That is, it doesn't matter whether “query” or “set” is executed first.

pfi79:
> > That's an excellent point. We don't go that deep into change. Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed All those checks you're talking about are deeper, so there's no way to bypass them.
> 
> Yes, and hence it changes the error handling behavior from the current implementation. For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query. So I suggested going through the lower level code and capturing all behavior changes, and necessary changes in the error handling at the lower level that may be needed. IMO, this should be part of the RFC.

Doesn't it work both ways?

it seems to me that operations that cannot be performed in one transaction cause an error without necessarily following the order.
That is, it doesn't matter whether “query” or “set” is executed first.

manish-sethi:
> > > That's an excellent point. We don't go that deep into change. Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed All those checks you're talking about are deeper, so there's no way to bypass them.
> > 
> > 
> > Yes, and hence it changes the error handling behavior from the current implementation. For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query. So I suggested going through the lower level code and capturing all behavior changes, and necessary changes in the error handling at the lower level that may be needed. IMO, this should be part of the RFC.
> 
> Doesn't it work both ways?
> 
> it seems to me that operations that cannot be performed in one transaction cause an error without necessarily following the order. That is, it doesn't matter whether “query” or “set” is executed first.

Yes, it does, and in the current implementation, the appropriate error is returned. If you make changes in the shim side only, the wrong error will be returned. In that case, it will be more appropriate to return a more general error than a wrong error.

C0rWin:
> I do agree the transparent batching makes more sense for ledger updates. For gets I still think the original intent makes sense, allow the chaincode developer to get N keys at once, so that they can use the values in the subsequent chaincode logic. However I think batch get is lower priority and not the intent of this RFC, so I'm ok with just doing the transparent batching for the updates as part of this RFC.

You got that right, the main intent was to have batched update operations transparent and hence the suggested modifications the least intrusive way to get to the point. 

> Question - do we need to consider splitting the batch into smaller pieces? For example if chaincode attempts to update 100000 keys should we send 10 batches of 10000 each? This may help to stay under limits such as grpc max message size (default 100MB), but on the other hand such large transactions would likely cause problems regardless due to the ultimate large write set. I'd prefer to keep things as simple as possible, just checking your thought.

Yes, we need to split it since we cannot send unlimited proto messages hence the batch has to respect the limitation. Honestly, I also do not think this is too difficult to implement.

C0rWin:
> > > > That's an excellent point. We don't go that deep into change. Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed All those checks you're talking about are deeper, so there's no way to bypass them.
> > > 
> > > 
> > > Yes, and hence it changes the error handling behavior from the current implementation. For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query. So I suggested going through the lower level code and capturing all behavior changes, and necessary changes in the error handling at the lower level that may be needed. IMO, this should be part of the RFC.
> > 
> > 
> > Doesn't it work both ways?
> > it seems to me that operations that cannot be performed in one transaction cause an error without necessarily following the order. That is, it doesn't matter whether “query” or “set” is executed first.
> 
> Yes, it does, and in the current implementation, the appropriate error is returned. If you make changes in the shim side only, the wrong error will be returned. In that case, it will be more appropriate to return a more general error than a wrong error.

I have to admit I am not familiar to the very intimate level with this part of the code, I do agree @manish-sethi has a good point, and therefore what has to be done we need to make sure and list all such a cases. 

That being said, since the proposal suggests keeping state updates within the shim, validation of this sort could be completed at the shim side either, preserving error handling semantics.

C0rWin:
@denyeart what would be the next steps?

denyeart:
> @denyeart what would be the next steps?

The feedback from Manish certainly complicates things. The authors of the RFC need to incorporate the feedback and propose a plan for how it will be addressed.

Alternatively the RFC could be scaled back, for example instead of saving all writes for the end, the chaincode could send bursts of updates to the peer. Potentially give the chaincode developer the ability to control the batching by calling an operation that sets multiple keys at a time (as was envisioned years ago with SetStateMultipleKeys). This would keep things most simple.

Or as a compromise, perhaps the chaincode shim itself could manage the batches. As an example, if the chaincode does one GetState(), then 100 PutState(), and a final GetState(), the chaincode could send to the peer in 3 operations:

- first GetState(), sent interactively when it is hit in the chaincode
- the 100 PutStates(), collected on shim side and then sent as a batch upon hitting the next non-PutState() operation.
- the final GetState(), sent interactively when it is hit in the chaincode

I would be in favor of one of these partial batching approaches. This would allow for keeping the checks and error message generation on the peer side, which will reduce the scope, complexity, and risk of the change. It will also help with consistency across various chaincode shim language implementations. Especially now that the original developers aren't as active, we need to manage the scope and risk of such changes.

I'm not sure how such counter-proposals will impact the RFC author's use cases however. The RFC authors will therefore need to synthesize Manish and my responses, together with their use case requirements, and come back with an updated proposal.

pfi79:
> > > > That's an excellent point. We don't go that deep into change. Only the [handler.go](https://github.com/hyperledger/fabric/blob/main/core/chaincode/handler.go) file is changed All those checks you're talking about are deeper, so there's no way to bypass them.
> > > 
> > > 
> > > Yes, and hence it changes the error handling behavior from the current implementation. For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query. So I suggested going through the lower level code and capturing all behavior changes, and necessary changes in the error handling at the lower level that may be needed. IMO, this should be part of the RFC.
> > 
> > 
> > Doesn't it work both ways?
> > it seems to me that operations that cannot be performed in one transaction cause an error without necessarily following the order. That is, it doesn't matter whether “query” or “set” is executed first.
> 
> Yes, it does, and in the current implementation, the appropriate error is returned. If you make changes in the shim side only, the wrong error will be returned. In that case, it will be more appropriate to return a more general error than a wrong error.

I don't know what you're talking about yet. Can you give me a specific case.

Like previously sent this, and so received this message.
Now you suggest and it will be different and that's the problem.

I'll write a corresponding test and think about it.

I really don't understand what you're writing about.

pfi79:
I can honestly say that we already have the code ready and for our use cases we have got a good acceleration.

I will put here links to commits of the changes we want to propose to make, after the approval of pr in rfcs.

https://github.com/pfi79/fabric-protos/commit/1c27d38c844c2a66312c10844190bedd5765b06d
https://github.com/pfi79/fabric-chaincode-go/commit/8954ab1d18a9b42fdaf5685c2e76b11f51c4cf49
https://github.com/pfi79/fabric/commit/43e186d521c1e281a355ee8a9875a870b1fb8487

denyeart:
@pfi79 Basically what Manish is saying is that if you change the order of the reads and writes, you may get different behavior or a different error message. To address this, you could shift all these checks and error messages from peer side to shim side. However, that would be a lot of work and we'd likely miss something. I would prefer to keep the order of reads and writes the same as it is currently (the order as seen in the chaincode), so that the behavior remains the same. What I'm suggesting is that you could still batch consecutive writes, so that the order of reads and writes remains the same, while still getting the performance improvements of batching. Of course, the chaincode would have to be written intelligently to do consecutive writes rather than read-write-read-write.

pfi79:
> @pfi79 Basically what Manish is saying is that if you change the order of the reads and writes, you may get different behavior or a different error message. To address this, you could shift all these checks and error messages from peer side to shim side. However, that would be a lot of work and we'd likely miss something. I would prefer to keep the order of reads and writes the same as it is currently (the order as seen in the chaincode), so that the behavior remains the same. What I'm suggesting is that you could still batch consecutive writes, so that the order of reads and writes remains the same, while still getting the performance improvements of batching. Of course, the chaincode would have to be written intelligently to do consecutive writes rather than read-write-read-write.

Let's say Manish is right. 
Can you or he provide a concrete example of this violation? 
I've examined the existing code and I can't think of an example that violates as Manish says.
Most likely I haven't foreseen something and I don't know something. That's why I'm asking about an example of such a violation.

pfi79:
By the way, above I provided commites of my code variant. But the situation Manish is talking about was not revealed.
So we don't have tests for it. I am ready to write such a test that would clearly show the importance of execution order and check it on my changes.
Or is there no situation, or is it not so important?

pfi79:
> For example, if you make changes in only in the shim side of code, a user will get [this error](https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L242) whereas in his chaincode he may have performed a write before the query.

Sorry Manish, I don't understand your example.
Could you please describe it in more detail?
Then I will write a corresponding test, which we will take apart and see what is wrong.

denyeart:
Here is one example that is mentioned in documentation:
https://hyperledger-fabric.readthedocs.io/en/latest/private-data-arch.html#querying-private-data

>Chaincode that executes key range queries or JSON queries and updates data in a single transaction is not supported, as the query results cannot be validated on the peers that don’t have access to the private data.

If chaincode does write then query private data you will get an error based on this check:
https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L234

If we hold all writes until the end, you won't get this error. You will likely still get an error based on the other checks, but it would be a different error that doesn't match the developer's view of the chaincode context.

pfi79:
> Here is one example that is mentioned in documentation: https://hyperledger-fabric.readthedocs.io/en/latest/private-data-arch.html#querying-private-data
> 
> > Chaincode that executes key range queries or JSON queries and updates data in a single transaction is not supported, as the query results cannot be validated on the peers that don’t have access to the private data.
> 
> If chaincode does write then query private data you will get an error based on this check: https://github.com/hyperledger/fabric/blob/6c2e87c0824604c7cd7a8f4f93b2c5876af76875/core/ledger/kvledger/txmgmt/txmgr/tx_simulator.go#L234
> 
> If we hold all writes until the end, you won't get this error. You will likely still get an error based on the other checks, but it would be a different error that doesn't match the developer's view of the chaincode context.

@denyeart Thanks for the example. The same can still be presented, for example: GetStateWithPagination and PutState.

1. A transaction that should be executed will be executed. A transaction that should not be executed will not be executed. I believe this is the most important thing.
2. The above problem can be solved in this way: the fact that the order of calculations can be changed, and the developers of chaincodes may see other errors than the expected ones, should be described in the documentation. With an explanation of why and how to think about the execution of the chaincode in a new way.
3. I think that the chaincode developer writes the code so that transactions are executed, not terminated with an error. The presented error variants are caught in the first step of the chaincode development. After that they never appear again.

C0rWin:
> Or as a compromise, perhaps the chaincode shim itself could manage the batches. As an example, if the chaincode does one GetState(), then 100 PutState(), and a final GetState(), the chaincode could send to the peer in 3 operations:
> 
> * first GetState(), sent interactively when it is hit in the chaincode
> * the 100 PutStates(), collected on shim side and then sent as a batch upon hitting the next non-PutState() operation.
> * the final GetState(), sent interactively when it is hit in the chaincode
> 
> I would be in favor of one of these partial batching approaches. This would allow for keeping the checks and error message generation on the peer side, which will reduce the scope, complexity, and risk of the change. It will also help with consistency across various chaincode shim language implementations. Especially now that the original developers aren't as active, we need to manage the scope and risk of such changes.

OK, I think that a sane alternative might be to piggyback batches of updates (essentially write operations) with the get/read operations, basically not to wait until the end but rather send all writes with the get command, which anyway has to communicate with the peer. 

This obviously has a drawback in cases where you have a lot of interleaved reads and writes, where you will lose the benefit of the batching.


Another possible solution might be to have a set of explicit APIs something like:

```
stub.StartBatch()

Put/Get

results := stub.FinishBatch() // returns a serialized set of results for each Put/Get operation with the corresponding status
```

Of course, this is a little bit more intrusive approach, but on the other side, it introduces a clear separation between the two approaches and doesn't assume anything implicitly, which might be overall good for the end-user experience. 

@denyeart @manish-sethi thoughts?

C0rWin:
> 1. A transaction that should be executed will be executed. A transaction that should not be executed will not be executed. I believe this is the most important thing.

While this is indeed an important thing, we should try to avoid bringing breaking changes with new versions or at least avoid making people change a bunch of legacy code just because we decided to optimize something. Refactoring legacy code to accommodate new errors and programming models might nullify all the benefits of batching.

pfi79:
> While this is indeed an important thing, we should try to avoid bringing breaking changes with new versions or at least avoid making people change a bunch of legacy code just because we decided to optimize something. Refactoring legacy code to accommodate new errors and programming models might nullify all the benefits of batching.

I'm sorry, but I disagree.

I'm not sure there are breaking changes here. So far all the examples have been with chaincode transactions that don't work, that is, that return errors always.

That is, I have to imagine that now somewhere in production there is a chaincode installed and it has a transaction that calls a contradictory method, i.e. always ends with an error.

The only place it can be is in chaincode tests.

Can you show an example of a working transaction where this would be important?

pfi79:
> OK, I think that a sane alternative might be to piggyback batches of updates (essentially write operations) with the get/read operations, basically not to wait until the end but rather send all writes with the get command, which anyway has to communicate with the peer.

That could be a problem right there.

The user executes GetState and receives an error on the PutState operation, which flew away with the Read operation.

pfi79:
You can add our variant but leave the old way of working in the settings. Then we won't touch the inherited code. 

Others will switch to this mode consciously via core.yaml

pfi79:
@denyeart @manish-sethi 
Colleagues, 
Let's say we have legacy software, to which order is important and nobody will rewrite it.

Now the proposal is made in such a way that it automatically enables new work when new components are encountered.

I want to support @C0rWin suggestion, it is necessary to switch on the new work mode not automatically, but manually by calling the stub.StartBatch() (or BatchBegin) method. This mode will be switched off either by calling the stub.FinishBatch() (or BatchEnd) method. Or at the end of transaction execution.

The only observation to the suggestion is that yes - we send in batches, but we apply the values one by one. Therefore, results := stub.FinishBatch() should return only one error. 

Then the legacy software will be safe until the new mode is enabled in the chaincodes by changing the code.

Would such a change to the proposal suit you?

denyeart:
@pfi79 I would be in favor of explicit batching from a chaincode developer command. This way if there are any issues developer can easily enable/disable the batching at the chaincode level and compare the results to determine if an issue is related to batching, without needing to update the peer configuration.

It would also make sure the developer understands the update model. For example if we had automatic batching, if developer is watching the peer logs they may be confused to see all the reads executing prior to the writes. If they explicitly add the batch command they will understand that the writes are batched and processed at the end.

Finally, for existing users it would also allow for incremental transition to the new approach. For any chaincode that writes many keys they may want to enable the explicit batching. For other chaincodes they could leave it as-is without batching enabled.

Since the batching is only related to writes, maybe the command should be called `stub.StartWriteBatch()`?

denyeart:
@pfi79 @C0rWin @manish-sethi There is a Fabric contributor meeting scheduled for tomorrow November 20th. We could discuss during the meeting if you are able to attend.

C0rWin:
@denyeart, @manish-sethi, @satota2  guys, updated RFC as per our conversation during the community call. Please review.

text/0011-peer-chaincode-optimization.md
```diff
@@ -0,0 +1,265 @@
...
+
+```go
+// StartWriteBatch initializes a new batch for write operations.
+func (stub *ChaincodeStub) StartWriteBatch() {
+    // Implementation to begin batching write operations.
+}
+
+// FinishWriteBatch flushes the current batch and sends all collected write operations to the peer.
+// Returns a serialized set of results for each operation with the corresponding status.
+func (stub *ChaincodeStub) FinishWriteBatch() ([]BatchResult, error) {
```
pfi79: Sorry, but I'm confused by []BatchResult.
What will be written there? 
Usually a change operation returns an error or nil. 
Batch will be applied on peer sequentially. And if at least one operation will end with an error, then immediately the processing of the whole batch will end with an error. 
Therefore, either []BatchResult will consist of all nil, or some number of nil and one error.

Wouldn't it be better to replace []BatchResult with error?

C0rWin: That was only a suggestion, I think you have a point and having only an error should be indeed enough of indication, given that implementation wise we won't continue to try and execute batch once we stumbled into the error.

C0rWin:
@denyeart can we move forward to the voting and last call stage?

text/0011-peer-chaincode-optimization.md
denyeart: I agree a simple error will suffice, returned when the first error is hit. Please update the PR accordingly to ensure we are all on the same page. The sample code snippet will also need to be updated.

text/0011-peer-chaincode-optimization.md
```diff
@@ -0,0 +1,265 @@
...
+
+4. Optimizing Interleaved Operations
+
+	•	Solution: Allow partial batching where consecutive writes are batched, and reads or other operations trigger immediate processing of pending batches. This minimizes communication overhead without significantly altering the behavior for interleaved read-write scenarios.
+	•	Implementation: The shim can dynamically manage batches and flush them when necessary, ensuring optimal performance while maintaining behavior consistency.
+
+## Final Solution
+
+### Extend Chaincode Stub API with Explicit Batching APIs
+
```
denyeart, line 217: Clarify here that Option 3 above has been chosen.

text/0011-peer-chaincode-optimization.md
```diff
@@ -0,0 +1,265 @@
...
+3. Explicit Developer Control
+
+	•	Solution: Introduce explicit APIs for developers to enable and manage batching in their chaincode. This gives developers control over when and how batching is applied, allowing for easier debugging and incremental adoption.
+	•	Implementation: APIs like StartBatch and FinishBatch can provide clear demarcation points for batching operations, ensuring developers understand and control the batching process.
+
+4. Optimizing Interleaved Operations
+
+	•	Solution: Allow partial batching where consecutive writes are batched, and reads or other operations trigger immediate processing of pending batches. This minimizes communication overhead without significantly altering the behavior for interleaved read-write scenarios.
+	•	Implementation: The shim can dynamically manage batches and flush them when necessary, ensuring optimal performance while maintaining behavior consistency.
+
```
denyeart, line 214: I don't understand the difference between option 1 and 4. Not a big deal though, since we are moving forward with option 3.

denyeart:
> @denyeart can we move forward to the voting and last call stage?

Please address the final comments above in the PR. Then each of us can indicate approval. Then I will follow the process for Final Comment Period, which entails sending out on mailing list and waiting one week for any final comments before merging. After that we can accept PRs in fabric repository.

bestbeforetoday:
Do you have any plans to update the [Node](https://github.com/hyperledger/fabric-chaincode-node) and [Java](https://github.com/hyperledger/fabric-chaincode-java) chaincode packages to keep all the chaincode implementations consistent?

C0rWin:
> Do you have any plans to update the [Node](https://github.com/hyperledger/fabric-chaincode-node) and [Java](https://github.com/hyperledger/fabric-chaincode-java) chaincode packages to keep all the chaincode implementations consistent?

I think we have to.

text/0011-peer-chaincode-optimization.md
C0rWin: removed to avoid confusion

C0rWin:
> > @denyeart can we move forward to the voting and last call stage?
> 
> Please address the final comments above in the PR. Then each of us can indicate approval. Then I will follow the process for Final Comment Period, which entails sending out on mailing list and waiting one week for any final comments before merging. After that we can accept PRs in fabric repository.

Done.

pfi79 (approved):
Thank you. I'm fine with it.

pfi79:
> > @denyeart can we move forward to the voting and last call stage?
> 
> Please address the final comments above in the PR. Then each of us can indicate approval. Then I will follow the process for Final Comment Period, which entails sending out on mailing list and waiting one week for any final comments before merging. After that we can accept PRs in fabric repository.

How long do you have to wait to go to the **Final Comment Period**?

text/0011-peer-chaincode-optimization.md
```diff
@@ -0,0 +1,267 @@
...
+Introduce new APIs in the chaincode stub to explicitly control batching operations:
+
+```go
+// StartWriteBatch initializes a new batch for write operations.
+func (stub *ChaincodeStub) StartWriteBatch() {
+    // Implementation to begin batching write operations.
+}
+
+// FinishWriteBatch flushes the current batch and sends all collected write operations to the peer.
+// Returns a serialized set of results for each operation with the corresponding status.
```
denyeart, line 227: Need to update this comment to reflect the final design - a single error will be returned.

denyeart (approved):
Just one more nit in a function comment, but I'll go ahead and approve and send out for Final Comment Period now.

denyeart:
Final Comment Period announcement - https://lists.lfdecentralizedtrust.org/g/fabric/message/12422

satota2 (approved):
LGTM!

pfi79:
> Final Comment Period announcement - https://lists.lfdecentralizedtrust.org/g/fabric/message/12422

7 calendar days have passed. Any other edits needed in the rfc?

denyeart:
No changes from Final Comment Period, merging!

pfi79:
@denyeart Can you ask you to create a epic in fabric, I will attach pr on this topic there

denyeart:
> @denyeart Can you ask you to create a epic in fabric, I will attach pr on this topic there

@pfi79 We used to manage Epics in github by using the zenhub plugin. But we have stopped using zenhub. We could still create an issue with Epic in the title and link to other issues or have a task list in the issue. Since you will likely be the person managing the epic issue I'll defer to your preference. Go ahead and open the epic issue however you prefer to manage it.

PR #59, opened 2024-10-30
Update link to the issues tracker

C0rWin:
Since we no longer use JIRA, updated link to reflect the actual issue tracker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ a comment describing the rationale for the decision.
 Some accepted RFCs represent vital changes that need to be implemented right
 away. Other accepted RFCs can represent changes that can wait until a
 developer feels like doing the work. Every accepted RFC has an associated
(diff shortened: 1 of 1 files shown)


satota2 (commented):
@C0rWin 
Thank you for the revisions.
I noticed that on line 211, the word "JIRA" still appears.
I believe it would be best to update "Use JIRA" to something like "Use GitHub Issues" as well.

pfi79 (approved):

C0rWin:
> @C0rWin Thank you for the revisions. I noticed that on line 211, the word "JIRA" still appears. I believe it would be best to update "Use JIRA" to something like "Use GitHub Issues" as well.

thanks for point this out, I have pushed an update in separate PR: https://github.com/hyperledger/fabric-rfcs/pull/60, cause wasn't able to alter current one, since branch somewhat was marked as protected.

PR #60, opened 2024-10-31
Update link to the issues tracker

C0rWin:
Since we no longer use JIRA, we updated the link to reflect the actual issue tracker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,11 +204,11 @@ a comment describing the rationale for the decision.
 Some accepted RFCs represent vital changes that need to be implemented right
 away. Other accepted RFCs can represent changes that can wait until a
 developer feels like doing the work. Every accepted RFC has an associated
-issue tracking its implementation in the [Fabric JIRA issue tracker](https://jira.hyperledger.org/projects/FAB/issues).
+issue tracking its implementation in the [Fabric issues](github.com/hyperledger/fabric/issues).
 
 The author of an RFC is not obligated to implement it. Of course, the RFC
 author, as any other developer, is welcome to post an implementation for review
-after the RFC has been accepted. Use JIRA for this.
+after the RFC has been accepted. Use GitHub issues for this.
 
 ## License
 


C0rWin:
created this one to reflects comments from #59, wasn't able to update the original PR

C0rWin:
@satota2 can you please take a look now?

README.md
```diff
@@ -204,11 +204,11 @@ a comment describing the rationale for the decision.
 Some accepted RFCs represent vital changes that need to be implemented right
 away. Other accepted RFCs can represent changes that can wait until a
 developer feels like doing the work. Every accepted RFC has an associated
-issue tracking its implementation in the [Fabric JIRA issue tracker](https://jira.hyperledger.org/projects/FAB/issues).
+issue tracking its implementation in the [Fabric issues](github.com/hyperledger/fabric/issues).
```
satota2, line 207:
```suggestion
issue tracking its implementation in the [Fabric issues](https://github.com/hyperledger/fabric/issues).
```

satota2 (changes requested):
@C0rWin
Thank you for the update. It appears that the current link directs to an incorrect location under fabric-rfcs.
It should instead point to the correct address with the `https://` prefix.

Other than that, LGTM!

C0rWin:
> @C0rWin Thank you for the update. It appears that the current link directs to an incorrect location under fabric-rfcs. It should instead point to the correct address with the `https://` prefix.
> 
> Other than that, LGTM!

I am not sure why, but I cannot update the current branch as it is marked as protected... 

@satota2, likely to push separate PR again :(

satota2:
> I am not sure why, but I cannot update the current branch as it is marked as protected...

@C0rWin
I don’t have the permissions to check the Branch Protection Rules in fabric-rfcs, but it seems that protections may be applied to all branches, not just main.
Creating the PR from a forked repository, like `C0rWin:peer2chaincodeComm`, might work better (though it does mean creating another PR).

@denyeart  Could you please check and review the Branch Protection Rules? Protecting just the main branch should be sufficient.

denyeart:
> > I am not sure why, but I cannot update the current branch as it is marked as protected...
> 
> @C0rWin I don’t have the permissions to check the Branch Protection Rules in fabric-rfcs, but it seems that protections may be applied to all branches, not just main. Creating the PR from a forked repository, like `C0rWin:peer2chaincodeComm`, might work better (though it does mean creating another PR).
> 
> @denyeart Could you please check and review the Branch Protection Rules? Protecting just the main branch should be sufficient.

We don't allow force pushes in any of the repositories, don't want to risk changing the history and don't want to maintain branch protection rules for multiple branches. I always fork and PR from my own fork, I thought that was a pretty standard approach?

satota2:
> We don't allow force pushes in any of the repositories, don't want to risk changing the history and don't want to maintain branch protection rules for multiple branches. I always fork and PR from my own fork, I thought that was a pretty standard approach?

@denyeart 
Thank you very much for your feedback. After looking into the branches in the Fabric core repository, I noticed that not all branches are protected there, so I thought there might be room to consider a similar approach if prioritizing flexibility was a goal.

That said, I completely agree with you that “I always fork and PR from my own fork” is indeed the standard approach, and I fully understand the risks involved in modifying the branch protection rules. Based on this, I’ll withdraw my suggestion.

@C0rWin
Could I kindly ask you to create a new PR with the necessary fixes?

PR #62, opened 2025-04-23
Add rfc for the private event feature

yeasy:
The rfc describes the new feature called private event.


0000-private-event.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
+
+The peer recognizes this as a private event and tracks it for notification purposes.
+
+## Listening to Private Events
+
+The private events are restricted and stored in the similar way as the private data.
+
+Only peers in organizations and CNs listed in the event's `visibleTo` map will process and relay the event to their REST Proxy, which will forward it to eligible subscribers.
```
C0rWin: What is this REST Proxy?

0000-private-event.md
```diff
@@ -0,0 +1,146 @@
...
+## Concept: Private Chaincode Events
+
+Private Chaincode Events are triggered by invoking the standard `PutPrivateData()` method in chaincode, with a specially prefixed key (e.g., `__hlf_privateEvent__eventName`).
+
+Each private event must conform to a strict structure:
+
+```json
+{
+  "visibleTo": {
+    "Org1MSP": ["user1"],
```
C0rWin, line 56: what is this user1, user2 and user3? Are these users' identities? Where did it come from, and how is it validated?

0000-private-event.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Concept: Private Chaincode Events
+
+Private Chaincode Events are triggered by invoking the standard `PutPrivateData()` method in chaincode, with a specially prefixed key (e.g., `__hlf_privateEvent__eventName`).
+
+Each private event must conform to a strict structure:
+
+```json
+{
+  "visibleTo": {
```
C0rWin, line 55: Why can't you reuse the existing policy mechanism?

0000-private-event.md
```diff
@@ -0,0 +1,146 @@
...
+    "amount": 87
+  }
+}
+```
+
+The peer recognizes this as a private event and tracks it for notification purposes.
+
+## Listening to Private Events
+
+The private events are restricted and stored in the similar way as the private data.
```
C0rWin, line 83: I am slightly confused. If you store these events in PDC, why can't you use the existing PDC mechanism and instead suggesting quite intrusive changes?

0000-private-event.md
```diff
@@ -0,0 +1,146 @@
...
+## Replaying Events
+
+Private events are replay-able by leveraging their PDC storage. Even though they are not part of the public ledger, the peer can match them to their corresponding transaction ID and block, making replay feasible from a given block number.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Event Creation
+- A chaincode uses `PutPrivateData(<pdc>, "__hlf_privateEvent__<name>", <eventData>)` to raise a private event.
+- `eventData` must be valid JSON with the `visibleTo` and the `payload` fields.
```
C0rWin: If you would like to actually introduce a new capability, doesn't it deserve APIs on its own?

C0rWin:
Thank you @yeasy for this RFC on Private Chaincode Events! This valuable contribution addresses an important privacy need in the Hyperledger Fabric ecosystem. The concept of private events that can be targeted to specific organizations and users has significant potential.

While the RFC provides a good high-level overview, I believe it would benefit from more detailed explanations in several key areas:

## Motivation and Context
- Could you elaborate on specific use cases that drove this feature request?
- What limitations in the current public event system are most problematic for users?

## Technical Implementation
- **Backward Compatibility**: How will this feature coexist with the current event system? 
- **Mixed Network Compatibility**: What happens in mixed networks where some peers have this feature and others don't?
- **REST Proxy**: The document mentions a "REST Proxy" but doesn't explain this or how it fits into the Fabric architecture.
- **Peer Changes**: The current explanation of changes required on the peer side is relatively high-level. Could you provide more specifics on what modifications are needed?

## Security Model
- How is the security model designed to prevent unauthorized access to private events?
- **User Authentication**: What is the flow for authenticating specific users mentioned in the `visibleTo` field? How does a peer validate these identities?

## Implementation Approach
- The current approach of using `PutPrivateData()` with special prefixes seems to add custom logic to existing APIs rather than providing a more direct, purpose-built solution. Have you been able to consider alternatives?
- What would be the impact of this approach on smart contract development patterns?

## Configuration and Management
- **PDC Configuration**: What specific PDC configuration is required? The RFC doesn't mention how to set up and configure the collections needed.
- **Feature Enablement**: How should operators enable and configure this feature?
- **Lifecycle Management**: How are updates and deletions of private events handled?
- What happens when PDC state is purged according to policy?

## Client Experience
- How would client applications need to be modified to use this feature?
- What SDK changes would be required?

I appreciate the effort put into this proposal, and I believe it would be much stronger with more detailed explanations in these areas. The core concept is valuable, but the current RFC is at a high level and would benefit from addressing these technical details before moving forward.

Looking forward to your thoughts on these questions!

0000-private-event.md
yeasy: Will update.

0000-private-event.md
yeasy: It's possible. This choice can add flexibility to satisfy user's complicated requirement.

0000-private-event.md
yeasy: It's possible, will update.

yeasy:
> * Could you elaborate on specific use cases that drove this feature request?

Those details will be added in a new version. Thanks!

C0rWin:
> Those details will be added in a new version. Thanks!

Thanks @yeasy, however, most of my questions remain unanswered, hence I'd ask you to complete the RFC with more details. 

Moreover, to a large extent, I do not understand the essence of making any changes to Fabric. The mentioned functionality could be achieved solely with PDC without introducing any modifications to the Fabric. You can define a collection where you will put keys identifying events you would like to distribute. 

If you'd still like to change the Fabric, I suggest providing a more detailed design and proposal. Which pieces of fabric are you going to change? What changes do you need to make for SDKs? Which SDK are you planning to support and change? How will you sync, retain, and operate with these events? Can we extend the definition to add a new organization? To remove? What is the suggested configuration? Operation-wise, how and who will be responsible for configuring PDC to disseminate these events?

yeasy:
> > Those details will be added in a new version. Thanks!
> 
> Thanks @yeasy, however, most of my questions remain unanswered, hence I'd ask you to complete the RFC with more details.
> 
> Moreover, to a large extent, I do not understand the essence of making any changes to Fabric. The mentioned functionality could be achieved solely with PDC without introducing any modifications to the Fabric. You can define a collection where you will put keys identifying events you would like to distribute.
> 
> If you'd still like to change the Fabric, I suggest providing a more detailed design and proposal. Which pieces of fabric are you going to change? What changes do you need to make for SDKs? Which SDK are you planning to support and change? How will you sync, retain, and operate with these events? Can we extend the definition to add a new organization? To remove? What is the suggested configuration? Operation-wise, how and who will be responsible for configuring PDC to disseminate these events?

@C0rWin thanks for the review!

I understand you do not see why needs the change happens, because it's difficult to imagine the process by a pure RFC.

Let me give you one example. Currently, for the public events which are stored in the block, the client SDK can call event APIs to fetch them from the peer node. Today how can a client fetch private events? We need to change the peer code and potentially SDK code to handle the private event.

For peer side, the peer node needs to change: 1) to handle those special private value; 2) parse them 3) provide them via the event APIs when a client fetches them; 4) how to authenticate those clients. 5) And more.

For sdk side, we are thinking of minimizing the change on SDK client side, to keep compatible with existing applications.

For the adding/removing orgs, yes, we can, but this part mainly take advantage of the PDC functionality.

PDC is still managed by those private group members.

We hope the coming code PRs will provide more details, and welcome to help review them, too.

pfi79:
> > > Those details will be added in a new version. Thanks!
> > 
> > 
> > Thanks @yeasy, however, most of my questions remain unanswered, hence I'd ask you to complete the RFC with more details.
> > Moreover, to a large extent, I do not understand the essence of making any changes to Fabric. The mentioned functionality could be achieved solely with PDC without introducing any modifications to the Fabric. You can define a collection where you will put keys identifying events you would like to distribute.
> > If you'd still like to change the Fabric, I suggest providing a more detailed design and proposal. Which pieces of fabric are you going to change? What changes do you need to make for SDKs? Which SDK are you planning to support and change? How will you sync, retain, and operate with these events? Can we extend the definition to add a new organization? To remove? What is the suggested configuration? Operation-wise, how and who will be responsible for configuring PDC to disseminate these events?
> 
> @C0rWin thanks for the review!
> 
> I understand you do not see why needs the change happens, because it's difficult to imagine the process by a pure RFC.
> 
> Let me give you one example. Currently, for the public events which are stored in the block, the client SDK can call event APIs to fetch them from the peer node. Today how can a client fetch private events? We need to change the peer code and potentially SDK code to handle the private event.
> 
> For peer side, the peer node needs to change: 1) to handle those special private value; 2) parse them 3) provide them via the event APIs when a client fetches them; 4) how to authenticate those clients. 5) And more.
> 
> For sdk side, we are thinking of minimizing the change on SDK client side, to keep compatible with existing applications.
> 
> For the adding/removing orgs, yes, we can, but this part mainly take advantage of the PDC functionality.
> 
> PDC is still managed by those private group members.
> 
> We hope the coming code PRs will provide more details, and welcome to help review them, too.

The way rfc is described now, it's very difficult to understand the final intent, and adjust it.
Take the latest, fresh one as an [example](https://github.com/hyperledger/fabric-rfcs/pull/58).
Both the pr description and the document.

C0rWin:
> > > Those details will be added in a new version. Thanks!
> > 
> > 
> > Thanks @yeasy, however, most of my questions remain unanswered, hence I'd ask you to complete the RFC with more details.
> > Moreover, to a large extent, I do not understand the essence of making any changes to Fabric. The mentioned functionality could be achieved solely with PDC without introducing any modifications to the Fabric. You can define a collection where you will put keys identifying events you would like to distribute.
> > If you'd still like to change the Fabric, I suggest providing a more detailed design and proposal. Which pieces of fabric are you going to change? What changes do you need to make for SDKs? Which SDK are you planning to support and change? How will you sync, retain, and operate with these events? Can we extend the definition to add a new organization? To remove? What is the suggested configuration? Operation-wise, how and who will be responsible for configuring PDC to disseminate these events?
> 
> @C0rWin thanks for the review! 
> 
> I understand you do not see why needs the change happens, because it's difficult to imagine the process by a pure RFC.
> 
> Let me give you one example. Currently, for the public events which are stored in the block, the client SDK can call event APIs to fetch them from the peer node. Today how can a client fetch private events? We need to change the peer code and potentially SDK code to handle the private event.
> 
> For peer side, the peer node needs to change: 1) to handle those special private value; 2) parse them 3) provide them via the event APIs when a client fetches them; 4) how to authenticate those clients. 5) And more.
> 
> For sdk side, we are thinking of minimizing the change on SDK client side, to keep compatible with existing applications.
> 
> For the adding/removing orgs, yes, we can, but this part mainly take advantage of the PDC functionality.
> 
> PDC is still managed by those private group members.
> 
> We hope the coming code PRs will provide more details, and welcome to help review them, too.

There is something I do not understand. Why can't you define PDC and make relevant clients of required organizations to fetch from these PDC using the current APIs available within the SDKs? Why do you need to introduce an additional mechanism here? 

And again, if you do want to bring new functionality and do intrusive changes to fabric code base, please make sure to provide enough of the details outlining design choices, the changes you going to make to the fabric, also taking into account SDKs, etc... I will go strongly oppose the RFC with comments like: "let's do code and you will see how it fits into fabric". This is not the way it should work. Please take your time and provide a comprehensive RFC covering all the requested aspects.

C0rWin:
@yeasy, please add some meaningful examples with the explanation of the use case, which might justify the suggested changes, from the descriptions so far it could be done with PDC and the only rational I got from your comments is that it might be more convenient to put some logic you currently have on application side into the peer's code. And I found this approach fundamentally wrong.

yeasy:
The major review comment is whether we should implement this private event at client SDK side, rather than the peer node side.

The driving force is to let fabric users keep using the default chaincode event mechanism without having to change their existing client applications to manage those events ad-hoc. This requirement is from real user cases.

If you have one application, it's fine. But if you have 10s or 100s of running applications, it will be extremely inconvenient to repeat the change on each application at client side.

Our approach allows you to stay at the already familiar framework and APIs, and not reinvent anything. By reusing the PDC mechanism, we also keep minimal changes on peer side, and follow the existing event creation, store, fetching workflow.

Will update those details into the newer version of RFC.

0000-private-event.md
yeasy: Those are example users in the msp. Can validate by comparing the CN in cert.

0000-private-event.md
yeasy: Current PDC mechanism does not offer to create an event by reading data from it. This still needs peer node change.

C0rWin:
@yeasy, Thank you for clarifying the user experience motivation behind implementing this at the peer level. I understand the convenience argument for avoiding changes across multiple client applications.

However, the primary concern remains the lack of sufficient technical details to properly evaluate this RFC. While the justification for user experience is valuable, we need concrete technical specifications to assess the proposed changes' feasibility, security implications, and overall impact.

To make an informed judgment on this RFC, please provide detailed explanations for:

### 1. Peer Implementation Specifics

* Which specific peer components will be modified (event service, PDC handler, etc.)?
* How will the peer differentiate between regular PDC data and private events during storage and retrieval?

### 2. Authentication & Authorization Model

* How exactly will peer nodes authenticate and authorize clients requesting private events?
* What happens when a client's permissions change after events are stored?
* How do you handle revocation scenarios for users specified in visibleTo?

###  3. API Design & SDK Changes

* What specific changes are required in each SDK (Node.js, Java, Go)?
* How will clients specify which private events they want to subscribe to?

### 4. Configuration & Operations

* What configuration changes are needed for channel configuration, PDC policies, and peer configuration?
* How do operators manage and monitor private event collections?
* What's the lifecycle management for private events (creation, updates, deletion, purging)?

### 5. Backward Compatibility & Migration

* How will this feature interact with existing event clients that don't support private events?
* What's the migration path for networks that want to enable this feature?
* How does this work in mixed environments where some peers support this and others don't?

## Alternative Implementation Consideration

One potential middle-ground approach could be:

* Develop a standardized client-side library/wrapper that abstracts the PDC-based private event handling
* This could provide the same developer experience benefits without requiring core Fabric changes
* The library could handle authentication, filtering, and event management transparently
* This would achieve your "familiar APIs" goal while avoiding intrusive peer modifications

yeasy:
> @yeasy, Thank you for clarifying the user experience motivation behind implementing this at the peer level. I understand the convenience argument for avoiding changes across multiple client applications.

Thanks for the understanding!

> 
> However, the primary concern remains the lack of sufficient technical details to properly evaluate this RFC. While the justification for user experience is valuable, we need concrete technical specifications to assess the proposed changes' feasibility, security implications, and overall impact.
> 
> To make an informed judgment on this RFC, please provide detailed explanations for:
> 
> ### 1. Peer Implementation Specifics
> * Which specific peer components will be modified (event service, PDC handler, etc.)?

* Event service;
* PDC
* Chaincode (if with an individual API)
* Ledger (allow fetching the events by block)

> * How will the peer differentiate between regular PDC data and private events during storage and retrieval?

Not big different. The private event will use a special prefix, as described in the RFC.

> 
> ### 2. Authentication & Authorization Model
> * How exactly will peer nodes authenticate and authorize clients requesting private events?

Check the msp and cn defined in the , as described in the RFC.

> * What happens when a client's permissions change after events are stored?

PDC part will not change. 
If the client's new permission satisfy the ACL in the `visibleTo` field, then it can fetch the events. Basically this is similar as the PDC one.

> * How do you handle revocation scenarios for users specified in visibleTo?

I guess you're talking about the cert revocation. When it happens, the peer side will get the new cert (as today), then the visibleTo checking will reject the old cert.

> 
> ### 3. API Design & SDK Changes
> * What specific changes are required in each SDK (Node.js, Java, Go)?

No need to change as today's design.

> * How will clients specify which private events they want to subscribe to?

I remember today the client can specify the event name? (correct me if there're more in latest release). This will keep the same.

> 
> ### 4. Configuration & Operations
> * What configuration changes are needed for channel configuration, PDC policies, and peer configuration?

No need to change. Unless we add a new capability, then the channel config will be updated accordingly.

> * How do operators manage and monitor private event collections?

We do not have special API for this requirement. It could be monitored by watching the collections with specific prefix in name.

> * What's the lifecycle management for private events (creation, updates, deletion, purging)?

As described, it can be: create, store, listen&fetch (client side). It's also OK to support deletion.

> 
> ### 5. Backward Compatibility & Migration
> * How will this feature interact with existing event clients that don't support private events?

No need to change. Clients can still use the public event.

> * What's the migration path for networks that want to enable this feature?

Similar as today's upgrade workflow. Basically need to replace the peer binary and restart. Update channel config (if need).

> * How does this work in mixed environments where some peers support this and others don't?

Only enabled peer allows fetching event form it.

> 
> ## Alternative Implementation Consideration
> One potential middle-ground approach could be:
> 
> * Develop a standardized client-side library/wrapper that abstracts the PDC-based private event handling
> * This could provide the same developer experience benefits without requiring core Fabric changes
> * The library could handle authentication, filtering, and event management transparently
> * This would achieve your "familiar APIs" goal while avoiding intrusive peer modifications

Yes, I agree.
As I said, there can be several approaches for various trade-off (client, peer, or others).
We want to keep the minimal change on client side to ease the application layer. As today's peer side already have good support of public events and PDC, it is a practical way to reuse these existing features as much as possible. And our proposed approach has been verified in real user cases and received good feedbacks.

pfi79:
@yeasy The rfc is still not clear to me. 
There is very little technical detail. 
A lot of things are not clear:
- why you don't use the existing way of handling private data, but invented a special key with the prefix __hlf_privateEvent?
- why did you invent a new structure for private events and not inherit the existing one?
- how will this private event be purged?
- what will the client see after purge when it requests the private event?
- what happens if a new chaincode stands in front of an old peer and tries to create a private event?
- what happens if the new client requests private events from the old peer?
- what happens if there are some new peers and some old peers on the network. And chaincode tries to create a private event?

All of this must be answered and reflected in the rfc. Later, most likely, there will be more questions and they must be answered. It is impossible to accept the rfc without it.

yeasy:
> @yeasy The rfc is still not clear to me. There is very little technical detail. A lot of things are not clear:
> 
> * why you don't use the existing way of handling private data, but invented a special key with the prefix __hlf_privateEvent?

We are using the existing way to handle private data. The reason of using special prefix is to easily distinguish from the user's stored private data.

> * why did you invent a new structure for private events and not inherit the existing one?

The new structure allows easily handling private event with flexible ACL.

> * how will this private event be purged?

Since the private event belongs to the PDC, it can be purged as the PDC (i.e., blockToLive).

> * what will the client see after purge when it requests the private event?

Return nil/empty.

> * what happens if a new chaincode stands in front of an old peer and tries to create a private event?

Return capabilities error (if with new capabilities flag) or not supported.

> * what happens if the new client requests private events from the old peer?

Return capabilities error (if with new capabilities flag) or not supported.

> * what happens if there are some new peers and some old peers on the network. And chaincode tries to create a private event?

New peers will work well, and old peers will return capabilities error (if with new capabilities flag) or not supported.

> 
> All of this must be answered and reflected in the rfc. Later, most likely, there will be more questions and they must be answered. It is impossible to accept the rfc without it.

pfi79:
> The new structure allows easily handling private event with flexible ACL.

Then I would ask you to start by creating a new separate rfc on adding a new flexible acl mechanism to fabric.

> Since the private event belongs to the PDC, it can be purged as the PDC (i.e., blockToLive).

It's not clear to me. 

> Return capabilities error (if with new capabilities flag) or not supported.

Where is this described in the rfc? How will it be done?

> New peers will work well, and old peers will return capabilities error (if with new capabilities flag) or not supported.

So a transaction at the endorsment stage can fail?

yeasy:
> > The new structure allows easily handling private event with flexible ACL.
> 
> Then I would ask you to start by creating a new separate rfc on adding a new flexible acl mechanism to fabric.

We have thought of it. However, this is only used by the private event feature as we know.

If there're other features to use it in future, we could put it as a separate one.

> 
> > Since the private event belongs to the PDC, it can be purged as the PDC (i.e., blockToLive).
> 
> It's not clear to me.

You can refer to https://hyperledger-fabric.readthedocs.io/en/latest/private-data/private-data.html#purging-private-data for more details.
Basically, if you enable the blockToLive param in your PDC, then the peer will periodically check and purge the outdated private data.

> 
> > Return capabilities error (if with new capabilities flag) or not supported.
> 
> Where is this described in the rfc? How will it be done?

This can be added into the RFC after the QA if needed. 

If you means the capabilities error, it is done by the existing channel capabilities mechanism (see common/capabilities/channel.go).
Basically the peer will compare its registered capabilities and the channel flags, and see it can support or not.

We are also open to the other way that does not add a new capabilities, which then will return errors from existing code.

Again, existing fabric already considers how to introduce new features and let peers of multiple versions work together, you do not need to reinvent anything.

> 
> > New peers will work well, and old peers will return capabilities error (if with new capabilities flag) or not supported.
> 
> So a transaction at the endorsment stage can fail?

I guess you mean the endorsement for chaincode execution on an old peer?
It's correct. The old peer does not have the new chaincode API and will generate a fail response to the client.

pfi79:
> > > The new structure allows easily handling private event with flexible ACL.
> > 
> > 
> > Then I would ask you to start by creating a new separate rfc on adding a new flexible acl mechanism to fabric.
> 
> We have thought of it. However, this is only used by the private event feature as we know.
> 
> If there're other features to use it in future, we could put it as a separate one.

I am strongly against creating a separate acl other than the one used for private data.
You are creating a private event, probably while manipulating private data. But according to you the data and the event can be accessed differently.
So far you haven't convinced me that this is the way to do it. I'm against it.

> 
> > > Since the private event belongs to the PDC, it can be purged as the PDC (i.e., blockToLive).
> > 
> > 
> > It's not clear to me.
> 
> You can refer to https://hyperledger-fabric.readthedocs.io/en/latest/private-data/private-data.html#purging-private-data for more details. Basically, if you enable the blockToLive param in your PDC, then the peer will periodically check and purge the outdated private data.
> 

The link you provided is not correct. There is no “blockToLive” string on this page.
But according to this page, there are two ways to purge private data.
As I understand your reply you suggest to do only one option for private events.
I am against it. I think we should keep both options.

> > > Return capabilities error (if with new capabilities flag) or not supported.
> > 
> > 
> > Where is this described in the rfc? How will it be done?
> 
> This can be added into the RFC after the QA if needed.
> 

Issues with working with previous versions of chaincodes, peers and clients should be reflected in the rfc.

> If you means the capabilities error, it is done by the existing channel capabilities mechanism (see common/capabilities/channel.go). Basically the peer will compare its registered capabilities and the channel flags, and see it can support or not.
> 
> We are also open to the other way that does not add a new capabilities, which then will return errors from existing code.
> 
> Again, existing fabric already considers how to introduce new features and let peers of multiple versions work together, you do not need to reinvent anything.
> 
> > > New peers will work well, and old peers will return capabilities error (if with new capabilities flag) or not supported.
> > 
> > 
> > So a transaction at the endorsment stage can fail?
> 
> I guess you mean the endorsement for chaincode execution on an old peer? It's correct. The old peer does not have the new chaincode API and will generate a fail response to the client.

How are you going to solve such errors? Because it can stop the execution of transactions in the fabric. 
The way you propose to solve these problems should be reflected in the rfc.

yeasy:
> https://hyperledger-fabric.readthedocs.io/en/latest/private-data/private-data.html#purging-private-data



> > > > The new structure allows easily handling private event with flexible ACL.
> > > 
> > > 
> > > Then I would ask you to start by creating a new separate rfc on adding a new flexible acl mechanism to fabric.
> > 
> > 
> > We have thought of it. However, this is only used by the private event feature as we know.
> > If there're other features to use it in future, we could put it as a separate one.
> 
> I am strongly against creating a separate acl other than the one used for private data. You are creating a private event, probably while manipulating private data. But according to you the data and the event can be accessed differently. So far you haven't convinced me that this is the way to do it. I'm against it.

This is optional to have a more flexible control. It's OK to keep reusing the ACL from PDC.

> 
> > > > Since the private event belongs to the PDC, it can be purged as the PDC (i.e., blockToLive).
> > > 
> > > 
> > > It's not clear to me.
> > 
> > 
> > You can refer to https://hyperledger-fabric.readthedocs.io/en/latest/private-data/private-data.html#purging-private-data for more details. Basically, if you enable the blockToLive param in your PDC, then the peer will periodically check and purge the outdated private data.
> 
> The link you provided is not correct. There is no “blockToLive” string on this page. But according to this page, there are two ways to purge private data. As I understand your reply you suggest to do only one option for private events. I am against it. I think we should keep both options.

For blockToLive, it's defined at https://hyperledger-fabric.readthedocs.io/en/latest/private-data-arch.html.

> 
> > > > Return capabilities error (if with new capabilities flag) or not supported.
> > > 
> > > 
> > > Where is this described in the rfc? How will it be done?
> > 
> > 
> > This can be added into the RFC after the QA if needed.
> 
> Issues with working with previous versions of chaincodes, peers and clients should be reflected in the rfc.

Yes, we can update all those questions/discussions into RFC.

> 
> > If you means the capabilities error, it is done by the existing channel capabilities mechanism (see common/capabilities/channel.go). Basically the peer will compare its registered capabilities and the channel flags, and see it can support or not.
> > We are also open to the other way that does not add a new capabilities, which then will return errors from existing code.
> > Again, existing fabric already considers how to introduce new features and let peers of multiple versions work together, you do not need to reinvent anything.
> > > > New peers will work well, and old peers will return capabilities error (if with new capabilities flag) or not supported.
> > > 
> > > 
> > > So a transaction at the endorsment stage can fail?
> > 
> > 
> > I guess you mean the endorsement for chaincode execution on an old peer? It's correct. The old peer does not have the new chaincode API and will generate a fail response to the client.
> 
> How are you going to solve such errors? Because it can stop the execution of transactions in the fabric. The way you propose to solve these problems should be reflected in the rfc.

We do not resolve those errors. If the peer does not support the new functionality, then it can do nothing but only warn. That's what fabric does today.
However, I guess we may enhance by adding some meaningful message into the response such as "The xx functionality needs to update the peer to xx version".

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+- Feature Name: private_chaincode_events
+- Start Date: 2025-04-22
+- RFC PR:
+- Fabric Component: core, peer
+- Fabric Issue:
+
+# Summary
+[summary]: #summary
+
+This RFC introduces the concept of **Private Chaincode Events** for Hyperledger Fabric. These events are generated by chaincode but differ from existing chaincode events in that they are stored and disseminated only to explicitly specified organizations and identities, without ever being recorded on the public ledger. This feature enhances privacy and fine-grained control over event visibility.
```
cendhu: To ensure immutability and non-repudiation, the hash of private data is recorded on the public ledger. Therefore, the hash of the private event must also be stored on the public ledger to maintain the fundamental properties of the blockchain.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Hyperledger Fabric today only provides events that are visible to all users in the channel, i.e., public events. In this case, even the users are using private data for privacy protection, the events are still visible to parties outside of the private group.
+
+Hence, numbers of Hyperledger Fabric users have expressed the need to emit chaincode events that are:
+
+1. Invisible to unauthorized parties. Those events are only available for authorized users.
+2. Not stored on the public ledger. This is to avoid private events being fetched by unauthorized users.
```
cendhu, line 26: Agree but the hash of the events need to be stored on the ledger.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+4. Replayable from a specific block or timestamp.
+
+Current public chaincode events cannot satisfy these privacy requirements. This proposal aims to provide a solution by building on existing private data collections (PDCs) and enhancing the event infrastructure to support controlled dissemination of private events.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Concept: Private Chaincode Events
+
+Private Chaincode Events are triggered by invoking a new `SetEvent(name string, payload []byte) error` method in chaincode, which will insert the event data using the standard `PutPrivateData()` method, with a specially prefixed key (e.g., `__hlf_privateEvent__eventName`). The special prefix is to easily distinguish from the user's stored private data.
```
cendhu: A few points on this proposal:

1. **API Naming:** To start with a simple point, the document should clarify if a new `SetPrivateEvent()` API will be created, as `SetEvent()` already exists.
2. **Inconsistent Design:** My main concern is that the proposal doesn't seem to introduce a true "private event" system. Public events are handled cleanly within `ChaincodeAction.events`. For consistency, a private event mechanism should logically be part of a new `ChaincodeAction.privateEvents` field. Instead, by storing events in the private data read-write set, the proposal feels more like a workaround.
3. **Misleading Abstraction:** This leads me to question the naming. Are we introducing private events, or are we actually creating a read-only sub-collection with an access control list (ACL)? The current proposal strongly resembles the latter. This feels like an inaccurate and potentially confusing abstraction. Can a chaincode, for instance, read a private event from the collection and attempt to modify it?

**Conclusion:** Overall, while I understand the pragmatic goal of reusing the private data infrastructure to minimize code changes, the proposed solution feels hacky. The naming doesn't seem to reflect the underlying mechanism, which risks creating a confusing and inconsistent feature.

cendhu: How do you guarantee that private data won't collide with event data, given that the same prefix could be manually added to a regular private key? Does the chaincode itself need to perform extra checks to prevent these collisions?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+The events can have the following structure by re-using PDC's access control.
```
cendhu: Could you please specify which access control policy you are referring to? 

1. The **collection membership policy**, which allows members to hold and pull private data from other org's peers.
2. The collection's endorsement policy.
3. The chaincode's endorsement policy.
4. The key-level endorsement policy.
5. Collection-level configurations like `memberOnlyRead` and `memberOnlyWrite`, which can permit access to non-members if set to `false`.

Clarifying which of these policies is relevant would be helpful.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+{
+  "payload": {
+    "from": "A",
+    "to": "B",
+    "amount": 87
+  }
+}
+```
+
+There are requirements that implement more fine-granular access control on top of the PDC policies. In this case, we can use an extra `visibleTo` field, which allows to check the MSP and CN in the identity cert. In this case, the private events will conform to a structure as below:
```
cendhu, line 51: As I understand it, the proposal is to store the private event within the collection's private data, which is part of the hashed read-write set. This means that under the current dissemination protocol, any peer member of the collection can pull the event data, regardless of what the `visibleTo` field specifies.

This creates a contradiction: the `visibleTo` field is intended for fine-grained access control, but it won't be effective unless we fundamentally change what data can be pulled during dissemination. While I also prefer minimal code changes, altering the private data protocol is a complex and highly error-prone task. This makes me question the viability of the proposed approach.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+```json
+{
+  "visibleTo": {
+    "Org1MSP": ["user1"],
+    "Org2MSP": ["user2", "user3"]
```
cendhu, line 57: This proposal raises two critical questions regarding identity and access control:

1. **Verifying the userID from a Certificate:**
Fabric does not use simple string user IDs like "user1". Instead, identity is based on roles (Peer, Client, etc.) and X.509 certificates that satisfy an MSP's requirements. How does the proposal intend to map a userID from the `visibleTo` field to a specific certificate? For example, would the system verify the user ID against the certificate's Common Name (CN) or another attribute in its Subject? This mapping needs to be clearly defined.

2. **ACL Enforcement at Event Registration:**
How would the proposed Access Control List (ACL) be enforced when a client registers to receive events? Could you clarify the workflow? For instance, when a peer identifies a private event in a block, does it then iterate through its list of registered event listeners, inspect each client's certificate, and check if their identity and MSP match the event's `visibleTo` map before sending the payload?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+  },
+  "payload": {
+    "from": "A",
+    "to": "B",
+    "amount": 87
+  }
+}
+```
+
+This structure can help do the parsing and the potential extending in the future. The `visibleTo` field can help provide additional access control besides the PDC. If the user only needs to use PDC's policy, then this field can be set to empty. When the client certs got revoked and the peer will get the update and reject the client's access.
```
cendhu, line 67: Does the `visibleTo` list function as a subset of the collection's members, or can it also include identities from organizations that are not members of the collection?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+We do not want to propose another access control mechanism, but to try best to reuse the PDC policies.
```
cendhu, line 69: For the initial implementation, I recommend that access be governed strictly by the existing Private Data Collection (PDC) membership policy. This approach would simplify the initial development and make the review process more manageable. The feature can be extended with more granular controls in a future version.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+Since the private events are re-using the existing PDC storage, they do not need additional database. And note that the stored private events will use a special prefix as described in the concept.
+
+
+## Listening to Private Events
+
+Existing event listening API will be adopted, to minimize the change on client side.
+
+The private events are restricted and stored in the similar way as the private data.
+
+Only peers in organizations and CNs listed in the event's `visibleTo` map will process and relay the event to eligible subscribers.
```
cendhu, line 85: I have two questions to clarify the implementation details:

1. **Data Storage vs. Event Delivery:** To confirm my understanding, will all peer members of the Private Data Collection (PDC) store the private event data, but the event payload itself is delivered only to the identities specified in the `visibleTo` list?

2. **Event Delivery API:** Regarding the client-side API, Fabric currently has `Deliver`, `DeliverFiltered`, and `DeliverWithPrivateData` services. Will this feature introduce a new delivery API, or will one of the existing services be modified to handle these private events?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+When specific users subscribe to the private event, their identifies will be checked to match the `visibleTo` rules. Those unauthorized users will be rejected.
```
cendhu, line 87: Fabric already has channel-level ACLs that control which clients can register for events using the Deliver APIs. These policies operate without specifying individual user IDs. Given this existing mechanism, it's unclear what specific changes to the access control logic are being proposed. Could you elaborate on how the new `visibleTo` check would integrate with or modify the current policy-based ACLs?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+If a valid client's permissions change after events are stored, it will not be able to access the private events it can fetch before, as the PDC will restrict the access.
```
cendhu, line 89: To ensure consistency with existing Fabric principles, access to historical private events must be based on point-in-time permissions.

For instance, the current private data reconciliation protocol correctly allows a peer to fetch data it missed, provided its organization was a member of the collection at the time the data was committed, even if its membership has since been revoked.

This same logic must apply here: if a client had the required permissions to receive an event when it was committed to the ledger, the system should guarantee their ability to access that event later, regardless of any subsequent changes to their permissions.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Replaying Events
+
+Private events are replay-able by leveraging their PDC storage. Even though they are not part of the public ledger, the peer can match them to their corresponding transaction ID and block, making replay feasible from a given block number.
```
cendhu, line 93: What does the system send to a subscribed client if the peer is missing the data for a private event? For example, in a scenario where the reconciliation process has not yet been triggered or has not yet successfully fetched the missing private event data, what will the client receive—an empty payload, an error, or no notification at all for that transaction?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+The peer needs to change to identify the private event, read them from the PDC storage, and send to the client SDK.
```
cendhu, line 95: More like what the  `DeliverWithPrivateData` API does.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Purging Events
+Since the private event belongs to the PDC, it can be purged as the PDC supports today. For example, if the PDC definition contains the `blockToLive` param, then the peer will periodically check and purge the outdated private data.
+
+After the events are purges, the request to the specific events will get a nil/empty return.
```
cendhu, line 100: Instead of sending a nil or empty response when the private event payload is missing, the peer could send the hash of the event, which is already available in the public read-write set. This hash could be accompanied by a flag indicating that the payload is currently unavailable. This approach would provide the client with immediate, verifiable proof that an event occurred and a key to use for reconciliation later.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Event Creation
+- A chaincode uses `PutPrivateEvent(<pdc>, <name>, <eventData>)` to raise a private event. There is also an option to use the underlying PDC  API such as `PutPrivateData(<pdc>, "__hlf_privateEvent__<name>", <eventData>)`. This adds minimal effort for exsiting chaincode development.
+- The `pdc` is the group of parties that can store the private events,`eventData` must be valid JSON with the `visibleTo` and the `payload` fields.
```
cendhu: Who is responsible for verifying the JSON validity of the private event payload? Is this check performed by the endorsing peers directly, or is the responsibility left to the chaincode shim library?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+After the events are purges, the request to the specific events will get a nil/empty return.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Event Creation
+- A chaincode uses `PutPrivateEvent(<pdc>, <name>, <eventData>)` to raise a private event. There is also an option to use the underlying PDC  API such as `PutPrivateData(<pdc>, "__hlf_privateEvent__<name>", <eventData>)`. This adds minimal effort for exsiting chaincode development.
+- The `pdc` is the group of parties that can store the private events,`eventData` must be valid JSON with the `visibleTo` and the `payload` fields. 
+- Peer intercepts and stores the event in the PDC and triggers event handling logic.
```
cendhu, line 108: I understand that during the transaction simulation, the peer adds the event's hash to the public read-write set and the full payload to the private read-write set. However, I'm unclear what additional "event handling logic" is triggered at the time of simulation. Could you clarify what this logic entails, since the actual event delivery to clients occurs after the block is committed?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Peer Notification Flow
+1. Peer receives the request for creating a new `privateEvent`.
+2. Peer processes filtered blocks and detects private event keys.
+3. Peer checks event visibility and constructs a payload with only allowed private events.
+4. Client can fetch the event from valid peers using the existing events APIs. This requires minimal or no change of the client SDKs.
```
cendhu, line 114: I question the claim that no changes are needed for the client SDK. Currently, the SDK retrieves public events from the `ChaincodeAction.events` field in a transaction. This proposal, however, stores private events in the read-write set. Given this fundamental difference in how the data is stored and accessed, wouldn't the client SDK require changes to support this new mechanism?

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+### Event Creation
+- A chaincode uses `PutPrivateEvent(<pdc>, <name>, <eventData>)` to raise a private event. There is also an option to use the underlying PDC  API such as `PutPrivateData(<pdc>, "__hlf_privateEvent__<name>", <eventData>)`. This adds minimal effort for exsiting chaincode development.
+- The `pdc` is the group of parties that can store the private events,`eventData` must be valid JSON with the `visibleTo` and the `payload` fields. 
+- Peer intercepts and stores the event in the PDC and triggers event handling logic.
+
+### Peer Notification Flow
+1. Peer receives the request for creating a new `privateEvent`.
+2. Peer processes filtered blocks and detects private event keys.
```
cendhu: I do not understand the flow clearly. Please split the flow into simulation and post commit of a block.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+```
+
+Payloads are Base64 encoded in the final event message.
+
+### Replay Handling
+- Peer tracks block and transaction IDs associated with private events.
+- Replay request with block number or timestamp fetches eligible events from PDCs.
+- Purging policies of PDCs may affect replay availability. It can eliminate the creation and storage of related private events.
```
cendhu, line 135: I do not understand what replay handing being discussed. Please provide some context.

0000-private-event.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Other Considerations
+
+In order to keep the compatibility with old version of peer nodes (those do not support private events), a new capability can be introduced. For a network mixed with peers of those two versions, the old peers can still support public events, and the new peers will support private event. If a client send the private event request to the old peers, it will be rejected with error message as the peer nodes can not recognize the request. And the chaincode invoke request to the old peer nodes will fail due to the lack of the new chaincode API `PutPrivateEvent`.
```
cendhu, line 139: I have serious concerns about the proposed upgrade path using a capability flag in a mixed-version network. My understanding is that if you try to enable a new capability, the configuration block itself will be rejected by the consenus.

Furthermore, any channel configuration update requires signatures from a majority of organization admins. They would be unwilling to sign a configuration change before their own peers are successfully upgraded. For these reasons, I believe the proposed approach of enabling the capability before all peers are updated is not viable.

cendhu (commented):
Thank you, @yeasy, for putting this proposal together. While the proposed feature is valuable, I have several significant concerns regarding the current design. In its current state, the proposal seems incomplete and would benefit from further detail. Please refer to my individual comments, which outline these points. I look forward to reviewing an updated version of the RFC.

0000-private-event.md
yeasy: Yes? And this is the property of private data, do you mean we should mention it explicitly?

0000-private-event.md
yeasy: Yes, because we cannot calculate the original value from the hash.

0000-private-event.md
yeasy: 1. will update the name;
2. We cannot store the private event in chaincodeAction, which will leak the information to public scope;
3. the naming will change in newer version.

yeasy: We can set the prefix as reserved. It's possible the chaincode could override any private data, intentionally or not. This is what private data does today.

0000-private-event.md
yeasy: The collection's membership policy, a.k.a., the PDC policy.

0000-private-event.md
yeasy: `visibleTo` field is optional now.
However, in order to do #1, we can check the Common Name (CN) field.
For 2, when client connects and request the events, the peer can check its certificate, as happening today.

0000-private-event.md
yeasy: subset of the collection's members.

0000-private-event.md
yeasy: Yes as in current version. Will clarify in new version.

0000-private-event.md
yeasy: 1. True.
2. No, reuse these APIs.

0000-private-event.md
yeasy: `visibleTo` is optional, which is specific for the private event.
While the channel-level ACLs are used in other purposes such as chaincode endorsement.

0000-private-event.md
yeasy: client received a message that indicates the event is missing/not existed.

0000-private-event.md
yeasy: It's a good option.

0000-private-event.md
yeasy: The priority is that: chaincode shim, peer.
I suggest we do both.

0000-private-event.md
yeasy: Will update

0000-private-event.md
yeasy: The target is no change.

0000-private-event.md
yeasy: it's the process when peer are missing numbers of blocks, and then received them to replay the missed transactions and the corresponding events.

0000-private-event.md
yeasy: It's the same workflow as today's upgrade mechanism to enable new capability flag, why will be rejected by the consensus?
And the peer binaries's upgrade is done before the configuration change.
I do not see it's technically blocked, unless the users are not willing to, then no upgrade is possible, whatever type.

0000-private-event.md
yeasy: Similar.

0000-private-event.md
yeasy: Yes, it's also suggested by others, and in current version, we are reusing the PDC's membership policy.
The `visibleTo` is optionally to control the client's access.

0000-private-event.md
yeasy: The same logic is adopted as the private data's access control today.

0000-private-event.md
yeasy: Here we can add some additional processing, such as checking the event's format or further validations or optimizations.

yeasy:
@C0rWin @pfi79 @cendhu @denyeart 
This RFC has been reviewed for quite a while, all proposed questions are addressed with a new updated commit.
Please let me know if there's more question. Otherwise, please help approve it and we can continue with the code commits.
Thanks!

cendhu:
@yeasy My primary question is whether this functionality needs to be implemented at the core protocol level. Hyperledger Fabric's current model for private data ensures that all members of a private data collection receive the data. If finer-grained access control is needed to restrict which specific user within an organization can query that data, this can already be managed at the application or middleware layer.

For example, an external middleware application could inspect the full client identity certificate and enforce access control rules before returning private data. This approach achieves a similar outcome of restricting data access based on identity, without fundamentally altering the protocol's identity management and policy evaluation mechanisms.

This proposal shifts a significant piece of what is currently an application-level responsibility into the base protocol. Could you elaborate on the limitations of application-level approaches that make this protocol-level change necessary? I'm interested in better understanding the trade-offs of adding this complexity to the core protocol versus leveraging existing application patterns.

yeasy:
> @yeasy My primary question is whether this functionality needs to be implemented at the core protocol level. Hyperledger Fabric's current model for private data ensures that all members of a private data collection receive the data. If finer-grained access control is needed to restrict which specific user within an organization can query that data, this can already be managed at the application or middleware layer.

As I mentioned, the find-grained access control is optional. Hence we can think it's the same scope with the PDC policy by default.

> 
> For example, an external middleware application could inspect the full client identity certificate and enforce access control rules before returning private data. This approach achieves a similar outcome of restricting data access based on identity, without fundamentally altering the protocol's identity management and policy evaluation mechanisms.

Yes, it's controlled by the external application. And it's out of the scope of this proposal. We are targeting a private event mechanism, not for find-grained access control.

> 
> This proposal shifts a significant piece of what is currently an application-level responsibility into the base protocol. Could you elaborate on the limitations of application-level approaches that make this protocol-level change necessary? I'm interested in better understanding the trade-offs of adding this complexity to the core protocol versus leveraging existing application patterns.

I have answered the same question before, post it again:

"
The driving force is to let fabric users keep using the default chaincode event mechanism without having to change their existing client applications to manage those events ad-hoc. This requirement is from real user cases.

If you have one application, it's fine. But if you have 10s or 100s of running applications, it will be extremely inconvenient to repeat the change on each application at client side.

Our approach allows you to stay at the already familiar framework and APIs, and not reinvent anything. By reusing the PDC mechanism, we also keep minimal changes on peer side, and follow the existing event creation, store, fetching workflow.
"

yeasy:
Hi, all the questions have been addresssed and there's no new review since last week. Could someone help approve the RFC please? It has blocked the code contribution for 3 months.
Thanks!

C0rWin:
@yeasy Thank you for your continued work on this RFC and for addressing the various questions raised. After looking over the extensive discussion and technical details, I'd like to share some observations that might help move this forward.

Several technical issues remain that would benefit from clarification:

1. **Design Consistency**: The current approach of storing events as private data with a special prefix (`__hlf_privateEvent__`) creates an architectural inconsistency. These aren't truly "events" in the Fabric sense but rather PDC entries that happen to be treated specially. This raises questions like: Can chaincode read these "events" back? Can they be modified? The abstraction conflates two different concepts.

2. **Access Control**: The `visibleTo` field introduces a fundamental contradiction. Under the current PDC dissemination protocol, any peer that's a member of the collection can pull this data, regardless of what `visibleTo` specifies. Modifying the private data dissemination protocol to honor these restrictions would be a significant and risky change to core Fabric behavior.

3. **Identity Mapping**: The proposal needs to clarify how simple strings like "user1" in the `visibleTo` field map to actual X.509 certificates. Fabric's identity model is based on MSPs and certificates, not usernames. Would this check against the CN field? What about other certificate attributes?

4. **Collision Prevention**: How do we guarantee that regular private data operations won't accidentally use the `__hlf_privateEvent__` prefix? This could lead to data corruption or security issues.

While I understand the motivation to minimize changes across multiple client applications, I'm concerned about whether this justifies modifications to Fabric's core protocol. As several reviewers have noted, this functionality can be achieved at the application layer using existing PDC mechanisms.

The proposal asks Fabric core to handle what appears to be an application-specific requirement - managing event visibility for a subset of users within organizations. This pattern (pushing application logic into the protocol layer) could set a precedent that leads to protocol bloat over time.

Given that:
- This functionality can be implemented using existing PDC features at the application layer
- The use case, while valid, appears specific rather than universally applicable
- Multiple maintainers have expressed reservations about the approach
- The technical implementation has unresolved inconsistencies

Should we consider developing this as a **reference implementation or standard library** that sits above Fabric rather than modifying the core protocol? This would:
1. Allow organizations that need this feature to use it immediately
2. Prove out the design through real-world usage
3. Avoid risky changes to Fabric's core components
4. Enable iteration and improvement without protocol-level commitments

If adoption proves widespread and the design matures, we could revisit integration into Fabric core with the benefit of production experience and a proven implementation.

What do you think about this approach? It might offer a faster path to delivering value to users while avoiding the risks of premature protocol changes.

yeasy:
Hi, @C0rWin Thanks for the new comments!  Please see my response in line.

Besides, is the fabric team still interested with this proposal? It's OK to let us know if the team does not want to have it.

We feel this communication is kind of tardy. We repeatedly answer the questions, and then wait several months to get the feedback. These super long delays should never happen in an active open source community.

> @yeasy Thank you for your continued work on this RFC and for addressing the various questions raised. After looking over the extensive discussion and technical details, I'd like to share some observations that might help move this forward.
> 
> Several technical issues remain that would benefit from clarification:
> 
> 1. **Design Consistency**: The current approach of storing events as private data with a special prefix (`__hlf_privateEvent__`) creates an architectural inconsistency. These aren't truly "events" in the Fabric sense but rather PDC entries that happen to be treated specially. This raises questions like: Can chaincode read these "events" back? Can they be modified? The abstraction conflates two different concepts.

Considering we don't have a notion of "private ledger" in fabric, and considering chaincode events may be important to replay, storing them in State provides a simple way to deal with those requirements without inventing a completely separate mechanism from what already exists. Yes, such chaincode events can be played back; yes, in theory they can be modified - they are just entries in the StateDB; but, as for anything else, any action to modify them would need to be approved as part of the chaincode performing the action, and it would be tracked in ledger anyway.

> 2. **Access Control**: The `visibleTo` field introduces a fundamental contradiction. Under the current PDC dissemination protocol, any peer that's a member of the collection can pull this data, regardless of what `visibleTo` specifies. Modifying the private data dissemination protocol to honor these restrictions would be a significant and risky change to core Fabric behavior.

As we answered before, the `visibleTo` is not needed. It will not be included in the change.

> 3. **Identity Mapping**: The proposal needs to clarify how simple strings like "user1" in the `visibleTo` field map to actual X.509 certificates. Fabric's identity model is based on MSPs and certificates, not usernames. Would this check against the CN field? What about other certificate attributes?

No `visibleTo` field exists any more.

> 4. **Collision Prevention**: How do we guarantee that regular private data operations won't accidentally use the `__hlf_privateEvent__` prefix? This could lead to data corruption or security issues.

We don't - even if we could, of course; the user is likely not so dumb to use a field with a prefix showing a clearly "this is for internal use" intention. If they do that intentionally, then we go back to what mentioned above.

There's no perfect solution unless fabric has a new private ledger concept, which will be a big change. I believe this proposal is the potentially best solution without going to that way.

> 
> While I understand the motivation to minimize changes across multiple client applications, I'm concerned about whether this justifies modifications to Fabric's core protocol. As several reviewers have noted, this functionality can be achieved at the application layer using existing PDC mechanisms.
> 
> The proposal asks Fabric core to handle what appears to be an application-specific requirement - managing event visibility for a subset of users within organizations. This pattern (pushing application logic into the protocol layer) could set a precedent that leads to protocol bloat over time.
> 
> Given that:
> 
> * This functionality can be implemented using existing PDC features at the application layer
> * The use case, while valid, appears specific rather than universally applicable
> * Multiple maintainers have expressed reservations about the approach
> * The technical implementation has unresolved inconsistencies
> 
> Should we consider developing this as a **reference implementation or standard library** that sits above Fabric rather than modifying the core protocol? This would:
> 
> 1. Allow organizations that need this feature to use it immediately
> 2. Prove out the design through real-world usage
> 3. Avoid risky changes to Fabric's core components
> 4. Enable iteration and improvement without protocol-level commitments
> 
> If adoption proves widespread and the design matures, we could revisit integration into Fabric core with the benefit of production experience and a proven implementation.
> 
> What do you think about this approach? It might offer a faster path to delivering value to users while avoiding the risks of premature protocol changes.

pfi79:
> Hi, @C0rWin Thanks for the new comments! Please see my response in line.
> 
> Besides, is the fabric team still interested with this proposal? It's OK to let us know if the team does not want to have it.
> 
> We feel this communication is kind of tardy. We repeatedly answer the questions, and then wait several months to get the feedback. These super long delays should never happen in an active open source community.
> 
> > @yeasy Thank you for your continued work on this RFC and for addressing the various questions raised. After looking over the extensive discussion and technical details, I'd like to share some observations that might help move this forward.
> > Several technical issues remain that would benefit from clarification:
> > 
> > 1. **Design Consistency**: The current approach of storing events as private data with a special prefix (`__hlf_privateEvent__`) creates an architectural inconsistency. These aren't truly "events" in the Fabric sense but rather PDC entries that happen to be treated specially. This raises questions like: Can chaincode read these "events" back? Can they be modified? The abstraction conflates two different concepts.
> 
> Considering we don't have a notion of "private ledger" in fabric, and considering chaincode events may be important to replay, storing them in State provides a simple way to deal with those requirements without inventing a completely separate mechanism from what already exists. Yes, such chaincode events can be played back; yes, in theory they can be modified - they are just entries in the StateDB; but, as for anything else, any action to modify them would need to be approved as part of the chaincode performing the action, and it would be tracked in ledger anyway.
> 
> > 2. **Access Control**: The `visibleTo` field introduces a fundamental contradiction. Under the current PDC dissemination protocol, any peer that's a member of the collection can pull this data, regardless of what `visibleTo` specifies. Modifying the private data dissemination protocol to honor these restrictions would be a significant and risky change to core Fabric behavior.
> 
> As we answered before, the `visibleTo` is not needed. It will not be included in the change.
> 
> > 3. **Identity Mapping**: The proposal needs to clarify how simple strings like "user1" in the `visibleTo` field map to actual X.509 certificates. Fabric's identity model is based on MSPs and certificates, not usernames. Would this check against the CN field? What about other certificate attributes?
> 
> No `visibleTo` field exists any more.
> 
> > 4. **Collision Prevention**: How do we guarantee that regular private data operations won't accidentally use the `__hlf_privateEvent__` prefix? This could lead to data corruption or security issues.
> 
> We don't - even if we could, of course; the user is likely not so dumb to use a field with a prefix showing a clearly "this is for internal use" intention. If they do that intentionally, then we go back to what mentioned above.
> 
> There's no perfect solution unless fabric has a new private ledger concept, which will be a big change. I believe this proposal is the potentially best solution without going to that way.
> 
> > While I understand the motivation to minimize changes across multiple client applications, I'm concerned about whether this justifies modifications to Fabric's core protocol. As several reviewers have noted, this functionality can be achieved at the application layer using existing PDC mechanisms.
> > The proposal asks Fabric core to handle what appears to be an application-specific requirement - managing event visibility for a subset of users within organizations. This pattern (pushing application logic into the protocol layer) could set a precedent that leads to protocol bloat over time.
> > Given that:
> > 
> > * This functionality can be implemented using existing PDC features at the application layer
> > * The use case, while valid, appears specific rather than universally applicable
> > * Multiple maintainers have expressed reservations about the approach
> > * The technical implementation has unresolved inconsistencies
> > 
> > Should we consider developing this as a **reference implementation or standard library** that sits above Fabric rather than modifying the core protocol? This would:
> > 
> > 1. Allow organizations that need this feature to use it immediately
> > 2. Prove out the design through real-world usage
> > 3. Avoid risky changes to Fabric's core components
> > 4. Enable iteration and improvement without protocol-level commitments
> > 
> > If adoption proves widespread and the design matures, we could revisit integration into Fabric core with the benefit of production experience and a proven implementation.
> > What do you think about this approach? It might offer a faster path to delivering value to users while avoiding the risks of premature protocol changes.

Why do you think you haven't received any approval? 

I'll answer for myself. 
You have responded to all my comments, as it seems to you, but you have not changed anything in your PR. 
That is, apart from explanations, nothing happens. 
For me, your proposal in this form is not acceptable. 

Even the VisibleTo field is still present in your pr. And you stated above that he is no longer there.

yeasy:
> > Hi, @C0rWin Thanks for the new comments! Please see my response in line.
> > Besides, is the fabric team still interested with this proposal? It's OK to let us know if the team does not want to have it.
> > We feel this communication is kind of tardy. We repeatedly answer the questions, and then wait several months to get the feedback. These super long delays should never happen in an active open source community.
> > > @yeasy Thank you for your continued work on this RFC and for addressing the various questions raised. After looking over the extensive discussion and technical details, I'd like to share some observations that might help move this forward.
> > > Several technical issues remain that would benefit from clarification:
> > > 
> > > 1. **Design Consistency**: The current approach of storing events as private data with a special prefix (`__hlf_privateEvent__`) creates an architectural inconsistency. These aren't truly "events" in the Fabric sense but rather PDC entries that happen to be treated specially. This raises questions like: Can chaincode read these "events" back? Can they be modified? The abstraction conflates two different concepts.
> > 
> > 
> > Considering we don't have a notion of "private ledger" in fabric, and considering chaincode events may be important to replay, storing them in State provides a simple way to deal with those requirements without inventing a completely separate mechanism from what already exists. Yes, such chaincode events can be played back; yes, in theory they can be modified - they are just entries in the StateDB; but, as for anything else, any action to modify them would need to be approved as part of the chaincode performing the action, and it would be tracked in ledger anyway.
> > > 2. **Access Control**: The `visibleTo` field introduces a fundamental contradiction. Under the current PDC dissemination protocol, any peer that's a member of the collection can pull this data, regardless of what `visibleTo` specifies. Modifying the private data dissemination protocol to honor these restrictions would be a significant and risky change to core Fabric behavior.
> > 
> > 
> > As we answered before, the `visibleTo` is not needed. It will not be included in the change.
> > > 3. **Identity Mapping**: The proposal needs to clarify how simple strings like "user1" in the `visibleTo` field map to actual X.509 certificates. Fabric's identity model is based on MSPs and certificates, not usernames. Would this check against the CN field? What about other certificate attributes?
> > 
> > 
> > No `visibleTo` field exists any more.
> > > 4. **Collision Prevention**: How do we guarantee that regular private data operations won't accidentally use the `__hlf_privateEvent__` prefix? This could lead to data corruption or security issues.
> > 
> > 
> > We don't - even if we could, of course; the user is likely not so dumb to use a field with a prefix showing a clearly "this is for internal use" intention. If they do that intentionally, then we go back to what mentioned above.
> > There's no perfect solution unless fabric has a new private ledger concept, which will be a big change. I believe this proposal is the potentially best solution without going to that way.
> > > While I understand the motivation to minimize changes across multiple client applications, I'm concerned about whether this justifies modifications to Fabric's core protocol. As several reviewers have noted, this functionality can be achieved at the application layer using existing PDC mechanisms.
> > > The proposal asks Fabric core to handle what appears to be an application-specific requirement - managing event visibility for a subset of users within organizations. This pattern (pushing application logic into the protocol layer) could set a precedent that leads to protocol bloat over time.
> > > Given that:
> > > 
> > > * This functionality can be implemented using existing PDC features at the application layer
> > > * The use case, while valid, appears specific rather than universally applicable
> > > * Multiple maintainers have expressed reservations about the approach
> > > * The technical implementation has unresolved inconsistencies
> > > 
> > > Should we consider developing this as a **reference implementation or standard library** that sits above Fabric rather than modifying the core protocol? This would:
> > > 
> > > 1. Allow organizations that need this feature to use it immediately
> > > 2. Prove out the design through real-world usage
> > > 3. Avoid risky changes to Fabric's core components
> > > 4. Enable iteration and improvement without protocol-level commitments
> > > 
> > > If adoption proves widespread and the design matures, we could revisit integration into Fabric core with the benefit of production experience and a proven implementation.
> > > What do you think about this approach? It might offer a faster path to delivering value to users while avoiding the risks of premature protocol changes.
> 
> Why do you think you haven't received any approval?
> 
> I'll answer for myself. You have responded to all my comments, as it seems to you, but you have not changed anything in your PR. That is, apart from explanations, nothing happens. For me, your proposal in this form is not acceptable.
> 
> Even the VisibleTo field is still present in your pr. And you stated above that he is no longer there.

If you compare the current proposal with the previous version, you will find it has been updated.

The VisibleTo field is still present, but marked as "optional" to receive comments from other reviewers. 
No reviewer says it should be removed. 
If you think this is a blocker for the approval, sure, then it will be removed.

pfi79:
> If you compare the current proposal with the previous version, you will find it has been updated.
> 
> The VisibleTo field is still present, but marked as "optional" to receive comments from other reviewers. No reviewer says it should be removed. If you think this is a blocker for the approval, sure, then it will be removed.

Start by getting rid of the __hlf_privateEvent__ constant.
1. The user should not add it manually, and should not know that such a constant (or similar) exists. See how data storage works in fabric. There are some constants there, but they are hidden from users.
2. I understand that you wanted to use this constant to separate data and events. But you can have an event with the same name in each transaction. And therefore they should not be overwritten in the repository. I did not see in your proposal how you would separate events with the same name, but in different transactions.

So far, I consider your proposal not technically prepared.

yeasy:
> > If you compare the current proposal with the previous version, you will find it has been updated.
> > The VisibleTo field is still present, but marked as "optional" to receive comments from other reviewers. No reviewer says it should be removed. If you think this is a blocker for the approval, sure, then it will be removed.
> 
> Start by getting rid of the **hlf_privateEvent** constant.
> 
> 1. The user should not add it manually, and should not know that such a constant (or similar) exists. See how data storage works in fabric. There are some constants there, but they are hidden from users.

Are you looking at the latest version? In the latest version, the user does not add the constant manually.
```
Private Chaincode Events are triggered by invoking a new `SetPrivateEvent(name string, payload []byte) error` method in chaincode
``` 
The user only needs to specify the event name and payload in their chaincode.

> 2. I understand that you wanted to use this constant to separate data and events. But you can have an event with the same name in each transaction. And therefore they should not be overwritten in the repository. I did not see in your proposal how you would separate events with the same name, but in different transactions.

Please see the answers to the same question 1&4 from @C0rWin. 

It's kind of low efficient to repeat answering the same question from different reviewers. Is there a way that the team can help assign one or two reviewers dedicated to the proposal? This will help a lot.

> 
> So far, I consider your proposal not technically prepared.
We welcome any technical challenge, rather than silence. The solution has been working for long time well in real cases, but yes, any solution has some flaws. What we are seeking is a practical solution considering trade-off between known limits and technical flexibility.

pfi79:
> Are you looking at the latest version? In the latest version, the user does not add the constant manually.
> 

image

pfi79:
> Please see the answers to the same question 1&4 from @C0rWin.
> 
> It's kind of low efficient to repeat answering the same question from different reviewers. Is there a way that the team can help assign one or two reviewers dedicated to the proposal? This will help a lot.

OK. You keep insisting on your own. And you don't change your Proposal. You have the right to do so. But I refuse to support your proposal.

yeasy:
> > Are you looking at the latest version? In the latest version, the user does not add the constant manually.
> 
> image

"There is also an option". Do you mean we need to remove this sentence? Hard to guess why you think this mandate to add the constant manually?

yeasy:
> > Please see the answers to the same question 1&4 from @C0rWin.
> > It's kind of low efficient to repeat answering the same question from different reviewers. Is there a way that the team can help assign one or two reviewers dedicated to the proposal? This will help a lot.
> 
> OK. You keep insisting on your own. And you don't change your Proposal. You have the right to do so. But I refuse to support your proposal.

We never said to insist on something. We repeatedly answer the questions. The proposal has been updated accordingly several times. It's quite confusing what still needs to "change" per your statement.

pfi79:
> "There is also an option". Do you mean we need to remove this sentence? Hard to guess why you think this mandate to add the constant manually?

Because here we are agreeing on what will be implemented and how. A document is being generated, which can then be checked.
If it says "A", then it should be implemented. If there is a discussion history and "A" is deleted, then it definitely shouldn't be there.

C0rWin:
> Besides, is the fabric team still interested with this proposal? It's OK to let us know if the team does not want to have it.

Thank you, @yeasy, for your continued efforts on this RFC and for addressing the various questions raised throughout this extended review process.

After careful consideration of the discussions over the past several months, I'd like to share my assessment of where we stand:

**Limited Community Engagement**: Despite this RFC being open for review since April 2025, we haven't seen significant interest or support from the broader Fabric community. No other maintainers or community members have expressed a strong desire to sponsor or champion this feature, which suggests the use case may be more niche than initially anticipated.

**Marginal Protocol-Level Benefits**: While I understand the convenience argument for avoiding changes across multiple client applications, the benefits of implementing this at the protocol level appear marginal compared to the complexity it introduces. The proposal repurposes PDC storage with special prefixes to simulate private events, which, as several reviewers have noted, creates architectural inconsistencies without delivering fundamentally new capabilities that couldn't be achieved through existing mechanisms.

**Application-Layer Alternative**: The functionality described in this RFC can be implemented at the application level using existing PDC features without modifying core Fabric components. A well-designed library or middleware solution could provide the same developer experience benefits while avoiding the risks and maintenance burden of protocol-level changes.
Given these considerations, I recommend we take one of the following paths:

- Postpone this RFC until there's more unmistakable evidence of broader community need and support for protocol-level private events, or 
- Seek active sponsorship from other Fabric maintainers who are willing to champion this feature and help drive it forward with the necessary technical refinements.

Without stronger community backing or maintainer sponsorship, I don't see a clear path forward for this RFC in its current form. This doesn't diminish the value of the use case you've identified, but instead suggests that the solution is better pursued outside of Fabric core.

yeasy:
> > Besides, is the fabric team still interested with this proposal? It's OK to let us know if the team does not want to have it.
> 
> Thank you, @yeasy, for your continued efforts on this RFC and for addressing the various questions raised throughout this extended review process.
> 
> After careful consideration of the discussions over the past several months, I'd like to share my assessment of where we stand:
> 
> **Limited Community Engagement**: Despite this RFC being open for review since April 2025, we haven't seen significant interest or support from the broader Fabric community. No other maintainers or community members have expressed a strong desire to sponsor or champion this feature, which suggests the use case may be more niche than initially anticipated.
> 
> **Marginal Protocol-Level Benefits**: While I understand the convenience argument for avoiding changes across multiple client applications, the benefits of implementing this at the protocol level appear marginal compared to the complexity it introduces. The proposal repurposes PDC storage with special prefixes to simulate private events, which, as several reviewers have noted, creates architectural inconsistencies without delivering fundamentally new capabilities that couldn't be achieved through existing mechanisms.
> 
> **Application-Layer Alternative**: The functionality described in this RFC can be implemented at the application level using existing PDC features without modifying core Fabric components. A well-designed library or middleware solution could provide the same developer experience benefits while avoiding the risks and maintenance burden of protocol-level changes. Given these considerations, I recommend we take one of the following paths:
> 
> * Postpone this RFC until there's more unmistakable evidence of broader community need and support for protocol-level private events, or
> * Seek active sponsorship from other Fabric maintainers who are willing to champion this feature and help drive it forward with the necessary technical refinements.
> 
> Without stronger community backing or maintainer sponsorship, I don't see a clear path forward for this RFC in its current form. This doesn't diminish the value of the use case you've identified, but instead suggests that the solution is better pursued outside of Fabric core.

@C0rWin Thanks for letting us know the decision!

PR #66, opened 2025-08-18
CoAP Gateway for Hyperledger Fabric

bernardo-rf:
This RFC proposes the addition of a Constrained Application Protocol (CoAP) gateway to Hyperledger Fabric. The goal is to enable IoT and resource-constrained devices to interact with Fabric networks using a lightweight, UDP-based protocol. The CoAP gateway will provide a simplified interface for submitting transactions, evaluating chaincode, and receiving events, maintaining security through DTLS encryption and certificate-based authentication.

The current Hyperledger Fabric gRPC gateway has limitations for IoT and resource-constrained devices due to limited memory, processing power, and network bandwidth requirements. Many IoT deployments use UDP-based networks or have unreliable connectivity, making gRPC/HTTP2 impractical. The CoAP gateway addresses these limitations by providing lightweight, UDP-based communication with DTLS security for encrypted and authenticated connections.

Key features include:
- Lightweight, UDP-based communication using CoAP protocol
- DTLS security for encrypted and authenticated connections  
- Support for resource-constrained devices
- Event streaming capabilities
- Integration with existing Fabric MSP for device identity management
- Four core endpoints: /evaluate, /endorse, /submit, and /events
- Reuse of existing gRPC gateway logic and protobuf message structures

The CoAP gateway will be implemented as an optional peer service that acts as a protocol converter, translating CoAP requests into appropriate gRPC calls while maintaining the same security model and transaction lifecycle as the existing Fabric gateway.

This enhancement enables new use cases such as supply chain tracking with IoT sensors, smart city applications, and other scenarios requiring direct integration of resource-constrained devices with Fabric networks.

Signed-off-by: Bernardo Figueiredo <bernardo.figueiredo@voidsoftware.com>
Co-authored-by: Marco Ferreira <marco.ferreira@voidsoftware.com>
Co-authored-by: Marco Cova <marco.cova@voidsoftware.com>

### DIFF
--- a/0000-coap_gateway.md
+++ b/0000-coap_gateway.md
@@ -46,68 +46,115 @@ The main difference is the use of CoAP (UDP/DTLS) instead of gRPC (TCP/HTTP2), r
 
 Comprehensive documentation, configuration guides, security, performance, and troubleshooting guides will be provided to ensure users understand the capabilities and use of the CoAP gateway.
 
-The CoAP gateway introduces a DTLS security layer and uses X.509 certificates for client authentication, mapping them to Fabric identities. Authorization will be integrated with existing Fabric ACL mechanisms.
+The CoAP gateway introduces a **dual-certificate security model** that separates connection authentication from transaction authorization:
+
+* **DTLS Certificate**: Validates the device can connect to the gateway (checked during connection establishment)
+
+* **Signing Certificate**: Proves the device is authorized to execute specific transactions (checked by Fabric ACLs and endorsement policies)
+
+Both certificates are issued by Fabric CAs and validated against the channel's MSP configuration.
 
 ## Connecting the Physical World to Fabric
 
 Imagine you're building a supply chain network to track refrigerated goods. You have thousands of IoT sensors that monitor temperature and location. How do you get that data onto the blockchain in a secure, efficient, and standardized way?
 Historically, this required a complex middle layer: an IoT platform or custom application server that would receive data from devices (perhaps over MQTT or CoAP), validate it, and then use a Fabric SDK to submit a transaction. This adds latency, a central point of failure, development overhead, ***and requires implicit trust on that middle layer***.
-The Fabric CoAP Gateway solves this problem by creating a direct, lightweight, and secure bridge between CoAP-enabled IoT devices and the Fabric network. It allows resource-constrained devices to interact with chaincode by sending simple CoAP messages, extending the reach of Fabric to the very edge of your network.
+
+The Fabric CoAP Gateway solves this problem by creating a direct, lightweight, and secure bridge between CoAP-enabled IoT devices and the Fabric network. Using automatically generated client libraries, resource-constrained devices can interact with chaincode using the same APIs as traditional gRPC clients, but over the lightweight CoAP protocol. This extends the reach of Fabric to the very edge of your network without sacrificing developer experience or security.
 
 ## New Concepts
 
 To understand the gateway, let's introduce a few new concepts:
 
-* This is a new, optional peer service (or standalone component) that exposes a CoAP endpoint. It listens for incoming CoAP requests from registered IoT devices, translates them into Fabric transaction proposals, and returns the result to the device. Think of it as a specialized application client, like one built with the Fabric SDK, but one that speaks CoAP instead of gRPC and is optimized for IoT workloads.
+* This is a new, **optional** peer service that exposes a CoAP endpoint. It listens for incoming CoAP requests from registered IoT devices, translates them into calls to the existing gRPC Gateway implementation, and returns the result to the device. Think of it as a thin protocol adapter that makes the Fabric Gateway accessible to CoAP clients without duplicating any business logic.
 
-* The gateway acts as a protocol converter, translating CoAP requests into the appropriate gRPC calls. This API is designed to be a direct reflection of the existing Hyperledger Fabric gRPC gateway's interface, providing a consistent experience for developers.
+* The gateway acts as a lightweight protocol converter, translating CoAP requests into in-process gRPC Gateway calls. This API is an exact reflection of the existing Hyperledger Fabric gRPC gateway's interface, providing a consistent experience for developers across protocols.
 
-* Instead of mapping to dynamic URI-paths, the gateway exposes four core functions as fixed endpoints. The client is responsible for packaging all necessary information, such as the channel, chaincode name, and function arguments, within the Protobuf payload, just as they would when using the gRPC gateway directly.
+* The gateway exposes RPC-style endpoints that mirror the Gateway service methods. The client is responsible for packaging all necessary information, such as the channel, chaincode name, and function arguments, within the Protobuf payload, just as they would when using the gRPC gateway directly.
 
-Specifically, it uses:
+Specifically, it exposes:
 
-* `/evaluate` for evaluation requests
+* `/rpc/gateway.Gateway/Evaluate` for evaluation requests
 
-* `/endorse` for managing endorsements
+* `/rpc/gateway.Gateway/Endorse` for managing endorsements
 
-* `/submit` for processing transaction submissions
+* `/rpc/gateway.Gateway/Submit` for processing transaction submissions
 
-* `/events` for listening to events
+* `/rpc/gateway.Gateway/ChaincodeEvents` for listening to events
 
-* `/commit-status` for checking transaction commit status
+* `/rpc/gateway.Gateway/CommitStatus` for checking transaction commit status
 
-This design simplifies the gateway's logic by reusing the existing gRPC message structures and avoids the complexities of a dynamic mapping system. This ensures the CoAP gateway is a new entry point to existing functionalities without needing format or logic transformations.
+This design maximizes code reuse by delegating all business logic to the existing GatewayServer implementation. The CoAP adapter is purely a protocol translation layer, ensuring consistency and minimal maintenance overhead.
 
 A key difference from typical REST endpoints is that these endpoints don't receive JSON strings. Instead, they accept a binary stream encoded with Protobuf, just like the gRPC gateway. This simplifies the resource-to-chaincode mapping by allowing the CoAP gateway to act as a new entry point to existing functionalities without needing format transformations.
 
-Security is paramount. An IoT device cannot simply send data to the network. Each device must have a verifiable identity. The CoAP Gateway integrates directly with the Member Service Provider (MSP). Devices are enrolled through a Fabric CA and issued X.509 certificates, just like users or apps. These certificates are then used to establish a secure DTLS (Datagram TLS) session with the gateway, which authenticates the device and uses its identity and attributes for authorization.
+Security is paramount. An IoT device cannot simply send data to the network. Each device must have a verifiable identity. The CoAP Gateway integrates directly with the Member Service Provider (MSP). Devices are enrolled through a Fabric CA and issued two X.509 certificates: one for DTLS transport security and one for signing transactions. These certificates establish a secure DTLS (Datagram TLS) session with the gateway and authenticate the device's transactions through Fabric's standard validation mechanisms.
+
+## Client Generation and Usage
+
+The CoAP gateway provides the same developer experience as gRPC through automated client generation. When fabric-protos is built, a protoc plugin (`protoc-gen-go-coap`) generates type-safe Go clients that mirror the gRPC gateway interface.
+
+Developers simply import the generated client:
+
+```go
+import "github.com/hyperledger/fabric-protos-go-apiv2/gateway"
+
+// Create CoAP connection with DTLS certificate
+conn, err := coap.NewDTLSConnection("fabric-gateway.mycorp.com:5684",
+    coap.WithDTLSCert(transportCert, transportKey),
+    coap.WithRootCA(caCert))
+
+// Use generated client
+client := gateway.NewGatewayCoapClient(conn)
+
+// Type-safe method calls with IDE autocomplete
+resp, err := client.Endorse(ctx, &gateway.EndorseRequest{
+    TransactionId: "tx123",
+    ChannelId:     "mychannel",
+    ProposedTransaction: signedProposal,  // Signed with identity certificate
+})
+```
+
+This approach ensures that:
+
+* Client code is automatically in sync with protocol changes
+* Developers get full IDE support (autocomplete, type checking)
+* The experience matches gRPC client patterns exactly
+* When Gateway API changes, regeneration is automatic
 
 ## An Example: The Smart Cold Chain
 
-Let's start with the cold chain example. A temperature sensor on a truck needs to report its reading. The device is already enrolled and has a certificate.
+Let's start with the cold chain example. A temperature sensor on a truck needs to report its reading. The device has been enrolled and provisioned with:
 
-* The sensor wakes up, reads the temperature (e.g., -18°C). It must then create a transaction proposal, endorse it and finally submit it. For such, it will encode a protobuf transaction proposal the same way the sdk would for a gRPC endorse call and sends it to the node it is connected to:
-  * Target: coap://fabric-gateway.mycorp.com:5683/endorse
-  * Method: POST
-  * Payload: protobuf-encoded transaction
+* **DTLS certificate**: For secure connection (`device-345-dtls.pem`)
+* **Signing certificate**: For transaction signing (`device-345-identity.pem`)
 
-* The gateway receives this request over a DTLS-secured connection.
-  * It authenticates the device using its client certificate, identifying it as sensor-345 belonging to transporter-org.
-  * It extracts the standard Fabric transaction proposal, signed with the device's identity.
-  * It passes that signed proposal to the standard (gRPC) gateway for further processing.
-  * With the response calculated by the standard gateway, it encapsulates it into a CoAP response, and sends it back to the device.
+* The sensor wakes up, reads the temperature (e.g., -18°C). It must then create a transaction proposal, endorse it and finally submit it:
+  * Establishes DTLS connection using its **transport certificate**
+  * Creates and signs the proposal using its **signing certificate**
+  * **Target:** coaps://fabric-gateway.mycorp.com:5684/rpc/gateway.Gateway/Endorse
+  * **Method:** POST
+  * **Payload:** protobuf-encoded `SignedProposal` with embedded signing certificate
 
-* Once the transaction is endorsed, the gateway sends a CoAP response back to the device.
-  * Success: A 2.04 Changed response, confirming the data was recorded.
-  * Failure: A 4.03 Forbidden if the endorsement policy failed, or a 5.00 Internal Server Error if a system-level problem occurred.
+* The gateway receives this request over the DTLS-secured connection:
+  * **Stage 1**: Validates the DTLS certificate - confirms device is allowed to connect
+  * **Stage 2**: Extracts the signing certificate from `SignedProposal` and validates:
+    * Signature is valid
+    * Signing certificate belongs to `transporter-org` MSP
+    * Device has permission per channel ACLs
+  * Forwards the validated proposal to existing gRPC gateway implementation
+  * Returns CoAP response with endorsement
+
+* Once endorsed, the gateway sends a CoAP response:
+  * **Success:** `2.05 Content` with marshaled `EndorseResponse`
+  * **Failure:** `4.03 Forbidden` if ACL check failed, `5.00 Internal Server Error` for system issues
 
 After the endorsement is completed, the device starts once again with the submission phase, where the process is similar to the endorsement.
 
 ## Teaching This to Different Developers
 
-* For Established Fabric Developers: You already understand the transaction lifecycle, endorsement policies, and the MSP. The CoAP Gateway is simply a new adapter that plugs into this flow. You'll find that all your existing knowledge about chaincode development and policy definition applies directly. The main new tasks are learning the protobuf data encoding, which can be left to the sdk, and the best practices for managing IoT device identities within the MSP.
+* **For Established Fabric Developers:** The CoAP Gateway is simply a new protocol adapter that plugs into the existing Gateway implementation with zero changes to business logic. The main new concepts are: (1) the dual-certificate model for separating connection and transaction security and (2) using the generated CoAP client from `fabric-protos-go-apiv2`, which works exactly like the gRPC client you're already familiar with.
 
-* For New Fabric Developers: Fabric is a platform for building decentralized applications with trust. The CoAP Gateway is one of the most powerful ways to get trusted data into that platform from IoT platforms. It allows you to build end-to-end solutions that securely connect a physical device to a digital ledger without needing to become an expert in complex network protocols. Start with an example like our cold chain, and see how device requests can trigger powerful, trusted logic on the blockchain.
+* **For IoT Developers New to Fabric:** The CoAP Gateway enables direct integration of resource-constrained devices with blockchain networks. The dual-certificate approach is straightforward: one certificate connects you to the gateway, and another signs your transactions. Start with an example like our cold chain to see how device data flows securely onto the ledger.
 
 # Reference-level explanation
 
@@ -115,23 +162,49 @@ This section provides the technical details of the CoAP gateway implementation. 
 
 ## Protocol and Security
 
-The gateway utilizes the Constrained Application Protocol (CoAP) over Datagram Transport Layer Security (DTLS) to ensure secure, low-overhead communication.
+The gateway utilizes the Constrained Application Protocol (CoAP) over Datagram Transport Layer Security (DTLS) to ensure secure, low-overhead communication. Security is implemented through a **dual-certificate model** that separates transport-level authentication from transaction-level authorization.
 
-* **Transport Security**: All communication between the client and the CoAP gateway is secured using DTLS (CoAP-over-UDP with TLS). This provides authentication, confidentiality, and integrity for all data in transit.
+### Dual-Certificate Architecture
 
-* **Client Authentication**: Client identity is established via mutual DTLS, using the standard X.509 certificates issued by a Fabric Certificate Authority (CA). The gateway validates the client's certificate against the channel's MSP configuration. This means that existing client identities and certificates can be used without modification, and no separate identity management system is required for the CoAP gateway.
+Devices are provisioned with **two distinct X.509 certificate/key pairs**, both issued by a Fabric Certificate Authority (CA):
 
-* **Transaction Security**: Since the gateway forwards standard Fabric protobuf messages, the transaction-level security model remains unchanged. Transaction proposals must be signed by the client, and the gateway will collect signed endorsements from peers. All security and policy enforcement (like endorsement policies and ACLs) are handled by the underlying Fabric infrastructure.
+* **Transport Layer**: DTLS 1.2/1.3 with mutual X.509 authentication (connection security)  
+* **Application Layer**: Signed transaction proposals with identity certificates (transaction authorization)
+
+This is identical to the gRPC gateway's security model - the CoAP adapter validates DTLS certificates during connection establishment, then delegates all transaction-level security (ACL checks, endorsement policies) to the existing GatewayServer implementation.
+
+Devices require two certificates from Fabric CA:
+
+* **DTLS certificate:** For UDP transport authentication
+* **Identity certificate:** For signing `SignedProposal` payloads (embedded in request body)
+
+Both certificates are validated against channel MSP configuration.
+Compromised transport certificates cannot sign transactions;
+compromised identity certificates cannot intercept network traffic.
 
 ## API Endpoints
 
-The CoAP gateway exposes five main endpoints, each corresponding to a primary function of the Fabric gateway service. Requests and responses use the same protobuf message types as the gRPC service, marshaled into a binary byte array for the CoAP payload.
+The CoAP gateway exposes five main endpoints corresponding to the Fabric Gateway service. All paths follow the pattern `/rpc/gateway.Gateway/{Method}` and are automatically generated as constants in `fabric-protos-go-apiv2`.
+
+**Generated Path Constants** (in `gateway_coap.pb.go`):
+
+```go
+const (
+    Gateway_Endorse_CoAPPath      = "/rpc/gateway.Gateway/Endorse"
+    Gateway_Submit_CoAPPath       = "/rpc/gateway.Gateway/Submit"
+    Gateway_CommitStatus_CoAPPath = "/rpc/gateway.Gateway/CommitStatus"
+    Gateway_Evaluate_CoAPPath     = "/rpc/gateway.Gateway/Evaluate"
+    Gateway_ChaincodeEvents_CoAPPath = "/rpc/gateway.Gateway/ChaincodeEvents"
+)
+```
+
+Requests and responses use the same protobuf message types as the gRPC service, marshaled into binary payloads.
 
 ### 1. Evaluate Transaction
 
 Used to invoke a transaction function on a peer to query the ledger state without submitting a transaction for ordering.
 
-* **Endpoint**: `POST /evaluate`
+* **Endpoint**: `POST /rpc/gateway.Gateway/Evaluate`
 * **Request Payload**: The body of the CoAP request must contain a marshaled `gateway.EvaluateRequest` protobuf message.
 * **Success Response**:
   * **Code**: `2.05 Content`
@@ -141,7 +214,7 @@ Used to invoke a transaction function on a peer to query the ledger state withou
 
 Used to collect endorsements for a transaction from peers sufficient to satisfy the chaincode's endorsement policy.
 
-* **Endpoint**: `POST /endorse`
+* **Endpoint**: `POST /rpc/gateway.Gateway/Endorse`
 * **Request Payload**: The body of the CoAP request must contain a marshaled `gateway.EndorseRequest` protobuf message.
 * **Success Response**:
   * **Code**: `2.05 Content`
@@ -151,19 +224,19 @@ Used to collect endorsements for a transaction from peers sufficient to satisfy 
 
 Used to submit a previously endorsed and client-signed transaction to the ordering service.
 
-* **Endpoint**: `POST /submit`
+* **Endpoint**: `POST /rpc/gateway.Gateway/Submit`
 * **Request Payload**: The body of the CoAP request must contain a marshaled `gateway.SubmitRequest` protobuf message.
 * **Success Response**:
   * **Code**: `2.04 Changed`
   * **Payload**: The response body contains a marshaled `gateway.SubmitResponse` protobuf message.
 
-**Note**: The `/submit` endpoint waits for the transaction to be submitted to the ordering service before returning. To check whether the transaction has been committed to the ledger, clients must use the `/commit-status` endpoint.
+**Note**: The `/rpc/gateway.Gateway/Submit` endpoint waits for the transaction to be submitted to the ordering service before returning. To check whether the transaction has been committed to the ledger, clients must use the `/rpc/gateway.Gateway/CommitStatus` endpoint.
 
 ### 4. Receive Ledger Events
 
 Used to subscribe to a stream of chaincode events from the ledger. This endpoint uses the CoAP Observe option (RFC 7641) to deliver a stream of notifications to the client.
 
-* **Endpoint**: `GET /events` (with the `Observe` option set to `0`)
+* **Endpoint**: `GET /rpc/gateway.Gateway/ChaincodeEvents` (with the `Observe` option set to `0`)
 * **Request Payload**: The body of the initial CoAP GET request must contain a marshaled `gateway.ChaincodeEventsRequest` protobuf message.
 * **Response Stream**:
   * The server will register the client as an observer and send a success response (`2.05 Content`).
@@ -173,11 +246,64 @@ Used to subscribe to a stream of chaincode events from the ledger. This endpoint
 
 Used to check the status of a previously submitted transaction.
 
-* **Endpoint**: `POST /commit-status`
+* **Endpoint**: `POST /rpc/gateway.Gateway/CommitStatus`
 * **Request Payload**: The body of the CoAP request must contain a marshaled `gateway.SignedCommitStatusRequest` protobuf message.
 * **Success Response**:
   * **Code**: `2.05 Content`
   * **Payload**: The response body contains a marshaled `gateway.CommitStatusResponse` protobuf message.
+
+## Developer Experience
+
+### Client-Side Development
+
+External developers working with the CoAP gateway get the same ergonomic experience as gRPC:
+
+#### **Step 1: Import generated client**
+
+  ```go
+  import "github.com/hyperledger/fabric-protos-go-apiv2/gateway"
+  ```
+
+#### **Step 2: Create connection and client**
+
+  ```go
+  conn, err := coap.NewDTLSConnection("fabric-gateway.example.com:5684",
+      coap.WithDTLSCert(transportCert, transportKey),  // DTLS certificate
+      coap.WithRootCA(caCert))
+  if err != nil {
+      log.Fatal(err)
+  }
+
+  client := gateway.NewGatewayCoapClient(conn)
+  ```
+
+#### **Step 3: Use type-safe methods**
+
+  ```go
+  // Full IDE autocomplete and type safety
+  endorseResp, err := client.Endorse(ctx, &gateway.EndorseRequest{
+      TransactionId: txID,
+      ChannelId:     "mychannel",
+      ProposedTransaction: signedProposal,  // Signed with identity certificate
+  })
+  ```
+
+### Server-Side Development
+
+For Fabric core developers, the CoAP gateway requires minimal implementation:
+
+1. **No changes** to existing `GatewayServer` implementation
+2. **Thin adapter layer** in `fabric/internal/peer/gateway/coap/`
+3. **Standard patterns** for all handlers (unmarshal → call gRPC → marshal)
+4. **Reuse** existing tests and validation logic
+
+When the Gateway service adds new methods:
+
+* Protocol buffer definition updated
+* Run `make genprotos` in fabric-protos
+* `gateway_coap.pb.go` automatically regenerates
+* Add simple handler in adapter following existing pattern
+* External clients automatically get new methods
 
 ## Error Handling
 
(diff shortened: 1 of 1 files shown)


bernardo-rf:
@yacovm @denyeart  @manish-sethi  @tock-ibm  @ale-linux @C0rWin  @andrew-coleman  @bestbeforetoday  @satota2 @pfi79

This is the proper PR related to the RFC proposes adding a CoAP gateway to Fabric for IoT device integration, with the proper checks passed. Please review and provide feedback.

pfi79:
Thank you for your proposes.
I have a few questions.

1. Will you reuse proto files from the fabric-protos project or will you create your own message types?

2. I am interested in how the transaction will be processed by the client (IoT)? It is on the client that the transaction id is generated, the time of the transaction, the nonce, and the transaction is signed. Will you leave it as it is now or change it?

3. You may have already tried something in the form of go code. Could you please provide some parts of the code as an example.

4. In your proposes, you refer to the Fabric SDK. This repository is archived and is not recommended for use. You need to use the fabric-gateway and link to it.

bestbeforetoday:
The peer Gateway service provides a CommitStatus method in addition to the Evaluate, Endorse, Submit and ChaincodeEvents methods that you have modelled. This is used to check the eventual status of a previously submitted transaction. See the [protocol buffer definition](https://hyperledger.github.io/fabric-protos/protos.html#gateway-Gateway) and this [overview of the Gateway transaction flow](https://github.com/hyperledger/fabric-samples/blob/main/full-stack-asset-transfer-guide/docs/ApplicationDev/01-FabricGateway.md). From the RFC description, it sounds like your submit is doing both the Submit and CommitStatus operations.

bernardo-rf:
@pfi79 Thank you for your questions. Here are the answers.

> Thank you for your proposes. I have a few questions.
> 1. Will you reuse proto files from the fabric-protos project or will you create your own message types?

Yes, the CoAP gateway will reuse the existing proto files from the fabric-protos project. This ensures that the CoAP gateway maintains full compatibility with the existing Fabric gateway ecosystem.

This is stated in the RFC:
_"Requests and responses use the same protobuf message types as the gRPC service, marshaled into a binary byte array for the CoAP payload."_

The RFC specifically mentions using:
- `gateway.EvaluateRequest` and `gateway.EvaluateResponse`
- `gateway.EndorseRequest` and `gateway.EndorseResponse`
- `gateway.SubmitRequest`
- `gateway.ChaincodeEventsRequest` and `gateway.ChaincodeEventsResponse`

The CoAP gateway will unmarshal incoming CoAP payloads into these standard protobuf messages, process them through the existing gateway logic, and marshal responses back to CoAP format.
This approach ensures consistency with the existing gRPC gateway and avoids creating duplicate message definitions. The CoAP gateway acts as a protocol converter, translating CoAP requests into the same protobuf structures used by the gRPC gateway.


> 2. I am interested in how the transaction will be processed by the client (IoT)? It is on the client that the transaction id is generated, the time of the transaction, the nonce, and the transaction is signed. Will you leave it as it is now or change it?

The current transaction processing model will be maintained. The RFC states:
_"The client is responsible for packaging all necessary information, such as the channel, chaincode name, and function arguments, within the Protobuf payload, just as they would when using the gRPC gateway directly."_

And in the cold chain example:
_"It extracts the standard Fabric transaction proposal, signed with the device's identity."_

This means:
- Transaction ID generation: Remains on the client side
- Transaction timestamp: Handled by the client
- Nonce generation: Client responsibility
- Transaction signing: The device signs the transaction proposal with its identity

The IoT device will use the same transaction preparation logic as any other Fabric client, creating properly formatted protobuf messages that the CoAP gateway can directly forward to the existing Fabric infrastructure.

The CoAP gateway simply acts forwarding the already-prepared and signed protobuf messages to the existing Fabric infrastructure. This maintains the security model where IoT devices are treated as first-class Fabric identities with their own certificates and signing capabilities, ensuring no degradation of Fabric's security properties.

> 3. You may have already tried something in the form of go code. Could you please provide some parts of the code as an example.

While the RFC doesn't include specific code examples, I can provide a conceptual example based on the design described. This represents the planned implementation approach rather than existing code.

Here's what the implementation would look like using the existing Fabric protobuf definitions
```
// Example CoAP handler for endorsement
func (h *EndorseHandler) HandleEndorse(w coap.ResponseWriter, req *coap.Request) {
    // Extract protobuf payload from CoAP request
    payload := req.Body()
    
    // Unmarshal the standard Fabric gateway.EndorseRequest
    var endorseReq gateway.EndorseRequest
    if err := proto.Unmarshal(payload, &endorseReq); err != nil {
        w.SetCode(coap.BadRequest)
        w.Write([]byte("Invalid protobuf payload"))
        return
    }
    
    // Authenticate client using DTLS certificate
    clientCert := req.ClientCertificate()
    identity, err := h.authenticator.ValidateCertificate(clientCert)
    if err != nil {
        w.SetCode(coap.Unauthorized)
        return
    }
    
    // Forward to existing gRPC gateway logic
    response, err := h.gatewayService.Endorse(context.Background(), &endorseReq)
    if err != nil {
        w.SetCode(coap.InternalServerError)
        return
    }
    
    // Marshal response back to protobuf
    responseBytes, _ := proto.Marshal(response)
    w.SetCode(coap.Content)
    w.Write(responseBytes)
}


// Example server setup
func NewCoAPGateway(config *Config) *CoAPGateway {
    server := coap.NewServer()
    
    // Register handlers for each endpoint
    server.Handle("/endorse", &EndorseHandler{gatewayService: config.GatewayService})
    server.Handle("/submit", &SubmitHandler{gatewayService: config.GatewayService})
    server.Handle("/commit-status", &CommitStatusHandler{gatewayService: config.GatewayService})
    server.Handle("/evaluate", &EvaluateHandler{gatewayService: config.GatewayService})
    server.Handle("/events", &EventsHandler{gatewayService: config.GatewayService})
    
    return &CoAPGateway{server: server}
}
```
This implementation demonstrates how the CoAP gateway would integrate with existing Fabric infrastructure while providing the lightweight protocol interface needed for IoT devices.

> 4. In your proposes, you refer to the Fabric SDK. This repository is archived and is not recommended for use. You need to use the fabric-gateway and link to it.

 You raise an excellent point about the Fabric SDK being archived. Let me clarify the context of these references in the RFC.

The RFC uses "Fabric SDK" in two specific contexts, neither of which recommends its use:
Historical Problem Description (Line 51):
_"Historically, this required... use a Fabric SDK to submit a transaction"_
This just describes a previous approach before the CoAP gateway solution

Conceptual Analogy (Line 57):
_"Think of it as a specialized application client, like one built with the Fabric SDK, but one that speaks CoAP instead of gRPC"_
This uses Fabric SDK as a familiar reference point to help developers understand the new concept

The RFC correctly focuses on the current `fabric-gateway` in all technical sections:
- Uses `gateway.EvaluateRequest`, `gateway.EndorseRequest`, etc. (current protobuf definitions)
- Integrates with existing gRPC gateway infrastructure
- References the current gateway architecture

If you prefer, we could update these references to use more generic terms like "Fabric client library" or "existing Fabric client" or "Fabric Gateway" to avoid any confusion about the deprecated SDK. However, the current references serve their intended purpose of explaining historical context and providing conceptual clarity without recommending the use of deprecated libraries.

bernardo-rf:
@bestbeforetoday Thank you for your question. Here is the answer.

> The peer Gateway service provides a CommitStatus method in addition to the Evaluate, Endorse, Submit and ChaincodeEvents methods that you have modelled. This is used to check the eventual status of a previously submitted transaction. See the [protocol buffer definition](https://hyperledger.github.io/fabric-protos/protos.html#gateway-Gateway) and this [overview of the Gateway transaction flow](https://github.com/hyperledger/fabric-samples/blob/main/full-stack-asset-transfer-guide/docs/ApplicationDev/01-FabricGateway.md). From the RFC description, it sounds like your submit is doing both the Submit and CommitStatus operations.

You're correct! The current Fabric Gateway service includes a `CommitStatus` method that is not yet addressed in the CoAP gateway design.

To ensure the CoAP gateway provides the same functionality as the existing gRPC gateway, I propose adding a fifth endpoint to the RFC.

The current CoAP gateway RFC models four endpoints:
- `/evaluate` - for evaluation requests
- `/endorse` - for managing endorsements
- `/submit` - for processing transaction submissions
- `/events` - for listening to events

My proposal is to update the RFC with the following changes:

1. Add a `/commit-status` endpoint that accepts a `gateway.CommitStatusRequest` and returns a `gateway.CommitStatusResponse`.
2. Clarify the behavior of the `/submit` endpoint. The RFC should explicitly state whether /submit returns a result upon successful submission to the orderer or waits for the transaction to be committed to the ledger.
3. Update the API documentation to reflect the complete set of gateway methods, ensuring parity with the current gRPC gateway.

This would enable IoT devices to both submit transactions and check their status.

bernardo-rf:
Following discussions, I have created a presentation that summarizes the goals and design of this RFC.

I am sharing the presentation material on the Hyperledger Fabric mailing list to kick off wider community discussion.

Please find the mailing list thread here: [Link to the Mailing List Announcement](https://lists.lfdecentralizedtrust.org/g/fabric/topic/rfc_coap_gateway_for/115147787)

pfi79:
> Following discussions, I have created a presentation that summarizes the goals and design of this RFC.
> 
> I am sharing the presentation material on the Hyperledger Fabric mailing list to kick off wider community discussion.
> 
> Please find the mailing list thread here: [Link to the Mailing List Announcement](https://lists.lfdecentralizedtrust.org/g/fabric/topic/rfc_coap_gateway_for/115147787)

Thank you for your presentation.
But let me make a few comments. 

1. It seems that on the slide "Security Model: DTLS + Fabric Identity" you have a confusion.
The fact is that the "client" has two private keys and 2 certificates, respectively. the first is for Identity, the second is for TLS. If you try to get an Identity from a certificate intended for TLS, you will most likely get an error. Total: the certificate with which you will pass the "DTLS handshake" is needed only for this and that's it. But inside the SignedProposal lies the Identity and it is through it that access to resources is made.
2. If you're going to reuse proto files, it's probably better to do so. To develop a plug-in to `protoc` `protocol-gen-grpc-coap`, which would generate code for the client and server using proto files. 
Take a look at the `protocol-gen-grpc-gateway` as an example. This plugin makes it easy to add a RestAPI endpoint (if needed).
That is, one of the important components is the possibility of re-generating the coap code in case the proto file is changed.

bernardo-rf:
Hi @pfi79 ,

Thank you for your review. Your comments regarding the separation of security identities and the long-term maintenance strategy can significantly strengthened this proposal.

I agree that the best path forward involves adopting two key architectural changes to ensure the CoAP Gateway adheres to Hyperledger Fabric's security and maintenance standards.

I have drafted the specific changes below and plan to commit them to the RFC. **Could you confirm if this proposed resolution addresses your concerns and if you are aligned with this direction before I finalize the updated RFC text?**

---

### 1. Resolution for Security Model Confusion (DTLS vs. Fabric Identity)

You have a point regarding separating the transport identity from the signing identity is the most secure and robust approach. We will be moving from a unified certificate model to a **Dual-Certificate Model**.

This addresses the security separation and clarifies the two-stage authentication process within the gateway:

#### Proposed New Flow:

* **Device Provisioning:** The device will be provisioned with two distinct X.509 certificate/key pairs (both issued by a Fabric CA):
    * **Transport Certificate (DTLS):** Used only for the mutual DTLS handshake (connection authentication/confidentiality).
    * **Fabric Identity Certificate (Signing):** Used exclusively to sign the `SignedProposal` payload (transaction authorization).

* **Two-Stage Validation in the Gateway:**
    * **Stage 1: Transport Authentication:** The DTLS certificate is validated for connection access.
    * **Stage 2: Transaction Authorization (Crucial):** The gateway extracts and verifies the signature of the **Fabric Signing Certificate** found within the `SignedProposal` payload. All Fabric ACLs and endorsement policies will be enforced against *this* identity.

#### Key Updates to RFC Text:
The "Guide-level explanation," "Protocol and Security," and "Client Authentication" sections will be updated to explicitly describe this two-stage process and the requirement for two certificates.



---

### 2. Resolution for Protocol Generation and Maintainability

I agree that a manual implementation for Protobuf message handling creates an unacceptable maintenance burden.

#### Proposed Solution: `protoc` Generation Utility

We will introduce a **code generation utility** to ensure the CoAP API remains synchronized with the upstream Gateway Protobuf definitions.

* **Implementation Plan:** We will prioritize the development of this utility in **Phase 1: Core Implementation**.
    > *Develop the `protoc` generation utility to establish the API structure* before implementing the core handlers.

* **Detailed Design:** The RFC will be updated to include the **Protocol Generator Utility** as a core component, ensuring that CoAP request/response marshalling/unmarshalling boilerplate is automatically generated.

---

These changes are designed to make the CoAP Gateway proposal both secure and sustainable for the project. Please let me know if these resolutions meet your point of view. I will hold off on committing the final RFC text until I receive your confirmation.

Thank you again for your time.

pfi79:
Yes, I agree with what you wrote.

> ### 1. Resolution for Security Model Confusion (DTLS vs. Fabric Identity)
> You have a point regarding separating the transport identity from the signing identity is the most secure and robust approach. We will be moving from a unified certificate model to a **Dual-Certificate Model**.
> 
> This addresses the security separation and clarifies the two-stage authentication process within the gateway:
> 
> #### Proposed New Flow:
> * **Device Provisioning:** The device will be provisioned with two distinct X.509 certificate/key pairs (both issued by a Fabric CA):
>   
>   * **Transport Certificate (DTLS):** Used only for the mutual DTLS handshake (connection authentication/confidentiality).
>   * **Fabric Identity Certificate (Signing):** Used exclusively to sign the `SignedProposal` payload (transaction authorization).
> * **Two-Stage Validation in the Gateway:**
>   
>   * **Stage 1: Transport Authentication:** The DTLS certificate is validated for connection access.
>   * **Stage 2: Transaction Authorization (Crucial):** The gateway extracts and verifies the signature of the **Fabric Signing Certificate** found within the `SignedProposal` payload. All Fabric ACLs and endorsement policies will be enforced against _this_ identity.
> 
> #### Key Updates to RFC Text:
> The "Guide-level explanation," "Protocol and Security," and "Client Authentication" sections will be updated to explicitly describe this two-stage process and the requirement for two certificates.

In fact, leads what you came up with to what is in fabric. Therefore, it minimizes changes.

> ### 2. Resolution for Protocol Generation and Maintainability
> I agree that a manual implementation for Protobuf message handling creates an unacceptable maintenance burden.
> 
> #### Proposed Solution: `protoc` Generation Utility
> We will introduce a **code generation utility** to ensure the CoAP API remains synchronized with the upstream Gateway Protobuf definitions.
> 
> * **Implementation Plan:** We will prioritize the development of this utility in **Phase 1: Core Implementation**.
>   > _Develop the `protoc` generation utility to establish the API structure_ before implementing the core handlers.
> * **Detailed Design:** The RFC will be updated to include the **Protocol Generator Utility** as a core component, ensuring that CoAP request/response marshalling/unmarshalling boilerplate is automatically generated.

I agree with almost everything. But most likely, you will not have a choice of the order of "file generation". Because the files are generated in the `fabric-protos` repository, from which the changes will be transferred to `fabric-protos-go-apiv2`. And only then will they appear in `fabric`.

I understand that creating a utility that would work with `protoc` is long and expensive. But you can come up with something simpler, if only you could run `fabric-protos` in the repository.

bernardo-rf:
> I agree with almost everything. But most likely, you will not have a choice of the order of "file generation". Because the files are generated in the fabric-protos repository, from which the changes will be transferred to fabric-protos-go-apiv2. And only then will they appear in fabric.

>I understand that creating a utility that would work with protoc is long and expensive. But you can come up with something simpler, if only you could run fabric-protos in the repository.

@pfi79 I completely agree with the concern regarding the file generation order and the maintenance overhead of a custom `protoc plugin`. After reviewing the repository lifecycle (`fabric-protos` -> `fabric-protos-go-apiv2` -> `fabric`), I've moved away from the plugin-based generation approach.

Instead, I am proposing a **Descriptor-Driven Approach**. This approach leverages the protobuf descriptors that are already compiled into the `fabric-protos-go-apiv2` module used by the `Peer`.

How it solves the maintenance issue:
- We don't need to add new generation steps to `fabric-protos`. The CoAP server lives entirely within the Peer and consumes the official Go bindings.
- The CoAP server uses **Go’s protoreflect API** to **discover the Gateway service's methods** and types at runtime from the embedded descriptors.
- If a new RPC is added to the Gateway service, the CoAP server will automatically support it once the dependency is updated. We avoid the utility creation while ensuring the CoAP server is always in perfect sync with the proto definitions.

---------

## Why this approach ?

descriptor-driven-approach-flow

### Minimal Maintenance Burden
- Schema coupling is explicit: we track the same `fabric-protos-go-apiv2` version Fabric uses
- Changes to Gateway proto are caught at compile-time, not runtime

### Generic Implementation
- Uses `dynamicpb` for generic message handling
- Same code handles all Gateway RPCs without hard-coding types
- Easy to extend to new methods without code changes

### Compile-Time Safety
- Breaking changes in Gateway proto break the build, not production
- Clear dependency on `fabric-protos-go-apiv2` version
- Can fail fast at startup if schema mismatches are detected

This satisfies the need for a '**simpler**' solution that doesn't break the existing build flow while being essentially maintenance-free as the API evolves.

pfi79:
> > I agree with almost everything. But most likely, you will not have a choice of the order of "file generation". Because the files are generated in the fabric-protos repository, from which the changes will be transferred to fabric-protos-go-apiv2. And only then will they appear in fabric.
> 
> > I understand that creating a utility that would work with protoc is long and expensive. But you can come up with something simpler, if only you could run fabric-protos in the repository.
> 
> @pfi79 I completely agree with the concern regarding the file generation order and the maintenance overhead of a custom `protoc plugin`. After reviewing the repository lifecycle (`fabric-protos` -> `fabric-protos-go-apiv2` -> `fabric`), I've moved away from the plugin-based generation approach.
> 
> Instead, I am proposing a **Descriptor-Driven Approach**. This approach leverages the protobuf descriptors that are already compiled into the `fabric-protos-go-apiv2` module used by the `Peer`.
> 
> How it solves the maintenance issue:
> 
> * We don't need to add new generation steps to `fabric-protos`. The CoAP server lives entirely within the Peer and consumes the official Go bindings.
> * The CoAP server uses **Go’s protoreflect API** to **discover the Gateway service's methods** and types at runtime from the embedded descriptors.
> * If a new RPC is added to the Gateway service, the CoAP server will automatically support it once the dependency is updated. We avoid the utility creation while ensuring the CoAP server is always in perfect sync with the proto definitions.
> 
> ## Why this approach ?
> descriptor-driven-approach-flow
> ### Minimal Maintenance Burden
> * Schema coupling is explicit: we track the same `fabric-protos-go-apiv2` version Fabric uses
> * Changes to Gateway proto are caught at compile-time, not runtime
> 
> ### Generic Implementation
> * Uses `dynamicpb` for generic message handling
> * Same code handles all Gateway RPCs without hard-coding types
> * Easy to extend to new methods without code changes
> 
> ### Compile-Time Safety
> * Breaking changes in Gateway proto break the build, not production
> * Clear dependency on `fabric-protos-go-apiv2` version
> * Can fail fast at startup if schema mismatches are detected
> 
> This satisfies the need for a '**simpler**' solution that doesn't break the existing build flow while being essentially maintenance-free as the API evolves.

Thank you, and by and large, I support your initiative.

There are some minor nuances left for me.

When I suggested considering generation at the fabric-protos level, I was also thinking about generating functions for the client. At least for tests, and at the maximum for easier creation of clients to fabric via CoAP.

bernardo-rf:
> Thank you, and by and large, I support your initiative.
> 
> There are some minor nuances left for me.
> 
> When I suggested considering generation at the fabric-protos level, I was also thinking about generating functions for the client. At least for tests, and at the maximum for easier creation of clients to fabric via CoAP.

@pfi79  Thank you for the feedback. To address the concerns regarding the need for easy client/test creation, I am proposing a **Descriptor-Driven Approach** as the primary solution, **complemented** by a **Two-Layer Client Architecture** that lives entirely within the `fabric` repository.

This bypasses the `fabric-protos` generation bottleneck while delivering the "helper functions" you suggested.

### Generic Descriptor-Driven Client

The core client uses the `FileDescriptor` already embedded in `fabric-protos-go-apiv2`. It discovers methods and types at runtime. It uses the descriptors to handle URI paths and marshaling generically. This is what ensures the system never breaks when protos change.

```Go
// coapgateway/inner_client.go
func (c *GenericClient) InvokeUnary(ctx context.Context, methodName string, req, resp proto.Message) error {
    // 1. Look up method info from the embedded descriptors (not hardcoded)
    methodInfo, err := c.registry.GetMethod(methodName) 
    if err != nil { return err }

    // 2. The path is built dynamically (e.g., /rpc/gateway.Gateway/Endorse)
    path := fmt.Sprintf("/rpc/%s/%s", c.registry.ServiceName, methodName)

    // 3. Marshal/Send/Unmarshal using the standard proto library
    payload, _ := proto.Marshal(req)
    coapResp, err := c.conn.Post(ctx, path, message.AppOctetStream, payload)
    
    // ... error handling and unmarshaling ...
    return proto.Unmarshal(coapResp.Body(), resp)
}
```

### High-Level Type-Safe Wrapper

To fulfill the goal of making it easy for clients and tests to use CoAP, I’ve added a thin wrapper that provides the specific functions you mentioned (e.g., Endorse, Submit).

```Go
// coapgateway/client.go - The high-level wrapper for users/tests
type GatewayClient struct {
    inner *GenericClient // The descriptor-driven engine
}

// Named functions provide the "generated" feel without the generator complexity
func (c *GatewayClient) Endorse(ctx context.Context, req *gw.EndorseRequest) (*gw.EndorseResponse, error) {
    resp := &gw.EndorseResponse{}
    // inner.InvokeUnary handles the URI pathing and marshaling generically
    err := c.inner.InvokeUnary(ctx, "Endorse", req, resp)
    return resp, err
}

func (c *GatewayClient) Submit(ctx context.Context, req *gw.SubmitRequest) (*gw.SubmitResponse, error) {
    resp := &gw.SubmitResponse{}
    err := c.inner.InvokeUnary(ctx, "Submit", req, resp)
    return resp, err
}
```
Because this wrapper is 'logic-less', it only needs to be updated if a brand-new RPC method is added to the Gateway. Given the stability of the Gateway API, this is a much smaller burden than maintaining a custom `protoc` plugin and modifying the `fabric-protos` build pipeline.

- **For Tests:** Developers get a standard `client.Endorse(ctx, req)` experience. It is type-safe, supports autocomplete, and feels like the existing SDK.

- **For Easier Client Creation:** External Go developers can simply import this client package to interact with the Peer via CoAP without ever seeing a 'raw' CoAP packet or URI string.

The primary focus remains the Descriptor-Driven Approach, providing a maintenance-free server. The two-layer client is a complementary addition to ensure a quality developer experience.

Does this approach satisfy both the architectural maintenance requirements and the need for a high-quality client/test DX?

pfi79:
> Because this wrapper is 'logic-less', it only needs to be updated if a brand-new RPC method is added to the Gateway. Given the stability of the Gateway API, this is a much smaller burden than maintaining a custom `protoc` plugin and modifying the `fabric-protos` build pipeline.
> 
> * **For Tests:** Developers get a standard `client.Endorse(ctx, req)` experience. It is type-safe, supports autocomplete, and feels like the existing SDK.
> * **For Easier Client Creation:** External Go developers can simply import this client package to interact with the Peer via CoAP without ever seeing a 'raw' CoAP packet or URI string.
> 
> The primary focus remains the Descriptor-Driven Approach, providing a maintenance-free server. The two-layer client is a complementary addition to ensure a quality developer experience.
> 
> Does this approach satisfy both the architectural maintenance requirements and the need for a high-quality client/test DX?

The fact is that the client should not import packages from the fabric repository. 
There are fabric-lib-go and fabric-protos-go-apiv2 repositories for this.

If there is even a small chance sometime in the future that a new method will be added to Gateway, I kindly ask you to consider your solution for this option.

bernardo-rf:
@pfi79  After reflecting on your feedback and the broader adoption goals, it's clear that providing an easy, proto-based client generation option is essential for the success of this feature.

Without automated client generation, we risk:
- Developers won't embrace a protocol requiring manual client implementation
- Gateway API changes would require coordinated manual updates across multiple repositories
- Manual implementations may diverge in subtle ways across different clients
- Compared to gRPC's auto-generated clients, a manual CoAP approach would be a significant step backward

Given these considerations, I propose revisiting the **protoc Generation Utility approach**, specifically: a code generation plugin that integrates directly into the `fabric-protos` build pipeline, similar to how `protoc-gen-go-grpc` currently works.


### Proposed Solution: `protoc-gen-go-coap` Plugin
Rather than creating a complex standalone tool, I propose developing a protoc plugin (`protoc-gen-go-coap`) that slots seamlessly into the existing `fabric-protos` generation workflow.

**1. Works Within Existing Infrastructure**
The `fabric-protos` repository already orchestrates multiple protoc plugins via `buf.gen.yaml`. Adding one more entry is straightforward:

 ```yaml
   plugins:
     - local: protoc-gen-go-grpc
       out: bindings/go-apiv2
       opt:
         - paths=source_relative
         - require_unimplemented_servers=false
     - local: protoc-gen-go-coap      # <-- New plugin
       out: bindings/go-apiv2
       opt:
         - paths=source_relative
``` 

**2. Minimal Development Complexity**
Leveraging Google's protogen library means we don't need to parse proto files manually the heavy lifting is already done. The plugin focuses solely on generating Go client code and path constants.

**3. Automatic Integration**
Once configured, `make genprotos` automatically generates `gateway_coap.pb.go` alongside `gateway_grpc.pb.go`. No additional manual steps required.

**4. Natural Pipeline Flow**
Generated files appear in `bindings/go-apiv2/gateway/` during the `fabric-protos` build, then flow naturally to `fabric-protos-go-apiv2 through your existing release process.

### Generated Output Example
For `gateway.proto`, the plugin generates `gateway_coap.pb.go:

```Go
package gateway

// Path constants for CoAP URIs
const (
    Gateway_Endorse_CoAPPath      = "/rpc/gateway.Gateway/Endorse"
    Gateway_Submit_CoAPPath       = "/rpc/gateway.Gateway/Submit"
    Gateway_CommitStatus_CoAPPath = "/rpc/gateway.Gateway/CommitStatus"
    Gateway_Evaluate_CoAPPath     = "/rpc/gateway.Gateway/Evaluate"
)

// GatewayCoapClient is the CoAP client API for Gateway service
type GatewayCoapClient interface {
    Endorse(ctx context.Context, in *EndorseRequest, opts ...CoapOption) (*EndorseResponse, error)
    Submit(ctx context.Context, in *SubmitRequest, opts ...CoapOption) (*SubmitResponse, error)
    CommitStatus(ctx context.Context, in *SignedCommitStatusRequest, opts ...CoapOption) (*CommitStatusResponse, error)
    Evaluate(ctx context.Context, in *EvaluateRequest, opts ...CoapOption) (*EvaluateResponse, error)
}

// Constructor and implementation
func NewGatewayCoapClient(conn CoapConnection) GatewayCoapClient { /* ... */ }

// Internal implementation struct and methods...
```

### Client-Side Benefits Output Example

External client usage becomes trivial:
```Go
import "github.com/hyperledger/fabric-protos-go-apiv2/gateway"

client := gateway.NewGatewayCoapClient(coapConn)

resp, err := client.Endorse(ctx, &gateway.EndorseRequest{
    TransactionId: "tx123",
    ChannelId:     "mychannel",
    // ...
})
``` 

This approach is directly consistent with:
- Follows the same pattern as gRPC, Java, and Node.js bindings already present in `fabric-protos
- External clients importing `fabric-protos-go-apiv2/gateway` get full type-safe client methods with IDE autocomplete
- When Gateway service adds new methods, regeneration is automatic
- Matches the ergonomics developers expect from gRPC clients


### Server-Side Benefits Output Example

With the protoc generation approach, the server-side implementation becomes simple: a lightweight **CoAP adapter that lives within the peer process** and **proxies to the existing gRPC Gateway** implementation.


Where the code lives:
```
fabric/
├── internal/peer/gateway/
│   ├── gateway.go              # Existing gRPC GatewayServer (UNCHANGED)
│   ├── endorser.go             # Existing business logic (UNCHANGED)
│   └── coap/
│       ├── adapter.go          # New: CoAP-to-gRPC adapter
│       └── server.go           # New: CoAP server initialization
└── core/peer/
    └── peer.go                 # Peer startup: add CoAP listener

fabric-protos/
└── bindings/go-apiv2/gateway/
    ├── gateway.pb.go           # Existing
    ├── gateway_grpc.pb.go      # Existing  
    └── gateway_coap.pb.go      # New: Generated constants & client
```

All existing endorsement, ACL enforcement, and transaction validation logic remains untouched in the current `GatewayServer implementation.

Each handler follows the same straightforward pattern:
```Go
// Every handler follows this simple pattern
func (a *CoAPAdapter) handle{Method}(w coap.ResponseWriter, req *coap.Request) {
    // 1. Unmarshal CoAP request
    request := &gateway.{Method}Request{}
    proto.Unmarshal(req.Body(), request)
    
    // 2. Call existing gRPC implementation (all business logic here)
    response, err := a.grpcGateway.{Method}(req.Context(), request)
    
    // 3. Marshal CoAP response
    payload, _ := proto.Marshal(response)
    w.Write(payload)
}
```

### Summary

By generating both client methods AND path constants in `fabric-protos-go-apiv2`, we eliminate manual maintenance burden on both sides while keeping the fabric peer implementation trivial. The adapter living inside the peer is actually an advantage it's a thin translation layer with zero network overhead, direct access to the `GatewayServer` instance, and seamless authentication/authorization context sharing.

pfi79:
> @pfi79 After reflecting on your feedback and the broader adoption goals, it's clear that providing an easy, proto-based client generation option is essential for the success of this feature.
> 
> Without automated client generation, we risk:
> 
> * Developers won't embrace a protocol requiring manual client implementation
> * Gateway API changes would require coordinated manual updates across multiple repositories
> * Manual implementations may diverge in subtle ways across different clients
> * Compared to gRPC's auto-generated clients, a manual CoAP approach would be a significant step backward
> 
> Given these considerations, I propose revisiting the **protoc Generation Utility approach**, specifically: a code generation plugin that integrates directly into the `fabric-protos` build pipeline, similar to how `protoc-gen-go-grpc` currently works.
> 
> ### Proposed Solution: `protoc-gen-go-coap` Plugin
> Rather than creating a complex standalone tool, I propose developing a protoc plugin (`protoc-gen-go-coap`) that slots seamlessly into the existing `fabric-protos` generation workflow.
> 
> **1. Works Within Existing Infrastructure** The `fabric-protos` repository already orchestrates multiple protoc plugins via `buf.gen.yaml`. Adding one more entry is straightforward:
> 
> ```yaml
>   plugins:
>     - local: protoc-gen-go-grpc
>       out: bindings/go-apiv2
>       opt:
>         - paths=source_relative
>         - require_unimplemented_servers=false
>     - local: protoc-gen-go-coap      # <-- New plugin
>       out: bindings/go-apiv2
>       opt:
>         - paths=source_relative
> ```
> 
> **2. Minimal Development Complexity** Leveraging Google's protogen library means we don't need to parse proto files manually the heavy lifting is already done. The plugin focuses solely on generating Go client code and path constants.
> 
> **3. Automatic Integration** Once configured, `make genprotos` automatically generates `gateway_coap.pb.go` alongside `gateway_grpc.pb.go`. No additional manual steps required.
> 
> **4. Natural Pipeline Flow** Generated files appear in `bindings/go-apiv2/gateway/` during the `fabric-protos` build, then flow naturally to `fabric-protos-go-apiv2 through your existing release process.
> 
> ### Generated Output Example
> For `gateway.proto`, the plugin generates `gateway_coap.pb.go:
> 
> ```go
> package gateway
> 
> // Path constants for CoAP URIs
> const (
>     Gateway_Endorse_CoAPPath      = "/rpc/gateway.Gateway/Endorse"
>     Gateway_Submit_CoAPPath       = "/rpc/gateway.Gateway/Submit"
>     Gateway_CommitStatus_CoAPPath = "/rpc/gateway.Gateway/CommitStatus"
>     Gateway_Evaluate_CoAPPath     = "/rpc/gateway.Gateway/Evaluate"
> )
> 
> // GatewayCoapClient is the CoAP client API for Gateway service
> type GatewayCoapClient interface {
>     Endorse(ctx context.Context, in *EndorseRequest, opts ...CoapOption) (*EndorseResponse, error)
>     Submit(ctx context.Context, in *SubmitRequest, opts ...CoapOption) (*SubmitResponse, error)
>     CommitStatus(ctx context.Context, in *SignedCommitStatusRequest, opts ...CoapOption) (*CommitStatusResponse, error)
>     Evaluate(ctx context.Context, in *EvaluateRequest, opts ...CoapOption) (*EvaluateResponse, error)
> }
> 
> // Constructor and implementation
> func NewGatewayCoapClient(conn CoapConnection) GatewayCoapClient { /* ... */ }
> 
> // Internal implementation struct and methods...
> ```
> 
> ### Client-Side Benefits Output Example
> External client usage becomes trivial:
> 
> ```go
> import "github.com/hyperledger/fabric-protos-go-apiv2/gateway"
> 
> client := gateway.NewGatewayCoapClient(coapConn)
> 
> resp, err := client.Endorse(ctx, &gateway.EndorseRequest{
>     TransactionId: "tx123",
>     ChannelId:     "mychannel",
>     // ...
> })
> ```
> 
> This approach is directly consistent with:
> 
> * Follows the same pattern as gRPC, Java, and Node.js bindings already present in `fabric-protos
> * External clients importing `fabric-protos-go-apiv2/gateway` get full type-safe client methods with IDE autocomplete
> * When Gateway service adds new methods, regeneration is automatic
> * Matches the ergonomics developers expect from gRPC clients
> 
> ### Server-Side Benefits Output Example
> With the protoc generation approach, the server-side implementation becomes simple: a lightweight **CoAP adapter that lives within the peer process** and **proxies to the existing gRPC Gateway** implementation.
> 
> Where the code lives:
> 
> ```
> fabric/
> ├── internal/peer/gateway/
> │   ├── gateway.go              # Existing gRPC GatewayServer (UNCHANGED)
> │   ├── endorser.go             # Existing business logic (UNCHANGED)
> │   └── coap/
> │       ├── adapter.go          # New: CoAP-to-gRPC adapter
> │       └── server.go           # New: CoAP server initialization
> └── core/peer/
>     └── peer.go                 # Peer startup: add CoAP listener
> 
> fabric-protos/
> └── bindings/go-apiv2/gateway/
>     ├── gateway.pb.go           # Existing
>     ├── gateway_grpc.pb.go      # Existing  
>     └── gateway_coap.pb.go      # New: Generated constants & client
> ```
> 
> All existing endorsement, ACL enforcement, and transaction validation logic remains untouched in the current `GatewayServer implementation.
> 
> Each handler follows the same straightforward pattern:
> 
> ```go
> // Every handler follows this simple pattern
> func (a *CoAPAdapter) handle{Method}(w coap.ResponseWriter, req *coap.Request) {
>     // 1. Unmarshal CoAP request
>     request := &gateway.{Method}Request{}
>     proto.Unmarshal(req.Body(), request)
>     
>     // 2. Call existing gRPC implementation (all business logic here)
>     response, err := a.grpcGateway.{Method}(req.Context(), request)
>     
>     // 3. Marshal CoAP response
>     payload, _ := proto.Marshal(response)
>     w.Write(payload)
> }
> ```
> 
> ### Summary
> By generating both client methods AND path constants in `fabric-protos-go-apiv2`, we eliminate manual maintenance burden on both sides while keeping the fabric peer implementation trivial. The adapter living inside the peer is actually an advantage it's a thin translation layer with zero network overhead, direct access to the `GatewayServer` instance, and seamless authentication/authorization context sharing.

I fully support you.
Please make appropriate changes to the document.

bernardo-rf:
## Technical Update: Standardized Generation & Security Model

Based on consensus with @pfi79 , I have pivoted the RFC to a **static generation model** and clarified the security handshake. These changes ensure the CoAP Gateway follows the existing Fabric toolchain without manual maintenance.

### 1. Toolchain Integration (`fabric-protos`)
- Replaced the previous approach with a lightweight `protoc-gen-go-coap` plugin.
- Bindings are generated in `fabric-protos` and flow through `apiv2` to fabric, matching the gRPC pattern.
- Zero manual updates required when `gateway.proto` changes.

### 2. Dual-Certificate Security
- Explicitly separated **Transport Identity (DTLS)** from **Signing Identity (MSP)**.
- This mirrors the standard pattern used in the gRPC gateway, ensuring no security regressions.

### 3. Implementation Details
- The gateway acts as a **thin adapter** with no business logic duplication; it delegates directly to the internal GatewayServer.

### Summary
- Automated sync with Protobuf definitions.
- Follows established Fabric patterns.
- Auto-generated client interfaces for IoT developers.

The RFC is updated and ready for review.

jt-nti:
Hi, I've just been catching up on the discussions related to fabric-protos. I like the idea of adding support for resource constrained devices, and extending the fabric-protos build pipeline to generate a coap client implementation seems really interesting but I don't think that the coap client code should be added to the existing `fabric-protos-go-apiv2` go bindings. 

The generated go, java and node bindings currently only contain what's defined in the protobuf definitions, which is a good thing and in my opinion should not change. I think it would be surprising for a project that only requires the go bindings to get a coap client implementation as well. (All the actual client implementations are in separate projects.)

Using a new protoc plugin in the fabric-protos build to generate a client implementation definitely sounds interesting but it probably only makes sense if the code can be 100% generated, which I think is the case based on...

> Zero manual updates required when gateway.proto changes.

If not, you're swapping having to manually update a client project, with maintaining a protoc plugin somewhere *and* having to manually update the client implementation anyway. (Mixing generated and hand written code can cause its own problems.)

pfi79 (approved):

pfi79:
> Hi, I've just been catching up on the discussions related to fabric-protos. I like the idea of adding support for resource constrained devices, and extending the fabric-protos build pipeline to generate a coap client implementation seems really interesting but I don't think that the coap client code should be added to the existing `fabric-protos-go-apiv2` go bindings.
> 
> The generated go, java and node bindings currently only contain what's defined in the protobuf definitions, which is a good thing and in my opinion should not change. I think it would be surprising for a project that only requires the go bindings to get a coap client implementation as well. (All the actual client implementations are in separate projects.)
> 
> Using a new protoc plugin in the fabric-protos build to generate a client implementation definitely sounds interesting but it probably only makes sense if the code can be 100% generated, which I think is the case based on...
> 
> > Zero manual updates required when gateway.proto changes.
> 
> If not, you're swapping having to manually update a client project, with maintaining a protoc plugin somewhere _and_ having to manually update the client implementation anyway. (Mixing generated and hand written code can cause its own problems.)

I may have misunderstood you.

fabric-protos-go-apev2 is used by both fabric and clients and libraries that implement changecodes.
That is, there is a single point at which changes in the proto protocol are immediately reflected in the code, for example on go.

If you're worried that there won't be such an implementation for other languages (node and java). There has already been such a precedent, and I am primarily interested in the implementation on go. In the future, it will be possible to open questions about the implementation of this in other languages.

If you are worried that the generation will take place directly in fabric-protos-go-apiv2. No way. All changes are only in fabric-protos. 

How do I make a plugin? There are options here. They can be discussed. Of course, I want it to be official as protocol-gen-go. But I think it won't be a big deal if the plugin is customized first.

bestbeforetoday:
Similar to James, I am nervous of adding CoAP bindings to fabric-protos-go-apiv2. This would force an unnecessary dependency on existing (gRPC-based) consumers: chaincode, admin SDK, client API and all the client applications that use those APIs. Could CoAP bindings (generated in fabric-protos from the same protobuf definitions) not be published to a sister repository for client consumption? For example, a fabric-coap-go repository.

pfi79:
> Similar to James, I am nervous of adding CoAP bindings to fabric-protos-go-apiv2. This would force an unnecessary dependency on existing (gRPC-based) consumers: chaincode, admin SDK, client API and all the client applications that use those APIs. Could CoAP bindings (generated in fabric-protos from the same protobuf definitions) not be published to a sister repository for client consumption? For example, a fabric-coap-go repository.

I don't understand why you're afraid of this. 

Could you explain in more detail?

pfi79:
> Similar to James, I am nervous of adding CoAP bindings to fabric-protos-go-apiv2. This would force an unnecessary dependency on existing (gRPC-based) consumers: chaincode, admin SDK, client API and all the client applications that use those APIs. Could CoAP bindings (generated in fabric-protos from the same protobuf definitions) not be published to a sister repository for client consumption? For example, a fabric-coap-go repository.

The thing is that the implementation of CoAP directly depends on proto files.
Do you want changes in fabric-protos to be written to the fabric-coap-go repository in the same way?
But the generated coap files will not work without grpc files.

Let me show you with a simpler example, I recently did an exercise (adding the rest api if we already have grpc)
How do I add an implementation? 
So (see the protocol-gen-grpc-gateway section):
```
version: v2
plugins:
  - local: protoc-gen-go
    out: bindings/go-apiv2
    opt: paths=source_relative
  - local: protoc-gen-go-grpc
    out: bindings/go-apiv2
    opt:
      - paths=source_relative
      - require_unimplemented_servers=false
  - local: protoc-gen-grpc-gateway
    out: bindings/go-apiv2
    opt:
      - paths=source_relative
      - generate_unbound_methods=true
```

Now let's look at the code that was generated - *.pb.gw.go. It directly calls functions from the *_grpc.pb.go file. As a rule, such files are located nearby.

For example, here is the code for peer/peer.proto (see `c.cc.Invoke`):

```
// Copyright the Hyperledger Fabric contributors. All rights reserved.
//
// SPDX-License-Identifier: Apache-2.0

// Code generated by protoc-gen-go-grpc. DO NOT EDIT.
// versions:
// - protoc-gen-go-grpc v1.5.1
// - protoc             (unknown)
// source: peer/peer.proto

package peer

import (
	context "context"
	grpc "google.golang.org/grpc"
	codes "google.golang.org/grpc/codes"
	status "google.golang.org/grpc/status"
)

// This is a compile-time assertion to ensure that this generated file
// is compatible with the grpc package it is being compiled against.
// Requires gRPC-Go v1.64.0 or later.
const _ = grpc.SupportPackageIsVersion9

const (
	Endorser_ProcessProposal_FullMethodName = "/protos.Endorser/ProcessProposal"
)

// EndorserClient is the client API for Endorser service.
//
// For semantics around ctx use and closing/ending streaming RPCs, please refer to https://pkg.go.dev/google.golang.org/grpc/?tab=doc#ClientConn.NewStream.
type EndorserClient interface {
	ProcessProposal(ctx context.Context, in *SignedProposal, opts ...grpc.CallOption) (*ProposalResponse, error)
}

type endorserClient struct {
	cc grpc.ClientConnInterface
}

func NewEndorserClient(cc grpc.ClientConnInterface) EndorserClient {
	return &endorserClient{cc}
}

func (c *endorserClient) ProcessProposal(ctx context.Context, in *SignedProposal, opts ...grpc.CallOption) (*ProposalResponse, error) {
	cOpts := append([]grpc.CallOption{grpc.StaticMethod()}, opts...)
	out := new(ProposalResponse)
	err := c.cc.Invoke(ctx, Endorser_ProcessProposal_FullMethodName, in, out, cOpts...)
	if err != nil {
		return nil, err
	}
	return out, nil
}

// EndorserServer is the server API for Endorser service.
// All implementations should embed UnimplementedEndorserServer
// for forward compatibility.
type EndorserServer interface {
	ProcessProposal(context.Context, *SignedProposal) (*ProposalResponse, error)
}

// UnimplementedEndorserServer should be embedded to have
// forward compatible implementations.
//
// NOTE: this should be embedded by value instead of pointer to avoid a nil
// pointer dereference when methods are called.
type UnimplementedEndorserServer struct{}

func (UnimplementedEndorserServer) ProcessProposal(context.Context, *SignedProposal) (*ProposalResponse, error) {
	return nil, status.Errorf(codes.Unimplemented, "method ProcessProposal not implemented")
}
func (UnimplementedEndorserServer) testEmbeddedByValue() {}

// UnsafeEndorserServer may be embedded to opt out of forward compatibility for this service.
// Use of this interface is not recommended, as added methods to EndorserServer will
// result in compilation errors.
type UnsafeEndorserServer interface {
	mustEmbedUnimplementedEndorserServer()
}

func RegisterEndorserServer(s grpc.ServiceRegistrar, srv EndorserServer) {
	// If the following call pancis, it indicates UnimplementedEndorserServer was
	// embedded by pointer and is nil.  This will cause panics if an
	// unimplemented method is ever invoked, so we test this at initialization
	// time to prevent it from happening at runtime later due to I/O.
	if t, ok := srv.(interface{ testEmbeddedByValue() }); ok {
		t.testEmbeddedByValue()
	}
	s.RegisterService(&Endorser_ServiceDesc, srv)
}

func _Endorser_ProcessProposal_Handler(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor grpc.UnaryServerInterceptor) (interface{}, error) {
	in := new(SignedProposal)
	if err := dec(in); err != nil {
		return nil, err
	}
	if interceptor == nil {
		return srv.(EndorserServer).ProcessProposal(ctx, in)
	}
	info := &grpc.UnaryServerInfo{
		Server:     srv,
		FullMethod: Endorser_ProcessProposal_FullMethodName,
	}
	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
		return srv.(EndorserServer).ProcessProposal(ctx, req.(*SignedProposal))
	}
	return interceptor(ctx, in, info, handler)
}

// Endorser_ServiceDesc is the grpc.ServiceDesc for Endorser service.
// It's only intended for direct use with grpc.RegisterService,
// and not to be introspected or modified (even as a copy)
var Endorser_ServiceDesc = grpc.ServiceDesc{
	ServiceName: "protos.Endorser",
	HandlerType: (*EndorserServer)(nil),
	Methods: []grpc.MethodDesc{
		{
			MethodName: "ProcessProposal",
			Handler:    _Endorser_ProcessProposal_Handler,
		},
	},
	Streams:  []grpc.StreamDesc{},
	Metadata: "peer/peer.proto",
}

```

Yes, we don't need the rest api yet, but in CoaP-API I would like to see the same brevity and conciseness of the generated code. Therefore, it should be located next to the *_grpc.pb.go file.

You can not generate a file for all proto files, but for now only for the desired gateway.proto

pfi79:
In the case of fabric-protos, generation is divided by programming languages (go, java, and node). 
CoAP is not a new programming language.

jt-nti:
The `fabric-protos-go-apiv2` Go module, and other artefacts published from the [fabric-protos/bindings](https://github.com/hyperledger/fabric-protos/tree/main/bindings) directory are intended to specifically be generated language bindings for the fabric protos, so that other projects don't need to worry about managing and generating bindings themselves. They are not intended to be a collection of different possible clients or other code that dependant projects may not need.

The language bindings can be used for all kinds of things, such as extracting information from blocks received via mqtt for example. If the current bindings start pulling in unrelated and unwanted dependencies, it's likely that people would stop using them and go back to generating bindings themselves, which wouldn't be great.

Using the fabric-protos repo to generate other modules like a coap client sounds good to me if that's possible, and may even provide a blueprint that other use cases could follow, but the output should be separate to the currently published bindings. @bestbeforetoday's suggestion of a `fabric-coap-go` repository makes sense to me. The new coap module would just include a dependency on the existing bindings [the same way as the fabric gateway does](https://github.com/hyperledger/fabric-gateway/blob/198922008535bb22182ee6bf75f628ef64c8d586/go.mod#L8).

Perhaps moving `bindings` into a new more generic `templates` folder, along with new directories for `coap` and any other future extensions, would be a good way to broaden the current scope.

> If you're worried that there won't be such an implementation for other languages (node and java).

I'm not worried about coap not covering all the same languages as the existing bindings. (As an aside I'm hoping that fabric-protos will eventually expand to cover bindings for more languages, such as C# and Rust.)

pfi79:
I will ask one more clarifying question. 
Let's forget about CoAP for a minute.

How do you feel about the protocol-gen-grpc-gateway plugin? If we need to create a rest api based on proto files in fabric, how do you see it? Will you record this in a separate repository? 

The protocol-gen-grpc-gateway plugin requires direct access to proto files and the direct presence (i.e., they should be nearby) of _grpc.pb.go files. And there will be dependencies too "github.com/grpc-ecosystem/grpc-gateway/v2/runtime" and "github.com/grpc-ecosystem/grpc-gateway/v2/utilities"

Try to do what you suggest for CoAP, do for gateway.

> The `fabric-protos-go-apiv2` Go module, and other artefacts published from the [fabric-protos/bindings](https://github.com/hyperledger/fabric-protos/tree/main/bindings) directory are intended to specifically be generated language bindings for the fabric protos, so that other projects don't need to worry about managing and generating bindings themselves. They are not intended to be a collection of different possible clients or other code that dependant projects may not need.

Are you afraid that in fabric-protos-go-apiv2, a new dependency type will appear in go.mod "github.com/...../CoaP "
 
> Using the fabric-protos repo to generate other modules like a coap client sounds good to me if that's possible, and may even provide a blueprint that other use cases could follow, but the output should be separate to the currently published bindings. @bestbeforetoday's suggestion of a `fabric-coap-go` repository makes sense to me. The new coap module would just include a dependency on the existing bindings [the same way as the fabric gateway does](https://github.com/hyperledger/fabric-gateway/blob/198922008535bb22182ee6bf75f628ef64c8d586/go.mod#L8).
> 

I consider your example with fabric-gaytway incorrect. It uses client functions from gateway/gateway_grpc.pb.go go

Here are the dependencies he's dragging in:
```
package gateway

import (
	context "context"
	grpc "google.golang.org/grpc"
	codes "google.golang.org/grpc/codes"
	status "google.golang.org/grpc/status"
)

............


type gatewayClient struct {
	cc grpc.ClientConnInterface
}

func NewGatewayClient(cc grpc.ClientConnInterface) GatewayClient {
	return &gatewayClient{cc}
}

func (c *gatewayClient) Endorse(ctx context.Context, in *EndorseRequest, opts ...grpc.CallOption) (*EndorseResponse, error) {
	cOpts := append([]grpc.CallOption{grpc.StaticMethod()}, opts...)
	out := new(EndorseResponse)
	err := c.cc.Invoke(ctx, Gateway_Endorse_FullMethodName, in, out, cOpts...)
	if err != nil {
		return nil, err
	}
	return out, nil
}

func (c *gatewayClient) Submit(ctx context.Context, in *SubmitRequest, opts ...grpc.CallOption) (*SubmitResponse, error) {
	cOpts := append([]grpc.CallOption{grpc.StaticMethod()}, opts...)
	out := new(SubmitResponse)
	err := c.cc.Invoke(ctx, Gateway_Submit_FullMethodName, in, out, cOpts...)
	if err != nil {
		return nil, err
	}
	return out, nil
}
```

That is, I can now redirect your claims to already existing entities. 
The only difference is that one has already been implemented and the community is used to it, while the other has not yet been implemented.

jt-nti:
> How do you feel about the protocol-gen-grpc-gateway plugin? If we need to create a rest api based on proto files in fabric, how do you see it? Will you record this in a separate repository?

For the output, yes, definitely, a RESTful Gateway API should be published as a new module, not included in the existing `abric-protos-go-apiv2` module.

For the build, no, it makes sense to go in `fabric-protos`. Since `protocol-gen-grpc-gateway` already exists, I'll create an example PR when I get the chance.

pfi79:
> > How do you feel about the protocol-gen-grpc-gateway plugin? If we need to create a rest api based on proto files in fabric, how do you see it? Will you record this in a separate repository?
> 
> For the output, yes, definitely, a RESTful Gateway API should be published as a new module, not included in the existing `abric-protos-go-apiv2` module.
> 
> For the build, no, it makes sense to go in `fabric-protos`. Since `protocol-gen-grpc-gateway` already exists, I'll create an example PR when I get the chance.

Thank you. Could you provide a link to your version? Maybe then I'll have no complaints about your option.

bestbeforetoday:
If there is contention about publishing CoAP bindings as a new artifact / repository only because the existing published artifacts contain both the protocol buffer and gRPC service bindings, I would favour splitting these too. This would result in three artifacts for each language:

1. Protocol buffer bindings
2. gRPC bindings
3. CoAP bindings

In some ways this might actually be better anyway since some use cases or target environments (for example, Web browsers) benefit from using the protocol buffer bindings but do not need or cannot use the gRPC bindings. Having the CoAP bindings there too just makes this problem worse.

bestbeforetoday (approved):
This is a really nice proposal. I appreciate the effort you have made to provide such a clear and detailed description.

Provided changes to the protobuf definitions do not result in additional (unnecessary) dependencies for existing gRPC (and protobuf-only) consumers of the protobuf bindings, I fully support this proposal.

jt-nti:
> f there is contention about publishing CoAP bindings as a new artifact / repository only because the existing published artifacts contain both the protocol buffer and gRPC service bindings, I would favour splitting these too. 

That makes a lot of sense, I definitely like that proposal:– a tool to inspect block data doesn't need either the gRPC or the CoAP bindings.

pfi79:
> If there is contention about publishing CoAP bindings as a new artifact / repository only because the existing published artifacts contain both the protocol buffer and gRPC service bindings, I would favour splitting these too. This would result in three artifacts for each language:
> 
> 1. Protocol buffer bindings
> 2. gRPC bindings
> 3. CoAP bindings
> 
> In some ways this might actually be better anyway since some use cases or target environments (for example, Web browsers) benefit from using the protocol buffer bindings but do not need or cannot use the gRPC bindings. Having the CoAP bindings there too just makes this problem worse.

For me, the CoAP binding is on a par with the binding to grpc and grpc-gateway.

Colleagues, in order for me to agree with you, show me a complete and standard example from the Internet where the bindings to grpc and grpc-gateway (in go) would be in different modules.

I searched almost the entire Internet and did not find such a thing. Maybe there is no such thing?
So far, I believe that by separating gRPC bindings
and CoAP bindings for the go language, you are creating big problems for the project in the future.

bestbeforetoday:
What problems do you forsee? And what are grpc and grpc-gateway? What is the problem with having a fabric-grpc artifact that in turn depends on a fabric-protos artifact? fabric-protos-apiv2 already depends on gRPC and protobuf libraries, and that seems to work OK.

Mixing protobuf message, gRPC and CoAP bindings in the same artifact means that consumers will be exposed to unnecessarily large dependency trees and have in a larger supply-chain attack surface. An application that only needs the protobuf message bindings will be forced to depend on both gRPC and CoAP, including all their dependencies. A CoAP application will be forced to depend on gRPC, despite not using it. Similar for gRPC applications.

An example where mixing the existing protobuf message bindings and the gRPC service bindings within the same fabric-protos artifact already causes problems is code to run within the Web browser. The gRPC bindings have dependencies on Node.js APIs that are not available within the browser runtime. Right now it is necessary to strip out the gRPC bindings from the published artifact to make it usable for browser applications that work with the protobuf messages.

Just to be clear, I am not saying that we _must_ split the protobuf and gRPC bindings into separate artifacts for this feature. I am saying that, if you insist that having the protobuf and gRPC bindings in a single artifact means that the CoAP bindings must also live in that same artifact, we can solve your objection by splitting them.

pfi79:
> What problems do you forsee? And what are grpc and grpc-gateway? What is the problem with having a fabric-grpc artifact that in turn depends on a fabric-protos artifact? fabric-protos-apiv2 already depends on gRPC and protobuf libraries, and that seems to work OK.
> 
> Mixing protobuf message, gRPC and CoAP bindings in the same artifact means that consumers will be exposed to unnecessarily large dependency trees and have in a larger supply-chain attack surface. An application that only needs the protobuf message bindings will be forced to depend on both gRPC and CoAP, including all their dependencies. A CoAP application will be forced to depend on gRPC, despite not using it. Similar for gRPC applications.
> 
> An example where mixing the existing protobuf message bindings and the gRPC service bindings within the same fabric-protos artifact already causes problems is code to run within the Web browser. The gRPC bindings have dependencies on Node.js APIs that are not available within the browser runtime. Right now it is necessary to strip out the gRPC bindings from the published artifact to make it usable for browser applications that work with the protobuf messages.
> 
> Just to be clear, I am not saying that we _must_ split the protobuf and gRPC bindings into separate artifacts for this feature. I am saying that, if you insist that having the protobuf and gRPC bindings in a single artifact means that the CoAP bindings must also live in that same artifact, we can solve your objection by splitting them.

I hope that the protocol-gen-go-coap will be implemented "as standard". Like the existing tools: protocol-gen-go, protocol-gen-go-grpc, and protocol-gen-grpc-gateway.  All these tools generate files: *.pb.go , *_grpc.pb.go and *.pb.gw.go for one single go module (the module name is written in the proto file, for example `option go_package = "github.com/hyperledger/fabric-protos-go/gateway ";`). It is not intended that they will be located in different go modules.

We are already using [the protocol-gen-go and protocol-gen-go-grpc tools](https://github.com/hyperledger/fabric-protos/blob/main/buf.gen.yaml#L111-#L118).

The protocol-gen-grpc-gateway tool generates RestAPI implementation code based on proto files. And this tool generates files that should be located next to *.pb.go , *_grpc.pb.go. Otherwise it won't work.

Your proposal leads to the fact that the protocol-gen-go-coap tool will be non-standard, and therefore no one else needs it. And secondly, for the future: it makes it impossible in the future to create an endpoint RestAPI generated by the protocol-gen-grpc-gateway tool.

bestbeforetoday:
Since this is so difficult, maybe as a first implementation we should not try to generate CoAP service implementations from the existing Gateway gRPC service definitions. I don't think a CoAP API absolutely needs to be rigidly tied to the gRPC service implementation. CoAP (and some hypothetical future REST API) could be defined in Swagger instead. Generating CoAP service and client stubs from the gRPC service definitions does not seem essential to deliver a viable implementation.

pfi79:
> Since this is so difficult, maybe as a first implementation we should not try to generate CoAP service implementations from the existing Gateway gRPC service definitions. I don't think a CoAP API absolutely needs to be rigidly tied to the gRPC service implementation. CoAP (and some hypothetical future REST API) could be defined in Swagger instead. Generating CoAP service and client stubs from the gRPC service definitions does not seem essential to deliver a viable implementation.

Why is it "difficult"? Have you tried it? 
I tried it - it's simple (definitely easier than swagger). 

The difficult thing is to convince you that you need to do it the way the entire go developer industry does.

C0rWin:
I'd like to share some thoughts on the binding location discussion, as I've been following this RFC closely.

I believe @pfi79 raises a valid and important architectural point that deserves careful consideration before we establish a precedent here.

The standard `protoc` toolchain was designed with the assumption that all generated artifacts for a given `.proto` file coexist in the same Go package. This isn't an arbitrary convention — it's a deliberate design choice by the protobuf/gRPC teams. The generated `_grpc.pb.go` files reference types from `.pb.go` directly, without cross-module imports, because they are part of the same compilation unit. A `_coap.pb.go` file would follow the same pattern. Moving it to a separate module breaks this contract and forces the plugin author to work around the toolchain rather than with it.

This matters practically for two reasons:

1. **Maintenance cost of a non-standard plugin.** If `protoc-gen-go-coap` must emit code targeting a different module than the proto definitions it reads, every generated file needs explicit cross-module import paths. This is fragile — module paths change, versioning gets coupled in non-obvious ways, and the plugin becomes significantly harder to maintain than one that follows the standard output convention. We'd be creating a bespoke tool that no one outside this project would recognize or reuse.

2. **Future extensibility.** As @pfi79 noted, `protoc-gen-grpc-gateway` (which generates REST reverse-proxies from gRPC service definitions) produces `.pb.gw.go` files that directly call into `_grpc.pb.go` client functions. If we establish the precedent that transport-layer bindings live in separate modules, adding a REST gateway later becomes architecturally painful or impossible without yet another deviation from standard patterns.

I understand the concern about dependency bloat in `fabric-protos-go-apiv2`. It's a real issue — browser environments already struggle with the gRPC dependencies. But I think the right solution is what @bestbeforetoday himself suggested: split the existing module into separate artifacts (protobuf-only, gRPC, CoAP). This can be done cleanly using Go multi-module workspaces within the same `fabric-protos` repository:

```
fabric-protos/
└── bindings/
    ├── go-apiv2/           # protobuf-only message types (go.mod)
    ├── go-apiv2-grpc/      # gRPC bindings, depends on go-apiv2 (go.mod)
    └── go-apiv2-coap/      # CoAP bindings, depends on go-apiv2 (go.mod)
```

This gives you:
- **Dependency isolation**: consumers import only what they need
- **Standard protoc generation**: each plugin outputs to its natural sibling directory
- **Single source of truth**: all generation happens in `fabric-protos` from the same proto files
- **Forward compatibility**: REST gateway or any future transport can follow the same pattern

The alternative — defining CoAP via Swagger independently of the proto definitions — would sacrifice the automatic synchronization that makes the `protoc-gen-go-coap` approach so valuable. We'd end up manually maintaining two API definitions that must stay in sync with `gateway.proto`, which is exactly the maintenance burden this RFC was designed to eliminate.

I think the community benefits most from staying close to the standard Go protobuf toolchain conventions. The split-module approach resolves the dependency concern without forcing non-standard plugin behavior.

C0rWin (approved):

C0rWin:
I think we have strong consensus on everything that an RFC needs to establish — the architecture, security model, API surface, and generation approach. Two maintainers have approved.

The binding location question (same module vs. separate repo) is a valid discussion, but it's a packaging and build pipeline concern, not a design concern. This is exactly the kind of detail that gets resolved naturally during implementation PRs, when there's real code to evaluate rather than hypotheticals.

RFCs set direction. Code reviews refine details. I'd propose we merge the RFC with an explicit note that the module/repository structure for CoAP bindings will be finalized during implementation, and move forward. This proposal has been in review for nearly six months — the design is solid and the community has spoken.

andrew-coleman (approved):

PR #67, opened 2025-10-27
add 0000-hyperledger_fabric_flutter_labs.md .

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
link the labs rfc to hyperledger fabric rfcs .

### DIFF
--- a/text/0000-hyperledger_fabric_flutter_labs.md
+++ b/text/0000-hyperledger_fabric_flutter_labs.md
@@ -1,0 +1,166 @@
+---
+layout : default
+title : hyperledger_fabric_flutter_labs
+parent : RFCs
+SPDX-License-Identifier : CC-BY-4.0
+---
+
+- Feature Name : support for hyperledger fabric version 2.4 or later on
+flutter and dart platforms 
+- Start Date : 2025-10-27 
+- RFC PR :
+https://github.com/LF-Decentralized-Trust-labs/LF-Decentralized-Trust-labs.github.io/pull/343
+- Fabric Component : fabric-gateway 
+- Fabric Issue :
+https://github.com/LF-Decentralized-Trust-labs/LF-Decentralized-Trust-labs.github.io/issues/344
+
+# Summary
+
+bringing purely client side support on flutter and dart platforms ( e.g.
+android ) for hyperledger fabric through hyperlefger fabric gateway
+client api for java .
+
+# Motivation
+
+since the introduction of hyperledger fabric version 2.5 , earlier
+releases of hyperledger fabric sdks deprecated .
+
+some of the labs components are deprecated , too . this includes
+archived fabric_client_flutter and fabric-server-node .
+
+in reviving the outdated fabric_client_flutter lab in sight of the
+hyperledger community's interests in the last five years and bringing up
+the flutter and dart platforms up to date and separating from relaying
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
I _thought_ that the Fabric Gateway client API should be runnable directly on Android since it allows the client application to select an appropriate gRPC channel service provider: okhttp for Android and some flavour of Netty for non-Android. See the [Fabric Gateway Java README](https://github.com/hyperledger/fabric-gateway/blob/main/java/README.md). However, the current [grpc-java README](https://github.com/grpc/grpc-java?tab=readme-ov-file#download) says that Android should also use grpc-protobuf-lite instead of grpc-protobuf. This might be a problem since both fabric-gateway and its [fabric-protos](https://hyperledger.github.io/fabric-protos/) dependency have a declared dependency on grpc-protobuf. That might be fixable by replacing grpc-protobuf with grpc-protobuf-lite in a client application's Gradle dependency resolution rules. If not, I am open to suggestions on how to resolve that in fabric-gateway and fabric-protos, provided it is not overly disruptive for existing users. This would allow Android devices to talk directly to the peer's Fabric Gateway service without an intermediary server-side Gateway service interacting with Fabric on the client's behalf.

The fabn-895 document you reference talks about off-line signing messages. A similar flow is already available using the Fabric Gateway client API, and referred to as "off-line signing" in the [Contract JavaDoc](https://hyperledger.github.io/fabric-gateway/main/api/java/org/hyperledger/fabric/client/Contract.html#off-line-signing-heading). A less chatty evolution of this existing capability might be to provide utilities that allow the Android client to build (and sign) the protocol buffer messages to be sent to Fabric in their entirety, and then just forwarded to a server-side intermediary to do the gRPC interaction with the peer's Gateway service. It might just about be possible to do this with the existing Fabric Gateway client API using a fake gRPC connection in the Android client application, then extracting the protocol buffer message using the offline-signing APIs to send on to the intermediary using REST or some other mechanism. I prototyped a more streamlined API for achieving this for JavaScript Web clients in [this pull request](https://github.com/hyperledger/fabric-gateway/pull/701/files).

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
@bestbeforetoday thanks for the documents . it ' s great to know that
grpc channel service providers are customizable .

the lab ' s main objective is to let users to connect directly to a
fabric network without an intermediary server and service .

the fabn-895 doc i referred to in the rfc is the one i ' ve implemented
in a low-level , mobile-first platform api for flutter and dart
platforms in connecting to a fabric network :
https://lf-decentralized-trust-labs.github.io/labs/archived/fabric-client-flutter.html
.

this lab is intended to replace the low-level apis with earlier sdk
versions with the high-level fabric-gateway client api . i ' ve
rephrased it in the latest commit in connection with fabn-895 .

in case if communicating through the proposed web api of fabric-gateway
, the dart package can be ported to js / wasm code for web directly and
be used to connect to the web api via js / wasm interface . although
with a intermediary rest server , etc . isn ' t a promised proposal
doing the offline signing .

the other obstacle i could think of is the objective-c / swift api for
the fabric-gateway once the certificate part is cleared on the apple
platforms . i ' ve also updated and reflected this in the latest commit
.

PR #68, opened 2026-08-09
feat: add Fabric snapshot to Fabric X migration RFC

syndbg:
## Description 

This RFC defines an offline migration from a Fabric Classic peer snapshot to
Fabric-X.

It specifies:

- validation of the source snapshot and deterministic genesis-data bundle;
- one Fabric channel per Fabric-X network, with mapped application namespaces;
- migration of public state and transaction IDs for anti-replay;
- explicit recording of PDC exclusions (hashes) without exposing private values or
  hashes as public state;
- atomic committer bootstrap at target anchor `B`; and
- verification, idempotency, activation, multi-organization agreement, and
  recovery.

Source block history, online migration, PDC values, automatic chaincode
translation, and channel consolidation are outside this proposal.

#### Proof of concept

- [Exporter and genesis-data bundle](https://github.com/syndbg/fabric-x-migrate-poc)
- [Fabric-X committer implementation](https://github.com/syndbg/fabric-x-committer/tree/feat-add-fabric-to-fabric-x-migration)

#### Related issues

- [Fabric to Fabric-X ledger migration mentorship project](https://github.com/LF-Decentralized-Trust-Mentorships/mentorship-program/issues/65)

cc @liran-funaro @cendhu @tock-ibm

### DIFF
--- a/text/0000-fabric-x-snapshot-migration.md
+++ b/text/0000-fabric-x-snapshot-migration.md
@@ -1,0 +1,1069 @@
+---
+layout: default
+title: Fabric Classic Snapshot Migration to Fabric-X
+nav_order: 3
+---
+
+- Feature Name: fabric_classic_snapshot_migration_to_fabric_x
+- Start Date: 2026-08-04
+- RFC PR: (leave this empty)
+- Fabric Component: Fabric Classic peer ledger, Fabric-X committer
+- Fabric Issue: https://github.com/LF-Decentralized-Trust-Mentorships/mentorship-program/issues/65
+
+# Summary
+[summary]: #summary
+
+This RFC defines an offline migration from Hyperledger Fabric Classic to
+Fabric-X. An official Fabric peer snapshot at source height `H` is the source of
+truth. A Go exporter validates the snapshot, maps channels and chaincode
+namespaces to Fabric-X networks and namespaces, excludes unsupported private
+data artifacts, and writes a deterministic genesis-data bundle. The Fabric-X
+committer imports the bundle into a new target in one atomic transaction. It
+records and verifies the result, then blocks further imports once the operator
+activates the target.
+
+The initial topology maps one Fabric channel to one Fabric-X network. Fabric-X
+currently supports one channel per network and multiple namespaces inside that
+channel. The migration preserves current public state and transaction IDs for
+anti-replay. It does not copy Fabric block history, transaction envelopes,
+historical values, private collection values, channel configuration bytes, or
+classic chaincode packages.
+
+An official peer snapshot does not contain Fabric transaction or block
+history. This RFC therefore preserves current state and the transaction-ID
+anti-replay baseline, while requiring the source ledger archive for historical
+audit and queries. It does not claim to reconstruct transactional history from
+data that the snapshot does not contain.
+
+# Motivation
+[motivation]: #motivation
+
+Fabric-X is a separate, high-throughput implementation within the Fabric
+ecosystem. Its microservice architecture, single-channel network model,
+namespace storage, distributed SQL state database, and Fabric Smart Client
+(FSC) or custom-endorser programming model are not drop-in replacements for a
+Fabric Classic peer.
+
+Existing Fabric deployments need a defined transition boundary before they can
+evaluate Fabric-X for production use. A direct database copy is not sufficient:
+
+* Fabric state is organized by channel, chaincode namespace, and private-data
+  collection, while Fabric-X uses one channel per network and application
+  namespaces.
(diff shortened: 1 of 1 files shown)


syndbg:
### PoC testing coverage

The [migration integration test](https://github.com/syndbg/fabric-x-committer/blob/feat-add-fabric-to-fabric-x-migration/integration/migration/migration_test.go) uses the exporter and disposable Fabric and Fabric-X networks. The primary path:

1. creates the Fabric-X network and target namespace;
2. records and freezes the target at anchor `B`;
3. exports and imports the genesis-data file;
4. verifies the imported state, transaction IDs, target configuration, namespace map, installed policy, and migration record;
5. activates the migration and restarts Fabric-X services; and
6. submits a normal transaction through the orderer and confirms that it updates an imported key from version `0`.

The integration scenarios cover:

- Fabric 2.5.16 and Fabric 3.1.5 LevelDB snapshots;
- a Fabric 3.1.5 CouchDB snapshot;
- retrying the same bundle without importing its rows again;
- rejecting imports after activation;
- two source channels migrated into separate Fabric-X networks;
- two committer organizations importing the same bundle, reporting identical verification results, and committing the same later transaction; and
- reconstructing an empty committer database by processing Fabric-X blocks through `B`, reapplying the same bundle, and then processing the later Fabric-X blocks.

The rebuild scenario reuses the original bundle. It does not test whether a replacement bundle is rejected after the database-local migration record has been lost.

The [committer bootstrap tests](https://github.com/syndbg/fabric-x-committer/blob/feat-add-fabric-to-fabric-x-migration/service/vc/bootstrap_test.go) cover the failure paths separately: atomic rollback, missing or non-empty namespaces, existing transaction IDs, missing block `0`, unsupported target policies, changed state or policy bindings, changed anchors, and attempts to use a different bundle while the migration record exists.

Run the focused tests with:

```sh
go test ./cmd/committer ./service/vc
go test -tags=integration ./integration/migration \
  -run TestInitFromSnapshot \
  -count=1 -v
```

text/0000-fabric-x-snapshot-migration.md
```diff
@@ -0,0 +1,1069 @@
...
+checkpoint block, but ordinary fabricated application transactions are not an
+honest representation.
+
+## Omit Fabric transaction IDs
+
+Rejected because a transaction committed before `H` could be submitted to the
+target again. Version `1` imports the IDs into `migrated_tx_ids`, and normal
+duplicate detection checks that immutable registry together with `tx_status`.
+The importer does not invent Fabric-X status or height values for source
+transactions.
```
syndbg, line 858: On a second note, not needed since they'll be rejected. To be removed from the snapshot and import.
Extra point - have an integration test verify this.

cendhu:
@syndbg Please open this rfc in the https://github.com/hyperledger/fabric-x-rfcs repo.

cendhu:
@syndbg 

 1. Simplify migration flow: initialize Fabric-X network with MSP and namespaces, stop the network, load snapshot data directly into each organization’s committer database using the migration tool, verify results, then restart the network.

 2. Avoid intermediate migration bundles or states. Mapping should happen inside migration tool while streaming snapshot records into target database. Remove .fxgenesis, USTAR, manifest, and exporter/importer split.

 3. Run migration tool independently for each organization. All runs must use identical source snapshots and mapping configuration. Compare resulting state digests and record counts before resuming network. The state digest computation for fabric-x database is being implemented as part of another feature. This feature can be used to verify digest across committers once the migration is done. 

 4. Remove transaction-ID migration and anti-replay handling. Migrated transaction IDs are not required in Fabric-X target state.

 5. Initialize every migrated key-value at Fabric-X version 0. Do not preserve Fabric Classic key versions as target versions or provenance.

 6. Import private-data hashes as Fabric-X application state. Private values remain excluded.

 7. Support channel consolidation directly in migration tool through administrator-provided mappings from source channel/namespace pairs to target network/namespace pairs. Reject mapping collisions.

cendhu:
> @syndbg
> 
> 1. Simplify migration flow: initialize Fabric-X network with MSP and namespaces, stop the network, load snapshot data directly into each organization’s committer database using the migration tool, verify results, then restart the network.

As suggested by Liran, we can reuse the chaincode policy in fabric as the namespace policy and reuse MSPs from Fabric itself. With that, we do not need to initalize a fabric-x network to create namespaces. We can directly load the data into the database using the migration tool and the snapshot file.

syndbg:
Continuing in https://github.com/hyperledger/fabric-x-rfcs/pull/9.
Thank you for the comments and review. ^^

